PR #279, opened 2023-02-03
[WIP] v0.2 API 

ml-evs:
This PR contains some skeleton ideas for the v0.2 API.

The idea is to use flask-openapi3 to annotate routes and automatically validate the responses with pydantic.

This will involve adding some response-specific pydantic models that are not part of the core data model --- need to decide what kind of versioning we can get away with for models themselves.


ml-evs:
stale

PR #293, opened 2023-02-13
Add cycle vs capacity plots to echem block

ml-evs:
Todo:

- [x] Make it a selectable option in the UI
- [ ] Consider adding other cycle properties like coulombic efficiency etc. and average voltages

### DIFF
--- a/pydatalab/pydatalab/blocks/echem_block.py
+++ b/pydatalab/pydatalab/blocks/echem_block.py
@@ -202,7 +202,7 @@ class CycleBlock(DataBlock):
             LOGGER.warning("No file_id given")
             return
 
-        derivative_modes = (None, "dQ/dV", "dV/dQ")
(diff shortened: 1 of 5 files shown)


jdbocarsly:
This seems to work pretty well now, but I did it by adding a database call to get the mass in the CycleBlock datablock. This isn't super elegant, but I didn't realize until after I had done that that you wanted to pass it in from the frontend instead. It wouldn't be  hard to swithc over to that approach—we can discuss what makes sense.

jdbocarsly:
also this breaks all the echem tests just because I renamed columns to include units ("Capacity" -> "capacity (mA)")

ml-evs:
Added filtering by cycle to this plot, going to merge as we discussed (with the expectation that we tidy this up at some point and improve the dataflow between blocks and plots)

PR #297, opened 2023-02-16
Intercept flask-dance redirects and point back to HTTP referer

### LINKED ISSUE #205: Server does not redirect correctly after first login

ml-evs:
Addresses #205, but does not yet close it.

Would be nice to make a login modal to replace the current direct link to `/login/github` that could link to the appropriate OAuth provider endpoints, e.g., "Sign in with Github" and "Sign in with ORCID"


ml-evs:
This is the best I can do re: redirects for now @jdbocarsly -- think we should merge and revisit at a later date.

jdbocarsly (approved):
look good but can you stop the pydatalab banner from scrolling since some people will see this.

PR #302, opened 2023-02-24
Add compression to heavy routes

jdbocarsly:
From some quick investigation, I think most of the load time on edit pages is actually the time spent in sending the large bokeh plots over network. We might be able to speed this up by compressing routes, which `flask-compress` does nicely. The caveat is that this will only speed things up if the time spent compressing is less than the network time. I'm guessing it will be, but we can check once deployed.

This currently compresses all routes, but probably we only need to compress the very large routes.

### DIFF
--- a/pydatalab/pydatalab/main.py
+++ b/pydatalab/pydatalab/main.py
@@ -2,6 +2,7 @@ from typing import Any, Dict
 
 from dotenv import dotenv_values
 from flask import Flask, redirect, request, url_for
+from flask_compress import Compress
 from flask_cors import CORS
 from flask_login import current_user, logout_user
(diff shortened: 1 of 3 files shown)


pydatalab/pydatalab/main.py
```diff
@@ -56,6 +59,10 @@ def create_app(config_override: Dict[str, Any] = None) -> Flask:
 
     pydatalab.mongo.create_default_indices()
 
+    # if False, don't compress all responses, we will add it on a per-view basis
+    # app.config["COMPRESS_REGISTER"] = False
```
ml-evs, line 63:
```suggestion
    app.config["COMPRESS_REGISTER"] = True
```

PR #304, opened 2023-02-25
Add translation of PIDINST schema to pydantic

ml-evs:
This PR simply adds a pydantic model for instruments that is a direct translation of the RDA PIDINST schema https://github.com/rdawg-pidinst/schema/blob/master/schema.rst.

They themselves provide two versions that includes a translation to DataCite's format. It's not clear how they expect this data to be reused.

I don't like the resource layout (it is very XML-y) so think we should use this as an example for having our own internally schema but being able to map out to others. I may offer to contribute this file back upstream to PIDINST (maybe as LinkML yaml instead) and see what they say.

### DIFF
--- a/pydatalab/pydatalab/models/instruments.py
+++ b/pydatalab/pydatalab/models/instruments.py
@@ -1,0 +1,148 @@
+import datetime
+from typing import List, Literal, Optional
+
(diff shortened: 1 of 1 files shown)


ml-evs:
Superseded by #531

PR #307, opened 2023-03-03
API key auth, admin database validity checks and various other refactors

ml-evs:
This PR bundles lots of associated refactoring around API key login, logging and other linting.

- [x] API key auth via `api_keys` collection which stores a sha-512 hash of a key that is refreshed by an authenticated user with `/get-api-key`.
- [x] Strip console.log in production for vue
- [x] Properly pass errors from the server to alerts in the UI
- [x] Update logging to use Rich and properly account for the root logger
- [x] Replace a deprecated pandas append call that was flooding the logs
- [x] Adds a task for checking item validity in the API by providing an API URL and key (can be run remotely)
- [x] Set default secret key as random string

### DIFF
--- a/pydatalab/pydatalab/bokeh_plots.py
+++ b/pydatalab/pydatalab/bokeh_plots.py
@@ -283,6 +283,7 @@ def double_axes_echem_plot(
         )
 
         p3.legend.location = "right"
+        p3.y_range.start = 0
 
     lines = []
     grouped_by_half_cycle = df.groupby("half cycle")


pydatalab/pydatalab/routes/utils.py
```diff
@@ -23,7 +23,8 @@ def get_default_permissions(user_only: bool = True) -> Dict[str, Any]:
         return {}
 
     if (
-        current_user.is_authenticated
+        not user_only
+        and current_user.is_authenticated
```
jdbocarsly, line 27:
```suggestion
        current_user.is_authenticated
```

PR #309, opened 2023-03-08
Fix empty string in float value bug

### LINKED ISSUE #308: Inputting text and then deleting in float fields causes save error
In fields that will be interpeted as Optional[Float] (e.g.) by the API, they originally are generally initialized with `null`/`None`. Upon inserting text, and then deleting it, these fields are left with values of the empty string instead of `null`. On save, the server throws an error as `""` cannot be interpreted as `Optional[Float]`.

jdbocarsly:
closes #308 by making the generic computed getters in `field_utils.js` automatically save null whenever an empty string is provided. This doesn't seem to break any existing functionality that I could find, but we should monitor.

### DIFF
--- a/webapp/src/field_utils.js
+++ b/webapp/src/field_utils.js
@@ -34,7 +34,7 @@ export function createComputedSetterForItemField(item_field) {
       console.log(value);
       store.commit("updateItemData", {
         item_id: this.item_id,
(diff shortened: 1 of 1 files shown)


ml-evs (approved):
:japanese_ogre:

PR #311, opened 2023-03-08
Add concept of refcodes

ml-evs:
This PR introduces refcodes as the real, persistent, globally unique identifiers used across datalab. These are prefixed by e.g., `grey:` and by default use 6 random upper case letters.

Also relaxes the constraints on IDs to allow '.'  in them, but enforces no punctuation at the start or end of an ID.


webapp/src/components/FormattedRefcode.vue
```diff
@@ -0,0 +1,60 @@
...
+      default: false,
+    },
+    maxLength: {
+      type: Number,
+      default: NaN,
+    },
+  },
+  computed: {
+    badgeColor() {
+      return "LightPink";
```
jdbocarsly, line 36:
```suggestion
      return "LightGrey";
```

jdbocarsly (changes requested):

PR #320, opened 2023-03-13
Mark remote files as not live if not written to within some period

ml-evs:
Old file sync logic:

1. If marked as `is_live` (any remote synced echem file), check local copy of file and see if it is older than the minimum cache age (a few days)
2. If so, redownload the file and update the database with the current timestamp

i.e., every single file is redownloaded for echem.

New logic:

1. If marked as `is_live`, call `stat` on the remote file and check if it is newer than the local version with some minimum difference of 1 hour.
2. If so, resync. If not, and the remote stat is older than 31 days, also mark the file as no longer `is_live`, such that future checks are not performed. This makes the echem block load about 100 times faster (as confirmed by timing the debug logs, also added in this PR).


ml-evs:
Need to check that files are not just currently being over-synced and adjust this PR accordingly.

PR #321, opened 2023-03-13
Graph view updates: filter out deleted nodes, add cells and item-centred graphs

ml-evs:
- filter out deleted nodes in graph endpoint
- add relationships to cells and render cells in the graph
- re-enable user interaction with the graph (zoom/pan etc.)
- add graph endpoint specialization for graph centred on given item


jdbocarsly:
I believe the centered graph feature will close #249

jdbocarsly (approved):
looks "ok" to me

PR #326, opened 2023-03-23
Add on-the-fly TIFF encoding to image block

ml-evs:
Convert TIFFs to base64 on-the-fly and add them to the media block.


jdbocarsly (approved):
media.....

PR #341, opened 2023-04-10
Block for LLM chat interface

ml-evs:
Very quick WIP PR, drags in too many extra commits and should be factored out into a plugin, but here is the basis of a chatblock for LLMs interacting with local JSON graph data.


ml-evs:
Superseded by #344

PR #369, opened 2023-05-03
fix cycle block bug that prevents use of live echem files

jdbocarsly:
When attempting to load a new live file into the cycle block, the page crashes with `"Error getting sample data: ("local variable 'cycle_summary_df' referenced before assignment",)` and the sample can no longer be opened.

I believe this is just a simple bug where `cycle_summary_df` is never set (to `None`) for live files. Hopefully moving this one line fixes the issue.


ml-evs (approved):
🙇

PR #370, opened 2023-05-04
Assorted echem tweaks and fixes

### LINKED ISSUE #360: Use better (or no) colourmap for echem block when there are few cycles
Currently with only 2 or 3 cycles you end up with a black line and a yellow line, which is not very clear. We should probably restrict the colourscheme slightly to the middle 75% of the one we are using now.

### LINKED ISSUE #367: Improve subsampling on echem plots
It would make the echem plots look better if we always ensured that the first and last point of each half cycle was included.

### LINKED ISSUE #368: Ivium column normalization
See https://github.com/BenSmithGreyGroup/navani/issues/4

### LINKED ISSUE #375: Unable to copy cell with existing constituents
As reported by @jdbocarsly

### LINKED ISSUE #376: Unable to create a sample from scratch with inline references
This was missing from #346.

ml-evs:
- [x] Apply fix time columns to in navani for Ivium and arbin (closes #368)
- [x] Always sample endpoints of each half cycle when reducing data (closes #367)
- [x] Trim the ends of the colour palette when making an echem plot (closes #360)
- [x] Fix copying bug from copying a cell with pre-existing components (closes #375)
- [x] Fix missing UI route for making sample with inline references (closes #376)

### DIFF
--- a/pydatalab/Pipfile
+++ b/pydatalab/Pipfile
@@ -16,7 +16,6 @@ scipy = "*"
 matplotlib = "*"
 pymongo = "*"
 galvani = "*"
-navani = {git = "git+https://github.com/the-grey-group/navani.git@v0.1.2"}
 openpyxl = "*"
 nmrglue = "*"
 python-dotenv = "*"
@@ -28,6 +27,7 @@ flask-compress = "*"
 pillow = "*"
 openai = "*"
 tiktoken = "*"
+navani = {git = "git+https://github.com/the-grey-group/navani.git@v0.1.3"}
 
 [dev-packages]
 pytest = "*"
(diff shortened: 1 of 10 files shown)


jdbocarsly (approved):
looks good

pydatalab/Pipfile
```diff
@@ -27,6 +26,7 @@ periodictable = "*"
 flask-compress = "*"
 pillow = "*"
 openai = "*"
+navani = {git = "https://github.com/the-grey-group/navani.git@v0.1.3"}
```
ml-evs, line 29:
```suggestion
navani = {git = "git+https://github.com/the-grey-group/navani.git@v0.1.3"}
```

ml-evs: Seems to work either way but will keep this comment open for future reference

PR #374, opened 2023-05-09
Add ability to have selectable color field in selectable_axes_plot

jdbocarsly:
Building on the recent generalizations of the selectable_axes_plot, this branch starts to add functionality to allow for selcting a field that controls the color of the points.

### DIFF
--- a/pydatalab/pydatalab/bokeh_plots.py
+++ b/pydatalab/pydatalab/bokeh_plots.py
@@ -42,6 +42,20 @@ SELECTABLE_CALLBACK_y = """
   yaxis.axis_label = column;
 """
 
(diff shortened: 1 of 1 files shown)


jdbocarsly:
This current version has two issues:

    It only works when there is only one set of data shown. It seems to crash the site if there are multiple. In general, this can be fixed fairly easily, but I'm not really sure what the use case for having multiple data sets, each colored with color maps would be- needs to be discussed.
    The label/title of the colorbar doesn't update properly, which is a bug in the version of bokeh we use. Apparently it should be fixed in v3.0+, so we would want to look at updating before deploying this feature.

PR #383, opened 2023-05-16
Add sample/user/cell counters to deployment about page

### LINKED ISSUE #361: Add vanity counter for number of users and number of samples on about page
Probably a nice thing to have, so that users can see they are not just adding all their data on their own...

ml-evs:
Closes #361 

This PR adds the endpoint `/info/stats` for getting some counters about a deployment, and then shows them on the about page in the app, currently just in a very basic table.

To-do:

- [x] make table nice

### DIFF
--- a/webapp/src/components/StatisticsTable.vue
+++ b/webapp/src/components/StatisticsTable.vue
@@ -1,0 +1,44 @@
+<template>
+  <div v-if="statsAvailable" class="center ml-auto mr-auto">
+    <table>
+      <tr>
+        <td>Active Users</td>
+        <td>Samples</td>
+        <td>Cells</td>
+      </tr>
+      <tr>
+        <td>{{ counts["users"] }}</td>
+        <td>{{ counts["samples"] }}</td>
+        <td>{{ counts["cells"] }}</td>
+      </tr>
+    </table>
+  </div>
+</template>
+
+<script>
+import { getStats } from "@/server_fetch_utils.js";
+
+export default {
+  data() {
+    return {
+      statsAvailable: false,
+    };
+  },
+  computed: {
+    counts() {
+      return this.$store.state.counts || { users: 0, samples: 0, cells: 0 };
+    },
+  },
+  methods: {
+    fetchStats() {
+      getStats().catch(() => {
+        this.statsAvailable = false;
+      });
+      this.statsAvailable = true;
+    },
+  },
+  created() {
+    this.fetchStats();
+  },
+};
+</script>
--- a/webapp/src/components/StatisticsTable.vue
+++ b/webapp/src/components/StatisticsTable.vue
@@ -1,15 +1,15 @@
 <template>
-  <div v-if="statsAvailable" class="center ml-auto mr-auto">
+  <div v-if="counts" class="mx-auto">
     <table>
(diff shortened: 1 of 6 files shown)


webapp/src/components/StatisticsTable.vue
```diff
@@ -0,0 +1,44 @@
...
+  </div>
+</template>
+
+<script>
+import { getStats } from "@/server_fetch_utils.js";
+
+export default {
+  data() {
+    return {
+      statsAvailable: false,
```
jdbocarsly, line 24:
```suggestion
      counts: null,
```

webapp/src/components/StatisticsTable.vue
```diff
@@ -0,0 +1,44 @@
...
+    };
+  },
+  computed: {
+    counts() {
+      return this.$store.state.counts || { users: 0, samples: 0, cells: 0 };
+    },
+  },
```
jdbocarsly, line 31:
```suggestion
```

webapp/src/components/StatisticsTable.vue
```diff
@@ -0,0 +1,44 @@
...
+  methods: {
+    fetchStats() {
+      getStats().catch(() => {
+        this.statsAvailable = false;
+      });
+      this.statsAvailable = true;
+    },
+  },
```
jdbocarsly, line 39:
```suggestion
```

webapp/src/store/index.js
```diff
@@ -25,6 +25,10 @@ export default createStore({
       // sampleSummaries is an array of json objects summarizing the available samples
       state.sample_list = sampleSummaries;
     },
+    setStats(state, counts) {
+      // sampleSummaries is an array of json objects summarizing the available samples
+      state.counts = counts;
+    },
```
jdbocarsly, line 31:
```suggestion
```

webapp/src/components/StatisticsTable.vue
```diff
@@ -0,0 +1,44 @@
+<template>
+  <div v-if="statsAvailable" class="center ml-auto mr-auto">
```
jdbocarsly, line 2:
```suggestion
  <div v-if="counts" class="center ml-auto mr-auto">
```

webapp/src/components/StatisticsTable.vue
```diff
@@ -0,0 +1,44 @@
...
+    fetchStats() {
+      getStats().catch(() => {
+        this.statsAvailable = false;
+      });
+      this.statsAvailable = true;
+    },
+  },
+  created() {
+    this.fetchStats();
+  },
```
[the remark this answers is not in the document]
jdbocarsly: I _think_ this will work as if there is an error this will just leave this.counts at null, but otherwise it might need to be something like this:
```javascript
this.counts = await getStats().catch(() => null)
```

jdbocarsly (changes requested):
Sorry for the many suggestions which are probably not so readable. Basically, I suggest we don't use the store for this since all the data is self-contained within a single component. This should simplify everything quite a bit.

Also suggesting not having a `statsAvailable` item, just having `counts` start at `null` and using  its own value in the check.

PR #390, opened 2023-05-19
Block loading fixes

jdbocarsly:
A few fixes to block loading:

1. Since blocks are still never fully deleted, deleted blocks were being rendered on each page load

3. blocks were not being saved on update (e.g., chatting to the chatbot or changing the file displayed) leading to annoying "unsaved" popup notifications or lost data. Now they are automatically saved, and the frontend keeps track of the saved status of each individual block to know whether to show the "unsaved" status warning/popup.

4. changed the chatbot so that it renders on both `to_db()` and `to_web()`. This way, when the block is first generated, the original message is saved in the db instead of just being sent to the web. This is a bit of a holdover until we implement a proper datablock lifecycle, but for the time being it should avoid an issue where blocks that were created but never saved would constantly call the openai API on every page load.

6. misc. bug fixes

### DIFF
--- a/pydatalab/pydatalab/blocks/blocks.py
+++ b/pydatalab/pydatalab/blocks/blocks.py
@@ -122,9 +122,9 @@ class DataBlock:
     @classmethod
     def from_web(cls, data):
         LOGGER.debug("Loading block %s from web request.", cls.__class__.__name__)
-        Block = cls(data["item_id"], unique_id=data["block_id"])
-        Block.update_from_web(data)
-        return Block
+        block = cls(data["item_id"], unique_id=data["block_id"])
+        block.update_from_web(data)
+        return block
 
     def update_from_web(self, data):
         """update the object with data received from the website. Only updates fields


pydatalab/pydatalab/blocks/blocks.py
```diff
@@ -123,7 +122,7 @@ def to_web(self):
     @classmethod
     def from_web(cls, data):
         LOGGER.debug("Loading block %s from web request.", cls.__class__.__name__)
-        Block = cls(data["item_id"])
+        Block = cls(data["item_id"], unique_id=data["block_id"])
         Block.update_from_web(data)
         return Block
```
ml-evs, line 127:
```suggestion
        block = cls(data["item_id"], unique_id=data["block_id"])
        block.update_from_web(data)
        return block
```

ml-evs (approved):
Literally one inconsequential suggestion (maybe don't even bother) but otherwise this looks like a great stop-gap until we overhaul blocks, I've deployed it on dev if you want to play around with it a bit more before merging.

PR #401, opened 2023-05-23
Load blocks asynchronously

jdbocarsly:
This PR should dramatically help how responsive the site feels, by first loading the page without passing the blocks through DataBlock objects (potentially triggering plotting, etc.) and _then_ asynchronously updating the blocks by calling `updateBlockFromServer` on each individual block. This means the page should load quickly with slower plots, etc. potentially coming in later.

### DIFF
--- a/pydatalab/pydatalab/routes/v0_1/items.py
+++ b/pydatalab/pydatalab/routes/v0_1/items.py
@@ -638,7 +638,7 @@ def save_item():
             400,
         )
 
-    return jsonify(status="success"), 200
+    return jsonify(status="success", last_modified=updated_data["last_modified"]), 200
 
 
 save_item.methods = ("POST",)  # type: ignore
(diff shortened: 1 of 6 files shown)


pydatalab/pydatalab/routes/v0_1/items.py
```diff
@@ -447,7 +447,9 @@ def delete_sample():
 delete_sample.methods = ("POST",)  # type: ignore
 
 
-def get_item_data(item_id, load_blocks=True):
+def get_item_data(item_id, load_blocks=False):
+    # load_blocks = request.args.get("load_blocks") or load_blocks
```
ml-evs, line 451:
```suggestion
def get_item_data(item_id, load_blocks: bool = False):
    """Generates a JSON response for the item with the given `item_id`,
    additionally resolving relationships to files and other items.
    
    Parameters:
       load_blocks: Whether to regenerate any data blocks associated with this
           sample (i.e., create the Python object corresponding to the block and 
           call its render function).
           
    """
```

Mostly for my benefit, does that description make sense?

ml-evs (dismissed):
Minor suggestion but if this works on dev then its good to go!

ml-evs (approved):
Great job!

PR #292, opened 2023-02-11
Add "collections" type

ml-evs:
Adds collections endpoints and UI for assigning samples within collections.

to-do:

- [x] All API endpoints and database/indexes
- [ ] API tests
   - [x] creating collections with entries
   - [x] creating entries in a collection
   - [ ] adding existing entry to collection
- [x] Collection table UI 
- [x] Edit page for collections with table of entries
- [ ] UI tests
- [ ] Support for collection blocks (e.g., chat block)
- [ ] Test for collection (comment block)
- [ ] Update UI for fancy table

Address #255

PR #403, opened 2023-05-26
Add `collections` type to API

ml-evs:
This PR takes all Python changes from #292 and separates them into a new PR.

Few things left to check/discuss:

- [ ] Add a test for adding an existing item to an existing collection
- [ ] Collections can currently share `item_ids` -- should we make this unique across all users for future-proofing? This would be the simplest constraint for now
- [ ] Need to add tests for adding blocks to a collection


ml-evs:
I *think* this is ready for review now @jdbocarsly (if by the time you are reading this the webapp tests have passed), when you get a spare moment... there's no UI here but I think the tests I've added are fairly extensive. We should merge your PRs before this, then release, then try to get the combined collection API + UI PRs in before the next release (might be 0.3 at this rate).

jdbocarsly (changes requested):

ml-evs:
Will merge this now so I can test UI stuff at some point this week.

PR #407, opened 2023-05-26
NMR block process number bugfix

### LINKED ISSUE #402: behavior when switching nmr filenames
If a process is selected in the nmr block, and then the filename is changed to a file that doesn't have that process number, the graph is not updated. In this case, it should just select the first process.

jdbocarsly:
Bruker NMR files can each have a different set of "process numbers". This PR is a small bugfix to automatically select an existing process if the block has an invalid process number selected for any reason ( _i.e._ if the filename has just been changed and the selected process is not a valid process number in the new experiment).

Closes #402


ml-evs (approved):
This has been working fine on dev for a while, will merge

PR #416, opened 2023-07-07
Add CITATION.cff file for repository

ml-evs:
This is to enable easier automated archives for each release on Zenodo.


ml-evs:
Don't think there's any problem if I force this one through...

PR #422, opened 2023-07-20
Implemented Raman block

elbee99:
Added a block for plotting Raman spectroscopy data as well as a new baseline fitter


ml-evs (approved):
Thanks for this @elbee99!

Issue #449, opened 2023-09-13
Generate block-level documentation with screenshots


ml-evs:
Currently our block docs are just rendered from Python docstrings, but really they should have standalone pages on the docs site and include screenshots and additional context.

PR #451, opened 2023-09-13
Manual overhaul of API docs

### LINKED ISSUE #450: API reference in docs site hierarchy is broken
At some point in a recent build, the hierarchy of the API reference page has been lost (as evidenced by the table of contents), probably because we are currently bundling the reference as one giant page that includes all modules.

The way around this is to add (automatically or otherwise) a file in the docs hierarchy per module. An example of this being done automatically can be found at https://github.com/Materials-Consortia/optimade-python-tools/blob/133bc8b4682cdb492ed652835c3a6d7ee7fe4629/tasks.py#L182-L290

ml-evs:
Closes #450 and begins works on #449.


ml-evs:
These flaky cypress tests are very annoying; still seems to depend on what order they run in but I cannot reproduce locally. Will merge this and investigate later.

PR #475, opened 2023-10-24
Adding limits on file uploads

### LINKED ISSUE #471: Add a file upload size limit
Related to #468, but this is something we need to do anyway as Flask does not enforce any limit itself and will happily fill its own memory and then temporary space (trying to force my own laptop to crash right now to see what happens).

We should limit this in a few places:

- UI file upload selection (fairly large but arbitrary restriction ~ 100 GB)
- API check disk space during upload
- API restrict uploads to some large number of GB that fit in memory, then figure out how to do chunked uploads in the future

ml-evs:
Closes #471 by adding file upload limits:

- [x] In the Uppy UI (100 GB in total and 10,000 files per upload)
    - [x] Both values configurable via env variables
- [x] In the API:
    - [x] At the level of Flask `MAX_CONTENT_SIZE` -- now this value is configurable to prevent massive files filling `/tmp` disks that could lead to instability on some systems -- think we are unaffected on the current deployment but will try to verify later
    - [x] At the level of remaining disk space in the configured mount point, e.g., whether there is enough space in `CONFIG.FILE_DIRECTORY` to actually store the thing. This is done with a pymongo lock that prevents new files being uploaded until previous files have been correctly stored in the fs and database.

Also this PR makes the files directory on app startup, if not present. Previously the routes themselves were making it on first upload, potentially with all required subdirectories. This can be a bit dangerous if the files directory is a dynamic mount point, as the wrong disk may silently be used if not previously mounted. We should probably make this more robust in the future, with better specification of e.g. remote data sources too for the files directory (S3 or other blob storage). May be worth abstracting this away with something like Maggma as a "store" controller but equally this may overcomplicate things for little gain for us.

Additionally, in the future/now if easily done, it would be nice to add better feedback on failures in the UI, i.e., overriding the default uppy error message with better messages directly from our API. This might be straightforward to do but tricky to test.

### DIFF
--- a/pydatalab/pydatalab/config.py
+++ b/pydatalab/pydatalab/config.py
@@ -115,6 +115,15 @@ class ServerConfig(BaseSettings):
         None, description="A dictionary containing metadata to serve at `/info`."
     )
 
+    MAX_CONTENT_LENGTH: int = Field(
+        100 * 1000 * 1000,
+        description=r"""Direct mapping to the equivalent Flask setting. In practice, limits the file size that can be uploaded.
+Defaults to 100 GB to avoid filling the tmp directory of a server.
+
+Warning: this value will overwrite any other values passed to `FLASK_MAX_CONTENT_LENGTH` but is included here to clarify
+its importance when deploying a datalab instance.""",
+    )
+
     @root_validator
     def validate_cache_ages(cls, values):
         if values.get("REMOTE_CACHE_MIN_AGE") > values.get("REMOTE_CACHE_MAX_AGE"):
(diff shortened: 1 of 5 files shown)


jdbocarsly:
Looks good to me, but I would set the max files and filesize much lower, at least for a demo folder. I can't think of a case we currently need to support 10,000 files uploaded at once-- and I imagine the server would not be happy.

jdbocarsly (approved):
lgtm

ml-evs:
> Looks good to me, but I would set the max files and filesize much lower, at least for a demo folder. I can't think of a case we currently need to support 10,000 files uploaded at once-- and I imagine the server would not be happy.

If the 10,000 were tracked as one "file object" in the database then I think it would be fine -- doesn't seem unreasonable once we move to uploading folders of in situ data, or even bruker folders, but take your point for the demo server

Issue #411, opened 2023-06-28
Support for Raman spectroscopy data


ml-evs:
As suggested by @elbee99, it would be great to support Raman spec data as a block. This issue can be used to track progress.

In particular, we would like to support the Renishaw inVia confocal Raman microscope located in the JCMaxwell building. This microscope produces proprietary (?) .wdf files but can also export to simple txt/tsv. I have started to collect these in the private repo at https://github.com/the-grey-group/datalab-file-format-examples until we know whether we can make them public.

- [x] 1D Raman from WDF files (#422)
- [x] 1D Raman from Labspec files (#477)
- [ ] #781

ml-evs:
The first part of this was solved in #422, with support for .wdf files hopefully appearing in a later PR

ml-evs:
Closed by #477

PR #477, opened 2023-10-27
Labspec 1D Raman compatibility

elbee99:
Modified Raman block to make Labspec Raman data exported to .txt compatible and added example data file

- Also loads metadata when provided in the Raman file


pydatalab/pydatalab/apps/raman/blocks.py
```diff
@@ -30,8 +30,10 @@ def load(self, location: str | Path) -> Tuple[pd.DataFrame, List[str]]:
         ext = os.path.splitext(location)[-1].lower()
 
         if ext == ".txt":
-            df = pd.read_csv(location, sep=r"\s+")
-            df = df.rename(columns={"#Wave": "wavenumber", "#Intensity": "intensity"})
+            try:
```
ml-evs, line 33: It would be nice to have some additional expected features of the Raman txt format specifically, as at the moment you could upload any two columns of numbers in a txt file and they would be successfully plotted as Raman data.

Looking at the example data, we could also parse the headers and check for various keys, e.g.,

```python
vendor = None # try to detect vendor first
header = []
with open(location, "r") as f:
    while header_line := f.readline().startswith("#"):
        header.append(header_line)
    
    if "#Wave" in header[0] and "#Intensity"  in header[0]:
       vendor = "renishaw"

    # otherwise try to parse the Labstep header
    else:
        metadata = {key: value for key, value in line.split("=") for line in header}
        if metadata.get("AxisType[0]") == "Intens" and metadata.get("AxisType[1]") == "Spectr":
            vendor = "labstep"
            self.data["metadata"] = metadata

    if not vendor:
        raise RuntimeError("Could not detect Raman data vendor -- this file type is not supported by this block.")
            
```

Let me know if you want to have a go at this, otherwise I can do it!

ml-evs (commented):
Thanks for this @elbee99! One minor comment below:

pydatalab/pydatalab/apps/raman/blocks.py
elbee99: That's fair, will give this a go, should be straightforward

ml-evs (approved):
LGTM, thanks again @elbee99!

ml-evs:
Related to #411

PR #484, opened 2023-10-30
Add ability to register and sign-in via email magic links

### LINKED ISSUE #469: Magic link sign-in
Easier authentication than ORCID/GitHub OAuth. Potential plan: magic link style login tied to contact email. This can hook into our current mechanism for saving/rotating/tombstoning API keys per user (also good for harvesting emails -- not that we want them for anything beyond sending datalab-related alerts in the future). This will also need the bit of UI I was working on in another PR (cannot link) to expose "Login via GitHub/email etc..."

ml-evs:
This PR adds a registration and sign-in mechanism via email and magic links.

Closes #469.

It uses JWT to generate magic links that expire after 1-use/24 hours that are sent to the requested email via an SMTP server (using Flask-Mail). If an account with that email does not exist, it creates the account, otherwise the magic link works for sign-in only.

There is still something slightly broken about getting the page to refresh after authenticating -- hopefully this can be resolved soon.

NB: magic links currently persist forever without being cleaned up, even if they have expired after 24 hours. We should either add a process that cleans them up, or have a way of checking whether it is a true user logging in or a link checking bot.

Related outstanding to-do: with link checkers in place, accounts are erroneously considered verified straight after the email is sent. We need a way to detect whether it is a real request. One option I have seen is to create a dummy JS form that the magic link responds to, and verification occurs only after that form has been automatically submitted by a browser.


pydatalab/pydatalab/email.py
jdbocarsly, line 1: For some reason, I run into an error when running the server which seems to be a namespace conflict of this `email.py` with the standard library `email` (_i.e._ when flask tries to `import email.utils` internally, it fails). Renamining this to `emails.py` and fixing the import elsewhere fixes this.

pydatalab/pydatalab/config.py
```diff
@@ -60,12 +60,17 @@ class RemoteFilesystem(BaseModel):
 class ServerConfig(BaseSettings):
     """A model that provides settings for deploying the API."""
 
-    SECRET_KEY: str = Field(os.urandom(12).hex(), description="The secret key to use for Flask.")
+    SECRET_KEY: str = Field(description="The secret key to use for Flask.")
```
jdbocarsly, line 63: Is this necessary? This breaks the CI since secret key isn't fixed, and also means local installs get a little more complicated because it will fail if you haven't set the secret key.

jdbocarsly: Alternatively, do we have a "template" config file or .env file? Sorry if I've missed it.

ml-evs: > Is this necessary? This breaks the CI since secret key isn't fixed, and also means local installs get a little more complicated because it will fail if you haven't set the secret key.

It kinda is, yeah. Quick reasons: setting a default random value here means that different workers get different keys in deployment, which breaks JWT. Setting a value in the config.json means it should be under version control = bad, and as of now, omitting it from config.json and adding it to .env leads to the default random config taking precedence.

For your local stuff just launch the server with `PYDATALAB_SECRET_KEY="blah"` until we can work out the best way of doing it (IMO just removing it entirely from our Python code would be preferable).

pydatalab/pydatalab/email.py
ml-evs: Oh weird, thats definitely bad, surprised I didn't run into that. I'll push a rename to `sendgrid.py` for now -- we want to move away from this soon anyway and just use SMTP.

pydatalab/pydatalab/config.py
ml-evs: Made the default a system-specific hash of platform info. Not exactly secure but probably pretty good (and better than an openly-available default). Could probably find hard to obtain metadata to add that is shared by all potential flask workers, e.g., server uptime, disk usage by a specific temporary file...

webapp/src/components/GetEmailModal.vue
```diff
@@ -0,0 +1,101 @@
...
+    async submitForm() {
+      this.validateEmail();
+      if (this.emailValidationMessage) {
+        return;
+      }
+      let response = await requestMagicLink(this.emailAddress);
+      if (response.status == "success") {
+        this.emailSent = true;
+        this.successMessage =
+          "Email sent! Please check your inbox for a link to sign in or register.";
```
jdbocarsly, line 59:
```suggestion
          "Email sent! Please check your inbox (and spam folder) for a link to sign in or register. It may take a few minutes to arrive.";
```

jdbocarsly (commented):
This is great. small comments for now. A little rework of the logic around the submit button disabling will make it possible to resubmit/try a different email if something goes wrong with the first submit, but that can wait.

jdbocarsly (dismissed):
approved pending CI

ml-evs:
Okay, going to merge this part of the PR (sans UI) now and debug the test failures elsewhere

Issue #44, opened 2021-08-31
Speed up CI tests


jdbocarsly:
We have discussed some possible ways to streamline the CI tests so that they run faster/use less compute time. Current time was about 10 minutes to check #39.

-Do we really need to build the webapp both inside and outside of the docker container, or should we just do all of the building, linting, and testing inside the docker container. 
- Can we re-organize to make some of the longer steps happen in parallel.

ml-evs:
Caching of the docker build is now working as of #497

PR #497, opened 2023-11-07
CI build optimisations

ml-evs:
This should speed up the CI a little bit by using `docker buildx bake` to do cached builds, plus the cypress GH action to avoid needing to install it.


ml-evs:
Closes #44

PR #512, opened 2023-11-20
Support for Labspec Raman data and metadata parsing 

elbee99:
Have added support for Labspec Raman data in a more specific manner. Have also tried to save the metadata for labspec and Renishaw .wdf files with varying degrees of success.


ml-evs:
Superseded by #477.

PR #529, opened 2024-01-02
Check and warn for missing secrets in server startup logs

ml-evs:
This PR simply loops over some common secrets and warns on startup if they are not present. At some point we could allow blocks to define these secrets dynamically rather than enumerating in main, but this should help with debugging of deployments.

### DIFF
--- a/pydatalab/pydatalab/logger.py
+++ b/pydatalab/pydatalab/logger.py
@@ -14,7 +14,7 @@ class AnsiColorHandler(logging.StreamHandler):
     LOGLEVEL_COLORS = {
         logging.DEBUG: "36m",
         logging.INFO: "32m",
-        logging.WARNING: "33m",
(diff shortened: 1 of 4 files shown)


pydatalab/pydatalab/logger.py
```diff
@@ -59,6 +59,7 @@ def setup_log(log_name: str = "pydatalab", log_level: Optional[int] = None) -> l
     logger.handlers = []
     logger.propagate = False
     handler = AnsiColorHandler()
+    # handler = RichHandler(rich_tracebacks=True, show_path=False, tracebacks_width=80)
```
ml-evs, line 62:
```suggestion
```

ml-evs: Why do the tests pass after removing this comment :cry:

PR #531, opened 2024-01-08
Add "equipment" item type

jdbocarsly:


pydatalab/pydatalab/routes/v0_1/items.py
```diff
@@ -74,6 +74,56 @@ def dereference_files(file_ids: List[Union[str, ObjectId]]) -> Dict[str, Dict]:
...
+        # },
+        "item_id": 1,
+        "name": 1,
+        "nblocks": {"$size": "$display_order"},
+        "characteristic_chemical_formula": 1,
+        "type": 1,
+        "date": 1,
+        "refcode": 1,
+        "location": 1,
+    }
```
ml-evs, line 105:
```suggestion
    _project = {
        "_id": 0,
        "item_id": 1,
        "name": 1,
        "type": 1,
        "date": 1,
        "refcode": 1,
        "location": 1,
    }
```

pydatalab/pydatalab/routes/v0_1/items.py
```diff
@@ -74,6 +74,56 @@ def dereference_files(file_ids: List[Union[str, ObjectId]]) -> Dict[str, Dict]:
     return results
 
 
+def get_equipment_summary():
+    if not current_user.is_authenticated and not CONFIG.TESTING:
+        return (
+            jsonify(
+                status="error",
+                message="Authorization required to access chemical inventory.",
```
ml-evs, line 82:
```suggestion
                message="Authorization required to access equipment list.",
```

pydatalab/pydatalab/routes/v0_1/items.py
```diff
@@ -74,6 +74,56 @@ def dereference_files(file_ids: List[Union[str, ObjectId]]) -> Dict[str, Dict]:
...
+    }
+
+    items = [
+        doc
+        for doc in flask_mongo.db.items.aggregate(
+            [
+                {
+                    "$match": {
+                        "type": "equipment",
+                        **get_default_permissions(user_only=False),
```
ml-evs, line 114: Does equipment need to be private?

jdbocarsly: Good point, we can make equipment open to all users (at least for the time being)

pydatalab/pydatalab/models/equipment.py
```diff
@@ -0,0 +1,20 @@
...
+from pydatalab.models.items import Item
+
+
+class Equipment(Item):
+    """A model for representing an experimental sample."""
+
+    type: str = Field("equipment", const="equipment", pattern="^equipment$")
+
+    serial_numbers: Optional[str]
+    """A string describing one or more serial numbers for the instrument."""
```
ml-evs, line 14: If it really is one or more

```suggestion
    serial_numbers: Optional[list[str]]
    """A string describing one or more serial numbers for the instrument."""
```

pydatalab/pydatalab/routes/v0_1/items.py
```diff
@@ -74,6 +74,45 @@ def dereference_files(file_ids: List[Union[str, ObjectId]]) -> Dict[str, Dict]:
...
+
+    _project = {
+        "_id": 0,
+        "item_id": 1,
+        "name": 1,
+        "type": 1,
+        "date": 1,
+        "refcode": 1,
+        "location": 1,
+    }
```
ml-evs, line 95: I'm wondering if we should move this kind of info to the models themselves, much easier to maintain the list of `summary_fields` as e.g.

```
class Equipment(Item):
    ...
    def _get_summary_fields(self):
        return ("item_id", "name", "type", "date", "refcode", "location")

```
then something like `_project = {f: 1 for f in Equipment._get_summary_fields}`, perhaps always explicitly excluding `_id`.

Should then be straightforward to generate summary routes

ml-evs (commented):
Couple of minor comments

pydatalab/pydatalab/models/equipment.py
jdbocarsly: I had also thought about this as well - the initial reasons for this being a string are similar to the reasons  previously discussed with hazards: its not trivial to make an ergonomic frontend input field for a list, and we just need a field to start keeping track of this stuff now.

But also, in this case, this is a field I envision will be mostly used by humans and could take different forms depending on what's useful (like, just: "XX123" or "XX123 (spectrometer), 789A/B (magnet)"). Making it a list could restrict the flexibilty here, and I don't think it really gives any benefits I can forsee (unlike hazard codes, where we can do cool things like display the pictographs once we properly keep track of them individually).

pydatalab/pydatalab/routes/v0_1/items.py
jdbocarsly: I like this idea. I was thinking about somehow including in each Field() an attribute like `include_in_summary: Bool`, though I wasn't sure how natural that would work with pydantic/ json schemas

jdbocarsly: I recommend we defer this to a separate PR, since it will involve refactoring across the whole app

pydatalab/pydatalab/models/equipment.py
ml-evs: This is probably fair enough, as long as we can make our free text search work with whatever format people provide (e.g., if its comma delimited or otherwise)

ml-evs (commented):
Code wise, this is looking great! 

Dare I suggest adding some API and e2e tests? 😅 

I guess the large outstanding thing to discuss is how to capture the relationships between equipment and items (and blocks) which I know has been started in #565

jdbocarsly:
> Dare I suggest adding some API and e2e tests? 😅

How dare you

pydatalab/pydatalab/models/__init__.py
```diff
@@ -13,6 +14,16 @@
     "samples": Sample,
     "starting_materials": StartingMaterial,
     "cells": Cell,
+    "equipment": Equipment,
```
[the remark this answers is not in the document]
jdbocarsly: oops you're right, I forgot to commit it. fixed now.

.github/workflows/ci.yml
```diff
@@ -156,6 +156,7 @@ jobs:
           spec: |
             ./cypress/e2e/sampleTablePage.cy.js
             ./cypress/e2e/editPage.cy.js
+            ./cypress/e2e/equipment.cy.js
```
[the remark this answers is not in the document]
ml-evs: Oh god, turn it back off!!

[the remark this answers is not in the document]
jdbocarsly: I think I fixed it now 😅. And rebased to squash some commits so it looks like I make fewer mistakes in getting this to work.

So it turned out the batch tests had some of the same errors: within() blocks that ran asynchronously and caused race conditions that only showed up when running the tests in CI. 

I decided most all of those within() blocks weren't needed anyways, and deleted most of them from all our tests- but, it turned out I was a little overzealous and lots of things broke! 

For future reference, cypress commands are always asynchronous, but they are mostly completed in order. `within()` causes this not to be true, leading to issues. We were using `within` to select html nodes inside of other nodes we had located, like this:
```js
cy.get(".vs__dropdown-menu").within(() => {
  cy.contains(".badge", "baseA").click();
});
 ``` 
This is now replaced with simply:
```js
cy.get(".vs__dropdown-menu").contains(".badge", "baseA").click();
```

And the slightly trickier case:
```js
cy.get(".table").within(() => {
  cy.get("td.table-item-id .formatted-item-name").click();
});
```
replaced with:
```js
cy.get(".table").find("td.table-item-id .formatted-item-name").click();
```
Note that in this case the `get()` inside the `within()` block has to be replaced with `find()`. This is because `get()` always searches the whole document when it is not used in within(), while `find()` only searches relative to the node it is chained off.

A general rule is that `within()` probably doesn't need to be used when there is only one command/chain of commands inside its callback. It should be used if multiple actions need to be taken within a selected node.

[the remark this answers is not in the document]
ml-evs: Nice work, thanks for the breakdown

I actually think that now we can run all our specs again, we can remove the spec list both here and for chrome? Should make it easier to introduce more e2e tests

ml-evs (approved):
:guitar:

PR #534, opened 2024-01-10
Added preliminary Raman mapping block

elbee99:
Have created a block for reading Renishaw *.wdf files for Raman mapping experiments. In its current state, the image doesn't appear on the plot where it should, and am not sure why.

### DIFF
--- a/pydatalab/Pipfile.lock
+++ b/pydatalab/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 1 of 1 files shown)


ml-evs (commented):
Hi @elbee99, great work on this so far! Just a couple of comments -- if you are happy with the state of how this is now working, I'm happy to refactor it into the old block myself, then the selection of spectra etc. can be left as a future PR?

PR #535, opened 2024-01-10
remove option to add a "Test Block" on the edit page

jdbocarsly:
The option to add a "Test Block" on the edit page was a holdover from early develoment. It doesn't work at this point, and doesn't serve a purpose, so this PR removes it.
Screen Shot 2024-01-10 at 1 32 44 PM


ml-evs (approved):
🚀

PR #404, opened 2023-05-26
Add UI for collections, fix local graphs and improve tables

ml-evs:
This PR supersedes #292 by taking out the basic collections UI functionality of

- [x] Creating/listing collections
- [x] Adding new items to a collection
- [x] New components for displaying collection names, search etc.
- [x] Some associated UI tweaks

Blocked by #403.

- [x] Fix blocks relative to store changes
- [ ] ~Buttons for adding old items to old collections~ (opened as #423)

PR #553, opened 2024-01-24
Add UI for adding an existing item to an existing collection

### LINKED ISSUE #423: Missing UI for adding/removing an existing item to a collection
I'll work this up in a PR shortly.

ml-evs:
This PR adds a `CollectionSelect` to the sample and cell pages that can be used put items into existing collections. This UI was missing from #404 and closes #423. This required a minor specialisation of the existing collections UI to avoid duplicate listed entries in cases where a sample is already in a collection.

Still missing is a way to add items to a collection from the collections page, as this requires more plumbing on the API side.

### DIFF
--- a/webapp/src/components/FancyCollectionSampleTable.vue
+++ b/webapp/src/components/FancyCollectionSampleTable.vue
@@ -21,14 +21,12 @@
     :items="samples"
     :loading="!sampleTableIsReady"
     :no-hover="true"
-    :checkbox-column-width="40"
     :expand-column-width="40"
     :search-value="searchValue"
     table-class-name="customize-table"
     header-class-name="customize-table-header"
     buttons-pagination
     @click-row="goToEditPage"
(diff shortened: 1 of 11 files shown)


webapp/src/components/FancyCollectionSampleTable.vue
jdbocarsly, line 1: If you remove line 31 (or change `"itemsSelected"` to `"unset"`), then the column of check boxes will go away. That may be a good idea since the checks don't do anything (depending on how long until the next PR that makes them do something is merged)

jdbocarsly (commented):
see below for small comment about the checkboxes. I'm also going to take a crack at one other thing in the collection list on the item pages...

webapp/src/components/FancyCollectionSampleTable.vue
ml-evs: done!

PR #560, opened 2024-02-02
[DRAFT] update install instructions

jdbocarsly:
Fix some issues with the current install instructions

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -43,7 +43,7 @@ The server should now be accessible at [http://localhost:5001](http://localhost:
 
 Should you wish to contribute to/modify the Python code, you may wish to perform these extra steps:
 
-1. From within the `pydatalab` folder, run `pipenv install --dev` to pull the development dependencies (e.g., `pre-commit`, `pytest`).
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -43,7 +43,7 @@ The server should now be accessible at [http://localhost:5001](http://localhost:
 
 Should you wish to contribute to/modify the Python code, you may wish to perform these extra steps:
 
-1. From within the `pydatalab` folder, run `pipenv install --dev` to pull the development dependencies (e.g., `pre-commit`, `pytest`).
+1. From within the `pydatalab` folder, run `pipenv install --dev` to pull the development dependencies (e.g., `pre-commit`, `pytest`). You may also need to install `pre-commit` globally.
```
ml-evs, line 46: I don't think this is the case, but probably you need to run `pipenv run pre-commit install` the first time, which we don't say below...

jdbocarsly: I jotted this down as a draft while we were struggling getting it to run on a clean install. I think the problem was that pre-commit didn't work when `git commit` was run outside of the pipenv. The solution may be to run everything from inside the pipenv.

ml-evs: If you install it the first time inside `pipenv` I think it should work; `pre-commit install` is simply copying a bash script into `.git/hooks/pre-commit` and it "remembers" the Python executable that it was first installed by.

PR #582, opened 2024-02-13
Create pydatalab Python package with dynamic versioning

ml-evs:
This should make releases etc. much easier and is something we will eventually need for the Python API. Just need to see how this interacts with `pipenv` in the Docker containers (probably the version will not be set properly unless we also do a `pip install .`).


ml-evs:
On-hold until we can nicely pass the git-derived version info into docker

PR #588, opened 2024-02-15
CSS cleanup

jdbocarsly:
This PR generally cleans up some of our CSS:

- [x] Move some component styles that were globally in `App.vue` to their specific SFC
- [x] Remove many styles that didn't seem to be in use in several SFCs (mostly because we had copied and pasted from other components, or refactored components and not moved the styles, etc.
- [x] Use `.clickable` as a global style in most cases where `cursor: pointer;` is required.

### DIFF
--- a/webapp/src/App.vue
+++ b/webapp/src/App.vue
@@ -94,17 +94,6 @@ body {
   border-left-color: #d9534f;
 }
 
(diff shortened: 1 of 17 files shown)


ml-evs (approved):
LGTM!

PR #589, opened 2024-02-15
Pin prettier version in Webapp to same version that is specified in pre-commit

jdbocarsly:
Prettier has changed the way it formats code from version-to-version. So, to ensure that the weapp's version of prettier (invoked using `yarn lint`) and pre-commit's don't clash, this pins the version.

Also explicitely specifies `trailingComma: all` in the `.prettierrc` to be explicity about which settings we are using.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,1 +1,2 @@
 printWidth: 100
+trailingComma: all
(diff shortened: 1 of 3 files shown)


ml-evs (approved):
:+1:

PR #592, opened 2024-02-16
Add a Tabular data block that can handle simple CSVs and text files

ml-evs:
This PR adds a simple block that passes a data file through `pd.read_csv()` and allows you to make a selectable  scatter plot of its columns. It also adds a base component `BokehBlock` that can probably be used by several other blocks with minor tweaking of the supported extensions (which should be dynamic soon anyway).

We could make this increasingly useful by being more robust with `read_csv` args and maybe wrapping our own file reader. e.g., I took some random SECM data and realised that the header is formatted in such a way that pandas fails to read it. I think one nice idea could be to write a wrapper to `read_csv` that does a binary search to detect the header 'properly', or b) reads a CSV file in reverse, populates the values until the first "broken" line, then treats them as column headers (this could actually be very useful generally, e.g., within pandas itself).


jdbocarsly:
Awesome!! This is going to be super useful. 

Already works for me for a basic csv file, but some thoughts on possible future developments:

I agree that a UI to set the parameters for read_csv would be great. I've been thinking of something with live feedback (i.e. you can see the file, and see how changing the parameters splits it up differently, like some spreedsheet applications can do). That's probably a longer term goal, but for now, just having options for `sep`, `skip_rows`, and `comment` would probably work for a majority of cases. (your idea of reading csvs backwards is interesting too...)

Adding a UI for naming/renaming columns and potentially dropping spurious columns would be the other useful feature, but would take more work. Add a button to open in google drive??

Incidentally, by adding `sep=None` to the `read_csv` call, pandas automatically tries to guess the separator using `csv.Sniffer().sniff()`. I tested this out, and it worked pretty well and let me read a variety of files including some .csv cycling data and some .xye synchrotron data (tab-delimited). If we want to add the UI for specifying skip_rows, etc., we can explicitely use Sniffer to guess starting parameters.

jdbocarsly:
A few more immediate comments:

- When I tried an invalid file, it through an error and make it so that I couldn't reload the sample ever again without erroring.
- Unfortunately our `selectable_axis_plot` doesn't save the selected columns anywhere, so you have to reselect the correct columns every time you load the page. I wonder if we can figure out an elegant way to save those. Honestly, it would probably involve moving the `select` components out of bokeh and generating them directly in vue so that they behave like all the other inputs in a block.
- Can you rename this to something like "CSVBlock", instead of "PlotterBlock"? The other blocks are named by what kind of data they deal with, rather than how they display the data.

pydatalab/pydatalab/blocks/common.py
```diff
@@ -47,3 +49,59 @@ def encode_tiff(self):
                 self.data["b64_encoded_image"][self.data["file_id"]] = base64.b64encode(
                     f.getvalue()
                 ).decode()
+
+
+class PlotterBlock(DataBlock):
```
jdbocarsly, line 54:
```suggestion
class CSVBlock(DataBlock):
```

webapp/src/resources.js
```diff
@@ -39,6 +40,7 @@ export const debounceTime = 250; // time after user stops typing before request
 export const blockTypes = {
   comment: { description: "Comment", component: DataBlockBase },
   media: { description: "Media", component: MediaBlock },
+  plotter: { description: "Simple plot", component: BokehBlock },
```
jdbocarsly, line 43:
```suggestion
  plotter: { description: "CSV file", component: BokehBlock },
```

webapp/src/components/datablocks/BokehBlock.vue
jdbocarsly, line 1: I guess the idea is that this is a generic plotting component, which could be re-used for any other blocks that just need a plot?

pydatalab/pydatalab/blocks/common.py
```diff
@@ -47,3 +49,59 @@ def encode_tiff(self):
...
+        if "file_id" not in self.data:
+            warnings.warn(
+                f"{self.__class__.__name__}.read_data_file(): No file set in the DataBlock"
+            )
+            return
+
+        file_info = get_file_info_by_id(self.data["file_id"], update_if_live=True)
+
+        try:
+            return pd.read_csv(file_info["location"], skip_blank_lines=False)
```
jdbocarsly, line 78:
```suggestion
            return pd.read_csv(file_info["location"], sep=None, skip_blank_lines=False)
```

jdbocarsly (commented):
some thoughts

webapp/src/components/datablocks/BokehBlock.vue
ml-evs: Yeah potentially, I tried to rebase the Raman block on it but got stuck trying to do inheritance with the components listed in `resources.js`, e.g., something like

`'raman': {'component': defineComponent(extends: bokehBlock, ...)}` (modulo correct syntax) but it didn't seem particularly ergonomic

ml-evs:
Should be merged after #590

pydatalab/pydatalab/blocks/common.py
```diff
@@ -49,3 +51,53 @@ def encode_tiff(self):
...
+
+
+class TabularDataBlock(DataBlock):
+    """This block simply tries to read the given file with pandas, and
+    expose an interface to plot its columns as scatter points.
+
+    """
+
+    blocktype = "tabular"
+    description = "Tabular Data Block"
```
ml-evs, line 63:
```suggestion
    name = "Tabular Data Block"
    description = "This block will load tabular data from common plain text files and allow you to create simple scatter plots of the columns within."
```

ml-evs:
I think this is good enough for a first hash, would be good to get in before 0.4.0

pydatalab/pydatalab/blocks/common.py
```diff
@@ -49,3 +51,54 @@ def encode_tiff(self):
...
+        if "file_id" not in self.data:
+            warnings.warn(
+                f"{self.__class__.__name__}.read_data_file(): No file set in the DataBlock"
+            )
+            return
+
+        file_info = get_file_info_by_id(self.data["file_id"], update_if_live=True)
+
+        try:
+            return pd.read_csv(file_info["location"], sep=None, skip_blank_lines=False)
```
jdbocarsly, line 81:
```suggestion
            return pd.read_csv(file_info["location"], sep=None, skip_blank_lines=False, engine="python")
```

jdbocarsly (changes requested):
Works well, but I get a few unecessary warning displays when using:

1. When making a new block, I get the following warning before starting:
Screenshot 2024-05-19 at 3 48 05 PM
2. When loading a sample.
Screenshot 2024-05-19 at 3 47 58 PM

ml-evs:
> Works well, but I get a few unecessary warning displays when using:
> 
>     1. When making a new block, I get the following warning before starting:
> 
> 
> Screenshot 2024-05-19 at 3 48 05 PM 2. When loading a sample. Screenshot 2024-05-19 at 3 47 58 PM

I've fixed these warnings, and also tweaked the pandas settings so that we can at least read the awkward Raman text files in our repo and a simple csv (with tests). Obviously there's no global solution for this unless we build out a whole UI, but think this should at least fail nicely...

ml-evs:
I'm going to merge this so we can get a feel for how useful it really is in production, having address @jdbocarsly's immediate comments

PR #593, opened 2024-02-16
Downgrade pre-commit prettier

ml-evs:
Related to #589; downgrades pre-commit prettier to use the same version as package.json.


jdbocarsly (commented):
👍

Issue #168, opened 2022-09-24
Python version, Dependabot and pipenv conflicts


ml-evs:
Some fun, pipenv does not allow you to broadly specify a range of Python versions (e.g., ">3.8"), only "3" or something specific.
By default, dependabot uses a recent Python (3.10), which causes the lockfile to break for all versions that have different requirements between 3.8 and 3.10

Two possible solutions:
- Pin dependabot to the lowest supported Python version using `.python_version` or `runtime.txt` files (see https://github.com/dependabot/dependabot-core/issues/1455 - still would be quite brittle).
- Only support one Python version for this project (would simplify quite a few things tbh, but could be restrictive on which dependencies we can use (e.g., if they don't support Python 3.10 themselves)).
- Scrap pipenv!

ml-evs:
Closed by #604

PR #604, opened 2024-02-19
Repackage server as proper Python package, removing `pipenv`

ml-evs:
Supersedes #582 and addresses #168

Moves around some config files and expands `pyproject` to be the single source of truth, with a dynamic version from the git tags. Also hacks the docker build so that this version info gets passed in production, and switches to using `uv` inside the docker builds.

For now, I'd like this PR to retain `Pipfile` and `Pipfile.lock` so that existing developer workflows should still work. In the future, we will switch to purely using `uv`-generated lockfiles for dependencies, which will allow us to use dependabot again properly, and make our dependency management less fraught with issues when wanting to update only one dep.

Warning: the deps in `Pipfile.lock` will not exactly match the `uv` generated lock files.


ml-evs:
Decision is to merge this PR immediately after the next release, to avoid breaking deployments.

INSTALL.md
```diff
@@ -21,9 +27,11 @@ Firstly, from the desired folder, clone this repository from GitHub to your loca
 ### `pydatalab` server installation
 
 The instructions in this section will leave you with a running JSON API on your host machine.
-This can hypothetically be used entirely independently from the web front-end through the JSON API.
+This can hypothetically be used entirely independently from the web front-end.
 
-1. Install `pipenv` on your machine.
+1. Install `pipenv` on your machine (optional).
+    - Detailed instructions for installing `pipenv`, `pip` and Python itself can be found on the [`pipenv` website](https://pipenv.pypa.io/en/latest/install/#installing-pipenv).
+1. Install `pipenv` on your machine (optional).
```
ml-evs, line 34:
```suggestion
```

INSTALL.md
```diff
@@ -21,9 +27,11 @@ Firstly, from the desired folder, clone this repository from GitHub to your loca
...
     - Detailed instructions for installing `pipenv`, `pip` and Python itself can be found on the [`pipenv` website](https://pipenv.pypa.io/en/latest/install/#installing-pipenv). You will need Python 3.10 or higher to run pydatalab.
```
ml-evs, line 35:
```suggestion
1. Install `pipenv` on your machine (optional).
    - Detailed instructions for installing `pipenv`, `pip` and Python itself can be found on the [`pipenv` website](https://pipenv.pypa.io/en/latest/install/#installing-pipenv). You will need Python 3.10 or higher to run pydatalab.
```

PR #606, opened 2024-02-19
Reset unsaved warning after `getSampleData()` receives data

### LINKED ISSUE #601: "Unsaved changes" warning appears on loading a sample a second time, after using the browser back button
The "unsaved warning" button and "are you sure you want to leave?" confirmation dialogue should only appear when something has actually been changed. However, in some cases, it shows up before any changes have been made. 

Steps to reproduce:
1. Start on the sample table page
2. Load a sample or cell
3. Press the browser "back" button
4. Load the same sample/cell again.
 -> "Unsaved changes" warning appears.
6. Try to leave this page
 -> "are you sure you want to leave?" dialogue comes up.

Previous work on this behavior: #223  #257

jdbocarsly:
Simple fix to prevent the annoying bug where the unsaved warning erroneously appears whenever a sample or cell page is loaded a second time without a refresh.

closes #601

### DIFF
--- a/webapp/src/views/EditPage.vue
+++ b/webapp/src/views/EditPage.vue
@@ -136,6 +136,7 @@ export default {
       next();
     } else {
       if (window.confirm("Unsaved changes present. Would you like to leave without saving?")) {
+        this.$store.commit("setItemSaved", { item_id: this.item_id, isSaved: true });
         next();
       } else {
         next(false);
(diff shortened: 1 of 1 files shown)


jdbocarsly:
@ml-evs , can you deploy this to dev when you get a chance? Would like to test it on a real server with latency, etc.

ml-evs:
> @ml-evs , can you deploy this to dev when you get a chance? Would like to test it on a real server with latency, etc.

Deployed on dev

jdbocarsly:
@ml-evs @be-smith 

Ok this is ready for review. The approach I took here is just to reset the warning on reload, which is the simplest approach. Note that we still don't clear the store. This may have pluses and minuses that could be considered for other behavior, but the unsaved warning issue seems to be fixed.

ml-evs (approved):

PR #631, opened 2024-03-08
Update xrdml parser so it works with v2.0

jdbocarsly:
The XRDML format apparently changed at some point from v1.5 to v2.0. In 2.0, the `<intensities unit="counts">` tag has been changed to `<counts unit="counts">`. This pull request updates the regex in the xrdml parser so that it looks for either "intensities" or "counts". It also adds a file with the new format to the `example_data` directory for testing.


ml-evs (approved):
LGTM, thanks!

PR #635, opened 2024-03-08
Allow browser context menu in tinymce fields

### LINKED ISSUE #633: Add ability to copy and paste using right-click in TinyMCE fields

jdbocarsly:
Currently, it is not possible to copy or paste from the right click context window in our tinymce fields. Tinymce has a copy/paste functionality, but apparently it is blocked in many browsers. So, the best option for now seems to be to not use tinymce's contextmenu at all, and instead enable the default browser right click.

closes #633

### DIFF
--- a/webapp/src/components/TinyMceInline.vue
+++ b/webapp/src/components/TinyMceInline.vue
@@ -13,6 +13,7 @@
         width: '50%',
         'margin-left': '1rem',
       },
+      contextmenu: false,
       toolbar:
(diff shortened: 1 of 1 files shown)


ml-evs (approved):

ml-evs:
E2E errors might be real this time?

```
The following error originated from your application code, not from Cypress.

  > Unexpected token '<'

When Cypress detects uncaught errors originating from your application it will automatically fail the current test.

This behavior is configurable, and you can choose to turn this off by listening to the uncaught:exception event.

https://on.cypress.io/uncaught-exception-from-application
```

ml-evs (approved):
What could go wrong...

PR #642, opened 2024-03-12
Add optional functionality to add starting materials from within the webapp

### LINKED ISSUE #648: Starting Material Table gives console warnings for "formattedItemNameMaxLength"
We get lots of console warnings on the Starting Material page

```
[Vue warn]: Property "formattedItemNameMaxLength" was accessed during render but is not defined on instance.
```

### LINKED ISSUE #654: No way to delete starting materials
In #642, the ability to add starting materials was added. However, I forgot to add a way to delete starting materials.

jdbocarsly:
- [x] Add "Create Item" Modal to Sample table page
- [x] Make backend tweaks to allow Starting Materials to be added from the app
- [x] Make starting material fields editable in the frontent (dependent on a configuration variable)
- [x]  Refactor `<inputs>` to a custom `<EditableInput>` field, which will clean up the code and make it easier to do read-only versions of other items in the future
- [x] Add ability to filter types allowed in `CreateItemModal.vue`
- [x] Additional field: CAS, hazards

Also closes #648 and closes #654

### DIFF
--- a/pydatalab/pydatalab/models/starting_materials.py
+++ b/pydatalab/pydatalab/models/starting_materials.py
@@ -31,6 +31,12 @@ class StartingMaterial(Item):
 
     full_percent: Optional[str] = Field(alias="Full %")
 
+    GHS_codes: Optional[str] = Field(
+        alias="GHS H-codes",
+        description="A string describing any GHS hazard codes associated with this item. See https://pubchem.ncbi.nlm.nih.gov/ghs/ for code definitions.",
+        examples=["H224", "H303, H316, H319"],
+    )
+
     name: Optional[str] = Field(alias="Container Name", description="name of the chemical")
 
     size: Optional[str] = Field(
--- a/pydatalab/pydatalab/models/starting_materials.py
+++ b/pydatalab/pydatalab/models/starting_materials.py
@@ -17,7 +17,7 @@ class StartingMaterial(Item):
         alias="Barcode", description="A unique barcode from ChemInventory"
     )
 
-    date_acquired: Optional[IsoformatDateTime] = Field(
+    date: Optional[IsoformatDateTime] = Field(
         alias="Date Acquired", description="The date the item was acquired"
     )
 
--- a/pydatalab/pydatalab/models/starting_materials.py
+++ b/pydatalab/pydatalab/models/starting_materials.py
@@ -31,7 +31,7 @@ class StartingMaterial(Item):
 
     full_percent: Optional[str] = Field(alias="Full %")
 
-    name: str = Field(alias="Container Name", description="name of the chemical")
+    name: Optional[str] = Field(alias="Container Name", description="name of the chemical")
 
     size: Optional[str] = Field(
         alias="Container Size", description="size of the container (see 'size_unit' for the units)"
--- a/webapp/src/components/StartingMaterialInformation.vue
+++ b/webapp/src/components/StartingMaterialInformation.vue
@@ -1,73 +1,57 @@
 <template>
-  <div class="container">
+  <div class="container-lg px-5 px-lg-0">
     <!-- Sample information -->
     <div id="starting-material-information" class="form-row">
-      <div class="form-group col-md-2 col-sm-4">
-        <label for="item_id" class="mr-2">Refcode</label>
+      <div class="form-group col-md-2 col-sm-3 col-sm-4">
+        <label for="item_id">Refcode</label>
         <div><FormattedRefcode :refcode="Refcode" /></div>
       </div>
-      <div class="form-group col-md-2 col-sm-4">
-        <label for="item_id" class="mr-2">Barcode</label>
-        <input id="item_id" class="form-control-plaintext" readonly="true" :value="item_id" />
+      <div class="form-group col-md-2 col-sm-3 col-sm-4">
+        <label for="item_id">Item ID</label>
+        <StyledInput id="item_id" readonly :modelValue="ItemID" />
       </div>
-      <div class="form-group col-md-6 col-sm-8">
-        <label for="name" class="mr-2">Name</label>
-        <input id="name" :value="item.name" class="form-control-plaintext" readonly="true" />
-      </div>
-      <div class="form-group col-md-2 col-sm-4">
-        <label for="date-opened" class="mr-2">Date acquired</label>
-        <input
-          id="date-opened"
-          :value="$filters.IsoDatetimeToDate(item.date_acquired)"
-          class="form-control-plaintext"
-          readonly="true"
-        />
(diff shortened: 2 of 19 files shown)


webapp/src/components/StartingMaterialInformation.vue
```diff
@@ -84,9 +79,13 @@
...
 import TinyMceInline from "@/components/TinyMceInline";
 import ChemicalFormula from "@/components/ChemicalFormula";
+import ChemFormulaInput from "@/components/ChemFormulaInput";
 import TableOfContents from "@/components/TableOfContents";
-import CollectionList from "@/components/CollectionList";
+import ToggleableCollectionFormGroup from "@/components/ToggleableCollectionFormGroup";
 import FormattedRefcode from "@/components/FormattedRefcode";
+import StyledInput from "@/components/StyledInput";
+
+import { IS_STARTING_MATERIAL_EDITABLE } from "@/resources.js";
```
ml-evs, line 88: Must say that I don't really like this; would prefer it to be added as config on the server side that the app can query in the style of #455 (eventually I'd like all "app" config except the API URL to be made server metadata).

Does it really need to be configurable? The other option would be that admins can always do it, and you only show the button if they are an admin...

jdbocarsly: I'll make it configurable from an environment variable now, since that's how all the other webapp config is done. A clean way of storing the specifying the app config is definitely something we should.

As for why this needs to be configurable: groups that are importing their inventory from a 3rd party chemical inventory manager need it to be "false" in most cases, to avoid the two inventories getting out of sync. Groups that want to use datalab as their chemical inventory will need it to be "true". If we add more roles like "inventory manager" in the future, it will be easy to fine-grain who can edit these fields.

jdbocarsly: Ok, now its an environment variable with default "false". Name has been changed to "EDITABLE_INVENTORY"

pydatalab/pydatalab/models/starting_materials.py
```diff
@@ -31,7 +31,9 @@ class StartingMaterial(Item):
 
     full_percent: Optional[str] = Field(alias="Full %")
 
-    name: str = Field(alias="Container Name", description="name of the chemical")
+    GHS_codes: Optional[str] = Field(alias="GHS H-codes")
```
ml-evs, line 34: Is a simple string enough here? Should it at the very least be a list of strings? In terms of making this "actionable" it would be good to have a GHS  H-code type/enum so that we can show proper warnings in the future.

This would however involve some extra validation and a better feedback loop between UI and server for this...

webapp/src/components/StartingMaterialInformation.vue
```diff
@@ -1,73 +1,70 @@
...
+        <StyledInput id="location" v-model="Location" :readonly="!isEditable" />
+      </div>
     </div>
 
-    <label for="location" class="mr-2">Location</label>
-    <input id="location" :value="item.location" class="form-control-plaintext" readonly="true" />
+    <div class="form-row">
+      <div class="form-group col-lg-3 col-sm-4">
+        <label for="cas">CAS</label>
+        <StyledInput id="cas" v-model="CAS" :readonly="!isEditable" />
```
ml-evs, line 58: Not for this PR but I'd like CAS to be something similar to the toggleable form; its a string field that will rarely have to change and it should do something "special" when locked in, e.g., link to a CAS registry search for that item

webapp/src/components/StartingMaterialInformation.vue
```diff
@@ -1,73 +1,70 @@
...
+      </div>
+      <div class="form-group col-lg-3 col-sm-4">
+        <label for="hazards">GHS Hazard Codes</label>
+        <StyledInput id="hazards" v-model="GHS" :readonly="!isEditable" />
```
ml-evs, line 62: Likewise, here, validating server side that the code valid, we should be able toggle editing and show the proper hazard badges

ml-evs (changes requested):
Mostly comments about future improvements, but wonder if the GHS codes model should be firmed up now (and then just modified slightly in the UI).

e.g., for now we could just accept comma-separated codes in the UI which are split into a list.

Once the Python test for this is in then I think this is otherwise good-to-go!

pydatalab/pydatalab/models/starting_materials.py
ml-evs: Also a description would be nice, with reference to https://pubchem.ncbi.nlm.nih.gov/ghs/, perhaps a couple of examples of what this field could look like in an `examples` field

jdbocarsly: I agree, but have left this as a string for now because we don't have an easy way of showing lists in the frontend at the moment, except maybe a multiselect. I would like use to do things like auomatically show the pictograms and automatically populate the hazards based on CAS in the future, but in my opinion its better to have a string field for now than nothing. I will add a description, though.

webapp/src/components/StartingMaterialInformation.vue
jdbocarsly: agreed. I would also like to have a way of automatically populating many of these fields based on CAS or another code, but will leave that for the future.

webapp/src/components/StartingMaterialInformation.vue
jdbocarsly: yep... project for a future PR. Wil make some issues for these future ideas for inventory management.

pydatalab/pydatalab/models/starting_materials.py
jdbocarsly: issue #657

webapp/src/components/StartingMaterialInformation.vue
jdbocarsly: issue #658

webapp/src/components/StartingMaterialInformation.vue
jdbocarsly: Issue #657

pydatalab/pydatalab/models/starting_materials.py
jdbocarsly: That would work for visualizing, but not for entering the codes in the first place through the UI

jdbocarsly: It will also be an issue in cheminventory xlsx imports, which appear to be strings of space-separated codes (e.g. "H260 H314")

BenjaminCharmes (approved):

ml-evs (approved):
LGTM, can continue the discussion on hazards at #657

Issue #640, opened 2024-03-11
Validate display names and contact emails


ml-evs:
Now that display names are editable (after #627 is merged, at least), we should adopt sensible limits on `display_name` and enforce them when updating a user (cc @BenjaminCharmes). This might mean adding a validator to the pydantic models, and then making sure any user updates are passed through the pydantic models. What we really need is some kind of partial validation; perhaps directly a `DisplayName` type (and also email, which is even easier) would help here.

ORCID for example does not allow names longer than 150 chars.

PR #627, opened 2024-03-08
[WIP] Created component for editing account settings

BenjaminCharmes:
Frontend only for now:

- Displaying current user `display_name`, but need to create a route for editing it.

- The buttons for connecting/disconnecting from Github and ORCHID exist but are not currently working either

PR #646, opened 2024-03-13
[WIP] Added validation for the model user

BenjaminCharmes:
Linked to #640 

Frontend:
- Add error message and disabled submitting if display_name or contact_email isn't validated

Backend:
- For more 'security', also add validator on the user model and in the save_user method so a user can't be updated with unvalidated data even if the form is already disabled.

### DIFF
--- a/pydatalab/pydatalab/models/people.py
+++ b/pydatalab/pydatalab/models/people.py
@@ -3,6 +3,7 @@ from typing import List, Optional
 
 import bson
 import bson.errors
+import re
 from pydantic import BaseModel, EmailStr, Field, validator
 
 from pydatalab.models.entries import Entry
@@ -90,6 +91,22 @@ class Person(Entry):
     @validator("type", pre=True)
     def set_default_type(cls, _):
         return "people"
+
+    @validator("display_name")
+    def validate_display_name_length(cls, v):
+        """Validate that the display name."""
+        if len(v) > 150:
+            raise ValueError(
+                "Display name must be at most 150 characters long.")
+        return v
+
+    @validator("contact_email")
+    def validate_contact_email_format(cls, v):
+        """Validate that the contact email has a valid email format."""
+        email_regex = re.compile(r"[^@]+@[^@]+\.[^@]+")
+        if v is not None:
+            not email_regex.match(EmailStr(v))
+        return v
 
     @staticmethod
     def new_user_from_identity(
--- a/pydatalab/pydatalab/routes/v0_1/users.py
+++ b/pydatalab/pydatalab/routes/v0_1/users.py
@@ -1,10 +1,8 @@
 from bson import ObjectId
 from flask import Blueprint, jsonify, request
-
 from flask_login import current_user
 
 from pydatalab.mongo import flask_mongo
-
 
 user = Blueprint("users", __name__)
 
@@ -20,8 +18,12 @@ def save_user(user_id):
         return jsonify(status="error"), 403
 
     update_result = flask_mongo.db.users.update_one(
-        {"_id": ObjectId(user_id)}, {"$set": {"display_name": user_name}})
+        {"_id": ObjectId(user_id)}, {"$set": {"display_name": user_name}}
+    )
     if update_result.nmodified != 1:
-        return jsonify(status="error", detail="User does not have the appropriate permissions to update the given user ID."), 403
+        return jsonify(
+            status="error",
+            detail="User does not have the appropriate permissions to update the given user ID.",
+        ), 403
 
     return jsonify(status="success"), 200
--- a/pydatalab/pydatalab/routes/v0_1/users.py
+++ b/pydatalab/pydatalab/routes/v0_1/users.py
@@ -11,6 +11,13 @@ def save_user(user_id):
     user_name = request.json.get("data", {}).get("user_name")
 
     user_id_object = ObjectId(user_id)
-    flask_mongo.db.users.update_one({"_id": user_id_object}, {"$set": {"display_name": user_name}})
+    update_result = flask_mongo.db.users.update_one(
+        {"_id": user_id_object}, {"$set": {"display_name": user_name}}
+    )
+    if update_result.nmodified != 1:
+        return jsonify(
+            status="error",
+            detail="User does not have the appropriate permissions to update the given user ID.",
+        ), 403
 
     return jsonify(status="success"), 200
--- a/pydatalab/pydatalab/routes/v0_1/users.py
+++ b/pydatalab/pydatalab/routes/v0_1/users.py
@@ -1,0 +1,20 @@
+from flask import Blueprint, jsonify, request
+from flask_login import current_user
+from bson import ObjectId
+
+from pydatalab.mongo import flask_mongo
+
+user = Blueprint("users", __name__)
+
+
+@user.route('/users/<user_id>', methods=['PATCH'])
+def save_user(user_id):
+    
+    user_name = request.json.get('data', {}).get('user_name')
+
+    user_id_object = ObjectId(user_id)
+    flask_mongo.db.users.update_one({"_id": user_id_object}, {'$set': {'display_name': user_name}})
+
+    return jsonify(status="success"), 200
+
+
--- a/pydatalab/pydatalab/routes/v0_1/users.py
+++ b/pydatalab/pydatalab/routes/v0_1/users.py
@@ -1,7 +1,10 @@
 from bson import ObjectId
 from flask import Blueprint, jsonify, request
 
+from flask_login import current_user
+
 from pydatalab.mongo import flask_mongo
+
 
 user = Blueprint("users", __name__)
 
@@ -10,14 +13,15 @@ user = Blueprint("users", __name__)
 def save_user(user_id):
     user_name = request.json.get("data", {}).get("user_name")
 
-    user_id_object = ObjectId(user_id)
+    if not current_user.authenticated:
+        return jsonify(status="error"), 401
+
+    if current_user.id != user_id and current_user.role != "admin":
+        return jsonify(status="error"), 403
+
     update_result = flask_mongo.db.users.update_one(
-        {"_id": user_id_object}, {"$set": {"display_name": user_name}}
-    )
+        {"_id": ObjectId(user_id)}, {"$set": {"display_name": user_name}})
     if update_result.nmodified != 1:
-        return jsonify(
-            status="error",
-            detail="User does not have the appropriate permissions to update the given user ID.",
-        ), 403
+        return jsonify(status="error", detail="User does not have the appropriate permissions to update the given user ID."), 403
 
     return jsonify(status="success"), 200
--- a/pydatalab/pydatalab/routes/v0_1/users.py
+++ b/pydatalab/pydatalab/routes/v0_1/users.py
@@ -9,18 +9,23 @@ user = Blueprint("users", __name__)
 
 @user.route("/users/<user_id>", methods=["PATCH"])
 def save_user(user_id):
-    user_name = request.json.get("data", {}).get("user_name")
+    request_json = request.get_json()
 
-    if not current_user.authenticated:
+    display_name = request_json.get("display_name")
+    contact_email = request_json.get("contact_email")
+
+    if not current_user.is_authenticated:
         return jsonify(status="error"), 401
 
     if current_user.id != user_id and current_user.role != "admin":
         return jsonify(status="error"), 403
 
     update_result = flask_mongo.db.users.update_one(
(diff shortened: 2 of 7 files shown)


pydatalab/pydatalab/routes/v0_1/users.py
```diff
@@ -0,0 +1,34 @@
...
+    request_json = request.get_json()
+
+    display_name = request_json.get("display_name")
+    contact_email = request_json.get("contact_email")
+
+    if not current_user.is_authenticated:
+        return jsonify(status="error"), 401
+
+    if current_user.id != user_id and current_user.role != "admin":
+        return jsonify(status="error"), 403
```
ml-evs, line 21: Now that the `Person` model has additional validation, we need to make sure the data is piped through the model here before updating the database. One way to do that would be to pull the existing user data from the db, update it with the new data, then create a model. If it passes the model, then update the db.

Alternatively, we could just individually validate the two fields considered here by defining specialist types for email addresses and display names, then use those in the `Person` model directly (see other comment)

pydatalab/pydatalab/models/people.py
```diff
@@ -91,6 +92,21 @@ def add_missing_type(cls, v):
     def set_default_type(cls, _):
         return "people"
 
+    @validator("display_name")
+    def validate_display_name_length(cls, v):
+        """Validate that the display name."""
+        if len(v) > 150:
+            raise ValueError("Display name must be at most 150 characters long.")
+        return v
```
ml-evs, line 100: Another approach (related to other comment) would be to define a custom type called `DisplayName` (or something) that is just a constrained string less than 150 chars long. We can then directly validate strings with it, without having to create the entire `Person` model

pydatalab/pydatalab/models/people.py
```diff
@@ -91,6 +92,21 @@ def add_missing_type(cls, v):
     def set_default_type(cls, _):
         return "people"
 
+    @validator("display_name")
+    def validate_display_name_length(cls, v):
+        """Validate that the display name."""
+        if len(v) > 150:
+            raise ValueError("Display name must be at most 150 characters long.")
```
ml-evs, line 99:
```suggestion
        max_len = 150
        if len(v) > max_len:
            raise ValueError(f"Display name must be at most {max_len} characters long.")
```

Minor tweak to make it easier to update

pydatalab/pydatalab/models/people.py
```diff
@@ -91,6 +92,21 @@ def add_missing_type(cls, v):
...
+        return v
+
+    @validator("contact_email")
+    def validate_contact_email_format(cls, v):
+        """Validate that the contact email has a valid email format."""
+        email_regex = re.compile(r"[^@]+@[^@]+\.[^@]+")
+        if v is not None:
+            not email_regex.match(EmailStr(v))
+        return v
```
ml-evs, line 108: I think this one is actually unnecessary because the `contact_email` field in the model already has the pydantic type `EmailStr` (which defines a regexp too). In this case, we could use this type directly in the user route to check that we are updating the db with an email.

pydatalab/pydatalab/routes/v0_1/users.py
```diff
@@ -0,0 +1,34 @@
...
+
+    update_result = flask_mongo.db.users.update_one(
+        {"_id": ObjectId(user_id)},
+        {"$set": {"display_name": display_name, "contact_email": contact_email}},
```
ml-evs, line 25: Following the comments above, it could look something like this:

```suggestion
        {"$set": {"display_name": DisplayName(display_name), "contact_email": EmailStr(contact_email)}},
```

maybe with a try/except to give a clear error message in case of failure.

ml-evs (commented):
Some minor immediate comments. Reviewing out of order as @jdbocarsly is going to also take a look at #627 first (which we will merge before this one).

Looks good so far, we can just discuss the right approach.

Although it seems a bit gross to have validation on both the client and server, I think it makes sense in this case since it really is easy to validate. One thing that might be missing is how the UI deals with a server-side validation, so you might want to test disabling the client side validation (e.g., making the display name max length on the client 151 and then providing a 151 char display name to the server to make sure that a nice error is returned).

pydatalab/pydatalab/models/people.py
```diff
@@ -62,6 +62,17 @@ def add_missing_verification(cls, v):
...
 
+class DisplayName(str):
+    """A constrained string less than 150 characters long."""
+
+    max_length = 150
+
+    def __new__(cls, value):
+        if len(value) > cls.max_length:
+            raise ValueError(f"Display name must be at most {cls.max_length} characters long.")
+        return str.__new__(cls, value)
```
ml-evs, line 73: We should also check for emptiness, and that all characters are in an alphabet.

Could you add some tests in the pure Python for this? e.g., passing a series of valid names and testing that they work, and a series of invalid names and checking that they don't? I can help talk you through the test set up process if need be

webapp/src/components/EditAccountSettingsModal.vue
```diff
@@ -103,6 +111,21 @@ export default {
     modelValue: Boolean,
   },
   emits: ["update:modelValue"],
+  computed: {
+    displayNameValidationMessage() {
+      if (this.user.display_name && this.user.display_name.length > 150) {
+        return "Name should be less than 150 characters.";
+      }
+      return "";
+    },
```
ml-evs, line 120: We should double check how this looks to the user when they put in something that fails at the API level

ml-evs (commented):
This looks good to me, thanks Ben! Just one comment about the display name (below) and the suggestion that we should add a simple test that the validator does what we expect.

pydatalab/pydatalab/models/people.py
```diff
@@ -62,6 +62,27 @@ def add_missing_verification(cls, v):
...
+
+    max_length = 150
+
+    def __new__(cls, value):
+        if not value.strip():
+            raise ValueError("Display name cannot be empty.")
+
+        if not all(char.isalpha() or char.isspace() for char in value):
+            raise ValueError("Display name can only contain alphabetic characters and spaces.")
+
```
ml-evs, line 79: These are fine for now, might be easier/more performant to write these constraints as a regexp instead (that we could then duplicate in the UI). Just a note to myself!

pydatalab/pydatalab/routes/v0_1/users.py
```diff
@@ -22,13 +24,13 @@ def save_user(user_id):
...
+        update["display_name"] = DisplayName(display_name)
+
     if contact_email or contact_email in (None, ""):
         if contact_email == "":
             contact_email = None
-        update["contact_email"] = contact_email
-
-    if display_name:
-        update["display_name"] = display_name
+        update["contact_email"] = EmailStr(contact_email)
```
ml-evs, line 33: I think we should probably catch and return `ValueError` on both of these calls, then simply alert whether it is the display name or contact email that is "bad" -- this alert can then be shown in the UI.

I think this is the one change that we should get in before merging, as at the moment I think you just get an incomprehensible message back

pydatalab/pydatalab/models/people.py
```diff
@@ -62,6 +62,27 @@ def add_missing_verification(cls, v):
...
+    that cannot be empty and should contain only alphabetic characters and spaces.
+    """
+
+    max_length = 150
+
+    def __new__(cls, value):
+        if not value.strip():
+            raise ValueError("Display name cannot be empty.")
+
+        if not all(char.isalpha() or char.isspace() for char in value):
```
ml-evs, line 77: We need to allow hyphens and apostrophes through in names too (maybe others?) I actually don't know how e.g., a name with Chinese characters would be best handled here

pydatalab/pydatalab/models/people.py
```diff
@@ -62,6 +62,27 @@ def add_missing_verification(cls, v):
...
+            raise ValueError("Display name can only contain alphabetic characters and spaces.")
+
+        if len(value) > cls.max_length:
+            raise ValueError(f"Display name must be at most {cls.max_length} characters long.")
+
```
ml-evs, line 82: I wonder if we should also canonicalize to "title case", e.g, first char of each name is capitalized. Awkward cases I can think of:

double-barelled names, Dutch style `van der Waals` or `t'Hooft`, American names with numbers at the end (John Smith III).

This could be done in the validator but not raise errors, just effects database storage

Probably no need to do that in this PR as I think it will be a bit annoying

webapp/src/components/EditAccountSettingsModal.vue
```diff
@@ -103,6 +111,26 @@ export default {
...
   },
   emits: ["update:modelValue"],
+  computed: {
+    displayNameValidationMessage() {
+      if (!this.user.display_name || !this.user.display_name.trim()) {
+        return "Name is required.";
+      } else if (this.user.display_name.length > 150) {
+        return "Name should be less than 150 characters.";
+      } else if (!/^[a-zA-Z\s]*$/.test(this.user.display_name)) {
+        return "Name should contain only alphabetic characters and spaces.";
```
ml-evs, line 121: Same comment as above that we should allow hyphens, apostrophes etc through

ml-evs (commented):
Hi @BenjaminCharmes, this is looking good, just a few minor comments (of things we can think about in future PRs) and one that I think should be addressed now.

pydatalab/pydatalab/models/people.py
ml-evs: I think I will retract all my comments about validation beyond checking if the name is empty, I fell down this trap: https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/

pydatalab/pydatalab/routes/v0_1/users.py
```diff
@@ -15,33 +17,47 @@ def save_user(user_id):
...
+    except ValueError as e:
+        return (jsonify({"status": "error", "message": str(e)}), 400)
 
-    if display_name:
-        update["display_name"] = display_name
+    try:
+        if contact_email or contact_email in (None, ""):
+            if contact_email == "":
+                contact_email = None
+            update["contact_email"] = EmailStr(contact_email)
```
ml-evs, line 40: `None` is not a valid `EmailStr` so we'll have to move that around a bit

ml-evs (commented):
See my comment here: https://github.com/the-grey-group/datalab/pull/646/files#r1539219700

I will remove some of the validation I asked for as we can never cover all cases, otherwise looks good!

pydatalab/pydatalab/models/people.py
ml-evs: See my comment above, I think this is a can of worms that we should try to close again...

ml-evs:
Okay, running into a few issues (that it looks like I ran into last time I did this since an issue I opened myself 2 years ago was the first result when I googled...)

Basically the custom types defined here either work in the pydantic model *OR* in the the type itself, but never in both. I've just updated them so that they work in both (by making them both constrained string types in pydantic and calling their validation methods on `__new__`).

ml-evs (dismissed):
> Okay, running into a few issues (that it looks like I ran into last time I did this since an issue I opened myself 2 years ago was the first result when I googled...)
>
> Basically the custom types defined here either work in the pydantic model OR in the the type itself, but never in both. I've just updated them so that they work in both (by making them both constrained string types in pydantic and calling their validation methods on __new__).

Have just tidied these up with my additional commits, as well as adding some integration tests to check that things behave. Happy to merge this now but will try to preserve the commit history so you can see what I changed @BenjaminCharmes!

Nice work and thanks again!

webapp/src/components/EditAccountSettingsModal.vue
```diff
@@ -103,6 +111,24 @@ export default {
     modelValue: Boolean,
   },
   emits: ["update:modelValue"],
+  computed: {
+    displayNameValidationMessage() {
+      if (!this.user.display_name || /^\s*$/.test(this.user.display_name)) {
+        return "Name is required.";
+      } else if (this.user.display_name.length > 150) {
+        return "Name should be less than 150 characters.";
```
jdbocarsly, line 119:
```suggestion
        return "Name should be no more than 150 characters.";
```

jdbocarsly (changes requested):
Looks good! Tiny comment on one of the error messages. 

Also, I'm noticing that when you are done with this modal and close it, the LoginDetails menu is still open. I think it would probably be more user friendly if it were closed, which you could do by changing line 28 in [LoginDetails.vue](https://github.com/BenjaminCharmes/datalab/blob/1bc621d38075a1da9d589a142d7bbfbb2e08060b/webapp/src/components/LoginDetails.vue#L28) to:

```
@click="editAccountSettingIsOpen = true; isUserDropdownVisible = false;"
```

Issue #653, opened 2024-03-20
Firefox e2e tests are persistently failing


ml-evs:
I think in my haste something has actually broken in the Firefox E2E testing, as its now failing for PRs that don't change much. I'll try to find time to investigate this soon.

ml-evs:
Looks like this is an upstream problem: https://github.com/cypress-io/cypress/issues/29172#issuecomment-2010756341

Can keep this open until we no longer see the failures

ml-evs:
This fix has been made upstream in cypress, we now just need to see how quickly it percolates down to the GitHub action we use. I'd prefer to wait rather than pinning specific browser versions in our CI.

ml-evs:
Ok it was nothing to do with waiting for the action to release; the action just uses whatever cypress version we have installed, so updating should fix it.

ml-evs:
Think we're now seeing some "real" firefox failures after merging #699...

ml-evs:
Reopened by #703

PR #655, opened 2024-03-21
User registration: default display name to GitHub username when no profile name is set

### LINKED ISSUE #652: Use GitHub username as datalab display name when no profile name is provided
Basically just a note to self, we should check the GitHub display name when someone registers, if it is empty, use their GitHub username as a display name.

ml-evs:
Closes #652.

Less important now that #627 is done but still an annoying bug!


ml-evs:
Only failures are related to #653, will merge

PR #656, opened 2024-03-21
Fix typo where CI tests were never run in chrome

ml-evs:
This is simply an oversight. If the tests pass I will merge.

Sadly I don't think the docker cache is being used during builds anymore, but I don't have the energy or wherewithal to go down that undocumented rabbit hole again.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -166,7 +166,7 @@ jobs:
           spec: |
             ./cypress/e2e/sampleTablePage.cy.js
             ./cypress/e2e/editPage.cy.js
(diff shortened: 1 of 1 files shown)


ml-evs:
> Sadly I don't think the docker cache is being used during builds anymore, but I don't have the energy or wherewithal to go down that undocumented rabbit hole again.

Ok now I'm just confused, this PR builds containers twice as fast as the previous one I merged, but still seems to not cache some stages. Either this is some lucky coincidence of hardware usage, or intermediate PRs are trashing the cache (by e.g., not including the most recent changes to builds).

PR #661, opened 2024-03-22
Convert Whinchat to use langchain 

vrajpatel9988:
Whinchat was originally built using an (outdated) openai API. Here, we change to using langchain to interface with openai.

### DIFF
--- a/pydatalab/Pipfile
+++ b/pydatalab/Pipfile
@@ -36,6 +36,7 @@ rosettasciio = "*"
 pyjwt = "*"
 invoke = "*"
 paramiko = "*"
+langchain = "*"
 
 [dev-packages]
 pytest = "*"
--- a/pydatalab/Pipfile
+++ b/pydatalab/Pipfile
@@ -25,7 +25,6 @@ pint = "*"
 periodictable = "*"
 flask-compress = "*"
 pillow = "*"
-openai = "==0.28"
 tiktoken = "*"
 galvani = "~= 0.4"
 navani = {git = "git+https://github.com/the-grey-group/navani.git@v0.1.5"}
--- a/pydatalab/Pipfile
+++ b/pydatalab/Pipfile
@@ -37,6 +37,7 @@ invoke = "*"
 paramiko = "*"
 langchain = "*"
 langchain-openai = "*"
+langchain-anthropic = "*"
 
 [dev-packages]
 pytest = "*"
--- a/pydatalab/Pipfile
+++ b/pydatalab/Pipfile
@@ -25,7 +25,7 @@ pint = "*"
 periodictable = "*"
 flask-compress = "*"
 pillow = "*"
-openai = "*"
+openai = "==0.28"
 tiktoken = "*"
 galvani = "~= 0.4"
 navani = {git = "git+https://github.com/the-grey-group/navani.git@v0.1.5"}
--- a/pydatalab/Pipfile
+++ b/pydatalab/Pipfile
@@ -25,7 +25,7 @@ pint = "*"
 periodictable = "*"
 flask-compress = "*"
 pillow = "*"
-openai = "~= 0.28"
+openai = "*"
 tiktoken = "*"
 galvani = "~= 0.4"
 navani = {git = "git+https://github.com/the-grey-group/navani.git@v0.1.5"}
--- a/pydatalab/Pipfile
+++ b/pydatalab/Pipfile
@@ -37,6 +37,7 @@ pyjwt = "*"
 invoke = "*"
 paramiko = "*"
 langchain = "*"
+langchain-openai = "*"
 
 [dev-packages]
 pytest = "*"
--- a/pydatalab/pydatalab/apps/chat/blocks.py
+++ b/pydatalab/pydatalab/apps/chat/blocks.py
@@ -2,7 +2,6 @@ import json
 import os
 from typing import Sequence
 
-import tiktoken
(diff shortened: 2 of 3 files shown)


pydatalab/pydatalab/apps/chat/blocks.py
```diff
@@ -14,23 +14,7 @@
...
-    num_tokens = 0
-    for message in messages:
-        num_tokens += tokens_per_message
-        for key, value in message.items():
-            num_tokens += len(encoding.encode(value))
-            if key == "name":
-                num_tokens += tokens_per_name
-    num_tokens += 3  # every reply is primed with <|start|>assistant<|message|>
-    return num_tokens
+openai_client = ChatOpenAI(api_key=os.environ.get("OPENAI_API_KEY"), model=MODEL)
```
ml-evs, line 17: The tests are currently failing on other blocks because this client fails to initialize without an API key. Probably the creation of this client should be pushed inside a block (and made separately for each new block), as e.g.,

```python
def __init__(self, ...):
    self.openai_client = ChatOpenAI(api_key=os.environ.get("OPENAI_API_KEY"), model=MODEL)

```

ml-evs (commented):
Looking good so far @vrajpatel9988! Just a minor comment on why the tests currently fail

pydatalab/pydatalab/apps/chat/blocks.py
vrajpatel9988: I have fixed this now.

pydatalab/pydatalab/apps/chat/blocks.py
```diff
@@ -2,9 +2,9 @@
 import os
 from typing import Sequence
 
-import openai
-import tiktoken
+from langchain_openai import ChatOpenAI
 
+# from langchain.chat_models import ChatOpenAI
```
ml-evs, line 7:
```suggestion
```

Presumably this can be deleted now

pydatalab/pydatalab/apps/chat/blocks.py
```diff
@@ -14,23 +14,7 @@
...
-    num_tokens = 0
-    for message in messages:
-        num_tokens += tokens_per_message
-        for key, value in message.items():
-            num_tokens += len(encoding.encode(value))
-            if key == "name":
-                num_tokens += tokens_per_name
-    num_tokens += 3  # every reply is primed with <|start|>assistant<|message|>
-    return num_tokens
+# openai_client = ChatOpenAI(api_key=os.environ.get("OPENAI_API_KEY"), model=MODEL)
```
ml-evs, line 17:
```suggestion
```

Likewise here

pydatalab/pydatalab/apps/chat/blocks.py
```diff
@@ -40,14 +24,17 @@ class ChatBlock(DataBlock):
...
-Answer questions in markdown. Specify the language for all markdown code blocks. You can make diagrams by writing a mermaid code block or an svg code block. When writing mermaid code, you must use quotations around each of the labels (e.g. A["label1"] --> B["label2"])
-Be as concise as possible. When saying your name, type a bird emoji right after whinchat 🐦.
+        You are embedded within the program datalab, where you have access to JSON describing an ‘item’, or a collection of items, with connections to other items. These items may include experimental samples, starting materials, and devices (e.g. battery cells made out of experimental samples and starting materials).
+        Answer questions in markdown. Specify the language for all markdown code blocks. You can make diagrams by writing a mermaid code block or an svg code block. When writing mermaid code, you must use quotations around each of the labels (e.g. A["label1"] --> B["label2"])
+        Be as concise as possible. When saying your name, type a bird emoji right after whinchat 🐦.
```
ml-evs, line 29: This whitespace actually breaks the multiline string -- it's a bit gross but multiline strings in Python should have no indent at all. We could use something like [`textwrap.dedent`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) but its probably better to leave this as it was:

```suggestion
You are embedded within the program datalab, where you have access to JSON describing an ‘item’, or a collection of items, with connections to other items. These items may include experimental samples, starting materials, and devices (e.g. battery cells made out of experimental samples and starting materials).
Answer questions in markdown. Specify the language for all markdown code blocks. You can make diagrams by writing a mermaid code block or an svg code block. When writing mermaid code, you must use quotations around each of the labels (e.g. A["label1"] --> B["label2"])
Be as concise as possible. When saying your name, type a bird emoji right after whinchat 🐦.
```

ml-evs (changes requested):
Looking good @vrajpatel9988! Just a couple of minor comments, in the meantime I'm going to try deploying the PR to check that everything works in the real system

pydatalab/Pipfile
```diff
@@ -36,6 +35,9 @@ rosettasciio = "*"
 pyjwt = "*"
 invoke = "*"
 paramiko = "*"
+langchain = "*"
+langchain-openai = "*"
+langchain-anthropic = "*"
```
ml-evs, line 40: This dep isn't necessary for this PR but given the follow-up in #680 there's no harm in adding it now.

jdbocarsly: Its shouldn't be necessary, but removing it caused lots of problems in pipenv for some reason...

ml-evs (approved):
LGTM, thanks @vrajpatel9988!

jdbocarsly (approved):
looks good!

PR #665, opened 2024-03-23
Allow echem block to plot non-cyclic data

### LINKED ISSUE #664: Unable to read some voltammetry files
I've noticed that some MPR files for voltammetry cannot be read by navani due to missing `state` column. We should just fill this in with nulls where not present.

ml-evs:
Currently we always try to compute cycle summaries on echem data. This fails when there is no `state` in the data, indicating when a cycle starts/ends. For e.g., linear/sweep voltammetry this data is not available, but we can still easily read the files and make useful plots. This PR simply ignores any errors from the navani summariser (for now) and lets the UI still plot whatever data is available.

Ideally we'd add some error checking in navani itself that handles this in a nicer way.

Closes #664


ml-evs:
Pretty trivial change so I'm going to force this one through so I can deploy it with some other PRs.

PR #666, opened 2024-03-24
Improve block warning/error reactivity in the UI

ml-evs:
This PR converts the cycle summary error from #665 into a warning, and makes sure that such warnings disappear when they are no longer valid in the UI.

This required a bit of manual UI hack, but probably fine for now.


jdbocarsly (approved):
Looks good (though we should figure out how to get current to show up in these plots as well)

PR #667, opened 2024-03-24
Serve data about available block types in API

ml-evs:
Supersedes #455 which had gone stale. Aiming to push this one just a little further to replicate our current behavior before adding e.g., block docs in the UI.

For now, this simply replicates the current behavior, but begins exposing some block info in the API directly. I also cleaned up some dead routes, test pages and did a fixup of some of the block docstrings.

This info can then be used by e.g., @BenjaminCharmes to begin populating block-level documentation in the app, in preparation for a future broader blocks rework in #616.


BenjaminCharmes (approved):
Great!

webapp/src/resources.js
```diff
@@ -1,5 +1,5 @@
 // Resources for the application
-import DataBlockBase from "@/components/datablocks/DataBlockBase";
+import GenericBlock from "@/components/datablocks/GenericBlock";
```
ml-evs, line 2:
```suggestion
import DataBlockBase from "@/components/datablocks/DataBlockBase";
```

webapp/src/resources.js
```diff
@@ -45,7 +45,7 @@ export const UPPY_MAX_NUMBER_OF_FILES =
 export const debounceTime = 250; // time after user stops typing before request is sent
 
 export const blockTypes = {
-  comment: { description: "Comment", component: DataBlockBase },
+  comment: { description: "Comment", component: GenericBlock },
```
ml-evs, line 48:
```suggestion
  comment: { description: "Comment", component: DataBlockBase },
```

BenjaminCharmes (approved):
Looks great !

PR #680, opened 2024-03-28
Add ability to select between different chat models in the whinchat

vrajpatel9988:
This PR adds the ability to switch between different LLM chat models (currently Anthropic and OpenAI models), and also adds some functionality to estimate price per message for each model.


pydatalab/pydatalab/apps/chat/blocks.py
```diff
@@ -10,14 +11,14 @@
...
 __all__ = "ChatBlock"
-MODEL = "gpt-3.5-turbo-0613"
-MAX_CONTEXT_SIZE = 4097
 
 
 class ChatBlock(DataBlock):
     blocktype = "chat"
     description = "Virtual assistant"
     accepted_file_extensions: Sequence[str] = []
+    ChatClient = None
```
ml-evs, line 20: Should import `langchain.chat_models.BaseChatModel` above somewhere then use this for the type. Also rename the attribute to `chat_client`.

```suggestion
    chat_client: BaseChatModel | None = None
```

pydatalab/pydatalab/apps/chat/blocks.py
```diff
@@ -60,7 +99,7 @@ def render(self):
                 {
                     "role": "user",
                     "content": f"""Here is the JSON data for the current item(s): {info_json}.
-Start with a friendly introduction and give me a one sentence summary of what this is (not detailed, no information about specific masses). """,
+                    Start with a friendly introduction and give me a one sentence summary of what this is (not detailed, no information about specific masses). """,
```
ml-evs, line 102:
```suggestion
Start with a friendly introduction and give me a one sentence summary of what this is (not detailed, no information about specific masses). """,
```

Again whitespace inside multiline strings starts at the edge of the file (col:0) (a bit of a Python oddity) -- it will count as several tokens(?) so we should avoid it.

pydatalab/pydatalab/apps/chat/blocks.py
```diff
@@ -81,8 +120,23 @@ def render(self):
                 return
 
         try:
+            model_name = self.data["model"]
+            model_dict = self.data["available_models"][model_name]
+            LOGGER.warning(f"Initializing chatblock with model: {model_name}")
+
+            if "claude" in model_name:
```
ml-evs, line 127:
```suggestion
            if model_name.startswith("claude"):
```

pydatalab/pydatalab/apps/chat/blocks.py
```diff
@@ -81,8 +120,23 @@ def render(self):
...
+                self.ChatClient = ChatAnthropic(
+                    anthropic_api_key=os.environ["ANTHROPIC_API_KEY"],
+                    model=model_name,
+                )
+            elif "gpt" in model_name:
```
ml-evs, line 132:
```suggestion
            elif model_name.startswith("gpt"):
```

probably slightly more robust; some OSS models use 'GPT' in the name too

pydatalab/pydatalab/apps/chat/blocks.py
```diff
@@ -96,31 +150,31 @@ def render(self):
                 else:
                     langchain_messages.append(AIMessage(content=message["content"]))
 
-            token_count = self.openai_client.get_num_tokens_from_messages(langchain_messages)
+            token_count = self.ChatClient.get_num_tokens_from_messages(langchain_messages)
```
ml-evs, line 153:
```suggestion
            token_count = self.chat_client.get_num_tokens_from_messages(langchain_messages)
```

pydatalab/pydatalab/apps/chat/blocks.py
```diff
@@ -96,31 +150,31 @@ def render(self):
...
                 self.data[
                     "error_message"
-                ] = f"""This conversation has reached its maximum context size and the chatbot won't be able to respond further ({token_count} tokens, max: {MAX_CONTEXT_SIZE}). Please make a new chat block to start fresh."""
+                ] = f"""This conversation has reached its maximum context size and the chatbot won't be able to respond further ({token_count} tokens, max: {model_dict['context_window']}). Please make a new chat block to start fresh, or use a model with a larger context window"""
                 return
 
-            # Call the OpenAI client with the invoke method
-            response = self.openai_client.invoke(langchain_messages)
+            # Call the ChatClient client with the invoke method
+            response = self.ChatClient.invoke(langchain_messages)
```
ml-evs, line 164:
```suggestion
            # Call the chat client with the invoke method
            response = self.chat_client.invoke(langchain_messages)
```

pydatalab/pydatalab/apps/chat/blocks.py
```diff
@@ -96,31 +150,31 @@ def render(self):
...
 
             langchain_messages.append(response)
 
-            token_count = self.openai_client.get_num_tokens_from_messages(langchain_messages)
+            token_count = self.ChatClient.get_num_tokens_from_messages(langchain_messages)
```
ml-evs, line 168:
```suggestion
            token_count = self.chat_client.get_num_tokens_from_messages(langchain_messages)
```

ml-evs (commented):
Looking good! Some minor style comments only

pydatalab/pydatalab/apps/chat/blocks.py
```diff
@@ -81,8 +125,23 @@ def render(self):
...
+                self.chat_client = ChatAnthropic(
+                    anthropic_api_key=os.environ["ANTHROPIC_API_KEY"],
+                    model=model_name,
+                )
+            elif model_name.startswith("gpt"):
+                self.chat_client = ChatOpenAI(
+                    api_key=os.environ.get("OPENAI_API_KEY"),
+                    model=model_name,
+                )
+
```
ml-evs, line 142: probably needs 


```suggestion
                )
            else: 
                raise RuntimeError(f"Unknown model {model_name}")
```

or something similar

ml-evs (commented):
Just one more minor comment below.

I think I still want to deploy this and play around with it before merging, so might take a day or two. Does langchain have any kind of "dummy" client that returns placeholder text that we could use for testing?

pydatalab/pydatalab/apps/chat/blocks.py
jdbocarsly: I'll put in an earlier check that the provided model is in the list of available models

jdbocarsly:
> I think I still want to deploy this and play around with it before merging, so might take a day or two. Does langchain have any kind of "dummy" client that returns placeholder text that we could use for testing?

It would appear there is: https://langchain-contrib.readthedocs.io/en/latest/llms/fake.html

pydatalab/pydatalab/apps/chat/blocks.py
```diff
@@ -10,28 +12,66 @@
...
     defaults = {
         "system_prompt": """You are whinchat (lowercase w), a virtual data managment assistant that helps materials chemists manage their experimental data and plan experiments. You are deployed in the group of Professor Clare Grey in the Department of Chemistry at the University of Cambridge.
 You are embedded within the program datalab, where you have access to JSON describing an ‘item’, or a collection of items, with connections to other items. These items may include experimental samples, starting materials, and devices (e.g. battery cells made out of experimental samples and starting materials).
-Answer questions in markdown. Specify the language for all markdown code blocks. You can make diagrams by writing a mermaid code block or an svg code block. When writing mermaid code, you must use quotations around each of the labels (e.g. A["label1"] --> B["label2"])
+Answer questions in markdown. Specify the language for all markdown code blocks. You can make diagrams by writing a mermaid code block or an svg code block. When jfjfriting mermaid code, you must use quotations around each of the labels (e.g. A["label1"] --> B["label2"])
```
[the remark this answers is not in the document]
ml-evs: ... :sweat_smile:

pydatalab/pydatalab/apps/chat/blocks.py
```diff
@@ -10,28 +12,66 @@
 from pydatalab.utils import CustomJSONEncoder
 
 __all__ = "ChatBlock"
-MODEL = "gpt-3.5-turbo-0613"
-MAX_CONTEXT_SIZE = 4097
 
 
 class ChatBlock(DataBlock):
```
ml-evs, line 17:
```suggestion
class ChatBlock(DataBlock):
   """This block uses API calls to external LLMs via Langchain to provide a conversational
   interface to a user's data. 
   
   Implemented models include:
   - the GPT series of models from OpenAI
   - Claude from Anthropic
   
   Each needs the server to be configured with the corresponding API keys:
   - `OPENAI_API_KEY`,
   - `ANTHROPIC_API_KEY`.
   
   A discussion of this block can be found in:
   
       Jablonka *et al*, Digital Discovery, 2023,2, 1233-1250, DOI: 10.1039/d3dd00113j
       
   """
```

something like this needs to be added. Ideally with even more description.

ml-evs (changes requested):
Also see my other remaining comment that fixes a typo.

Please can you add documentation regarding which API keys/environment variables need to be set on the server for this to work? For example, the server will check for the existence of `OPENAI_API_KEY` on startup, but not Anthropic, so this should be added at: https://github.com/the-grey-group/datalab/blob/fe7a9dce3c8baaf9ea6a9b7a247c2241aed39d76/pydatalab/pydatalab/main.py#L57-L72

I think it would also be good to write a docstring for the block itself (will add this as a comment below) that adds some more description, so at least this is searchable from the docs page.

ml-evs (commented):
Hi @vrajpatel9988, this seems to work well with GPT-based models but for Claude-based models I get the error:

```
Received an error from the API: Could not import transformers python package. This is needed in order to calculate get_token_ids. Please install it with `pip install transformers`..

 Consider choosing a different model and reloading the block.
```

ml-evs:
> Hi @vrajpatel9988, this seems to work well with GPT-based models but for Claude-based models I get the error:
> 
> ```
> Received an error from the API: Could not import transformers python package. This is needed in order to calculate get_token_ids. Please install it with `pip install transformers`..
> 
>  Consider choosing a different model and reloading the block.
> ```

I know this package was initially added as a dependency, but as these ML API packages move so fast (looks like there have been around 15 releases to each of the various langchain/openai/anthropic packages since this PR was opened), we should probably do a better job of pinning them.

ml-evs (dismissed):
I've added the transformers dep, in combination with some other merged PRs this now works fine for me locally, thanks @vrajpatel9988 !

jdbocarsly (approved):
👍

PR #687, opened 2024-04-03
Add `account_status` field (active, unverified or deactivated) to People model

BenjaminCharmes:
- [x] Add account_status to the People model
- [x] Modification of the saveUser to update account_status
- [x] Add a way to edit account status in the admin dashboard
- [ ] Add test ?


ml-evs (commented):
Thanks @BenjaminCharmes! All that remains is adding confirmation dialogue for the account status, as you did for roles.

Future notes:

- [ ] Follow up PR needed to decide what a deactivated account is allowed to do
- [ ] Maybe streamline the UI at some point with buttons for activate/deactivate that appear on context

ml-evs:
pre-commit.ci fix

pydatalab/pydatalab/models/people.py
```diff
@@ -90,6 +90,14 @@ def __new__(cls, value):
         return cls.validate(value)
 
 
+class AccountStatus(str, Enum):
+    """A string enum representing the account status."""
+
+    ACTIVE = "active"
+    UNVERIFIED = "unverified"
+    DISABLED = "disabled"
```
ml-evs, line 98: Just thinking about this again, I think maybe "disabled" was also a bad term, could we replace it with "deactivated" here and in the UI please?

ml-evs (commented):
Any ideas why the tests are failing? I can rebase this PR and have a look too, otherwise this looks nearly ready to go.

We will need to do follow-up work immediately to set the status of user accounts to "unverified" when they first sign-in, and add all the various permissions checks (and perhaps UI) to do with that.

BenjaminCharmes:
pre-commit.ci autofix

BenjaminCharmes:
pre-commit.ci autofix

PR #674, opened 2024-03-28
Add admin dashboard with user management controls

BenjaminCharmes:
Related to #576 

This PR adds UI and API routes to allow admins to manage user details and roles.

- [x]  Display list of users
- [x]  Admin can update user's role
- [x] Add test front and/or back ?

Issue #645, opened 2024-03-13
Connecting external accounts whilst logged in


ml-evs:
As discussed (cc @BenjaminCharmes), the next step of the account details page should let you connect e.g., an ORCID account when you are already logged in.

ml-evs:
Cloesd by #693.

PR #693, opened 2024-04-10
Enable ORCID connection by default

ml-evs:
Begins to address #645.

ORCID should now be enabled by default as a login source, when configured via the appropriate environment variables.

Eventually the UI should adapt to not show it as an option when not configured on the server.

For now, the user must still register an account via GitHub, but they can then connect their ORCID and use it for future logins.

Once #674 and #687 are merged, admins will be able to allow users to request an account via ORCID alone.

### DIFF
--- a/pydatalab/pydatalab/login.py
+++ b/pydatalab/pydatalab/login.py
@@ -54,6 +54,11 @@ class LoginUser(UserMixin):
         return self.person.display_name
 
     @property
+    def contact_email(self) -> Optional[str]:
+        """Returns the top-level display name for the user, if set."""
+        return self.person.contact_email
+
+    @property
     def identities(self) -> List[Identity]:
         """Returns the list of identities of the user."""
         return self.person.identities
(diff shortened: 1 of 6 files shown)


pydatalab/pydatalab/login.py
```diff
@@ -53,6 +53,11 @@ def display_name(self) -> Optional[str]:
         """Returns the top-level display name for the user, if set."""
         return self.person.display_name
 
+    @property
+    def contact_email(self) -> Optional[str]:
+        """Returns the top-level display name for the user, if set."""
```
BenjaminCharmes, line 58:
```suggestion
        """Returns the top-level display name for the user, if set."""
```

If I'm right ?

BenjaminCharmes (commented):
Looks good ! 
Just a small copy/paste error on a comment (I suppose?), not very important.

pydatalab/pydatalab/login.py
ml-evs: Well spotted, thanks!

PR #700, opened 2024-04-12
Added a way to refresh API Key from Account Settings UI

### LINKED ISSUE #695: Add API key refresh to account settings UI
Currently you get and refresh your API key by hitting an API route directly in the browser (`/generate-api-key`). We should add something to the account settings modal that lets people request a new API key and is easy to copy etc, with a confirmation box that it will remove their old API key.

The important thing to note is that we do not store the 'real' API key for a user, we store a hash, so once generated we cannot show them the same key again (but we can point out if it has expired).

This is probably one for you when you get a chance @BenjaminCharmes (so I will assign you)!

BenjaminCharmes:
Closes #695 

I've added a button in "Account Settings" to refresh the user's API key.
There's a warning message indicating that the old key will be deleted.
The key is displayed in the "Account Settings" Modal and disappears once the user closes the Modal.

### DIFF
--- a/webapp/src/components/EditAccountSettingsModal.vue
+++ b/webapp/src/components/EditAccountSettingsModal.vue
@@ -90,7 +90,13 @@
           <div class="form-group col-md-12">
             <label for="api-key" class="col-form-label">API Key:</label>
             <div v-if="apiKeyDisplayed" class="input-group">
-              <input type="text" class="form-control" style='text-overflow: ellipsis' :value="apiKey" readonly />
+              <input
+                type="text"
+                class="form-control"
+                style="text-overflow: ellipsis"
+                :value="apiKey"
+                readonly
+              />
               <div class="input-group-append">
                 <button class="btn btn-outline-secondary" type="button" @click="copyToClipboard">
                   <font-awesome-icon icon="copy" />
@@ -98,7 +104,9 @@
               </div>
             </div>
             <div>
-                <button class="btn btn-default mt-2" @click="requestAPIKey">Regenerate API Key</button>
+              <button class="btn btn-default mt-2" @click="requestAPIKey">
+                Regenerate API Key
+              </button>
             </div>
           </div>
         </div>
--- a/webapp/src/components/EditAccountSettingsModal.vue
+++ b/webapp/src/components/EditAccountSettingsModal.vue
@@ -2,7 +2,7 @@
   <form @submit.prevent="submitForm" class="modal-enclosure">
     <Modal
       :modelValue="modelValue"
-      @update:modelValue="$emit('update:modelValue', $event)"
+      @update:modelValue="resetForm"
       :disableSubmit="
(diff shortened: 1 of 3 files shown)


ml-evs (approved):
This is super useful, thanks @BenjaminCharmes!

I'm happy to merge as-is (and will do after I tidy up some other PRs), but just a couple of notes for things we might want to add in the future.

- [ ] Some documentation for what an API key is for, how and where to use it. Perhaps this could be a question mark next to the generate API key button that opens a modal or a popover with some examples? This kind of documentation could be put in many places around the app

webapp/src/components/EditAccountSettingsModal.vue
```diff
@@ -86,6 +86,20 @@
             >
           </div>
         </div>
+        <div class="form-row">
+          <div class="form-group col-md-6">
```
ml-evs, line 90:
```suggestion
          <div class="form-group col-md-12">
```

webapp/src/components/EditAccountSettingsModal.vue
```diff
@@ -86,6 +86,20 @@
...
+            <label for="account-name" class="col-form-label">API Key:</label>
+            <div v-if="apiKeyDisplayed" class="input-group">
+              <input type="text" class="form-control" :value="apiKey" readonly />
```
ml-evs, line 93:
```suggestion
              <input type="text" class="form-control" style='text-overflow: ellipsis' :value="apiKey" readonly />
```

ml-evs (changes requested):

webapp/src/components/EditAccountSettingsModal.vue
```diff
@@ -86,6 +86,20 @@
             >
           </div>
         </div>
+        <div class="form-row">
+          <div class="form-group col-md-6">
+            <label for="account-name" class="col-form-label">API Key:</label>
```
ml-evs, line 91:
```suggestion
            <label for="api-key" class="col-form-label">API Key:</label>
```

webapp/src/components/EditAccountSettingsModal.vue
```diff
@@ -86,6 +86,20 @@
...
+            <div v-if="apiKeyDisplayed" class="input-group">
+              <input type="text" class="form-control" :value="apiKey" readonly />
+              <div class="input-group-append">
+                <button class="btn btn-outline-secondary" type="button" @click="copyToClipboard">
+                  <font-awesome-icon icon="copy" />
+                </button>
+              </div>
+            </div>
+            <button class="btn btn-default mt-2" @click="requestAPIKey">Request New API Key</button>
```
ml-evs, line 100:
```suggestion
            <div>
                <button class="btn btn-default mt-2" @click="requestAPIKey">Regenerate API Key</button>
            </div>
```

ml-evs:
pre-commit.ci autofix

ml-evs (approved):

PR #706, opened 2024-04-17
Added `HelpBubble` component to display inline documentation in the UI

### LINKED ISSUE #705: Inline documentation in the UI
We would like to add some help bubbles at various places in the UI. These should often use values provided directly by the API (see e.g., #667).

The first one to add could be a description of what the API keys do next to the "Regenerate API key" button.

We can use some kind of popover library for this, e.g., [popper.js](https://popper.js.org/docs/v2/).

BenjaminCharmes:
Closes #705 

Adding a HelpBubble.vue component using popper.js

It's currently only displayed next to the "Regenerate API Key" button in Account Settings to show what it might look like.

CSS can be modified if needed and message content to be defined.


jdbocarsly:
This looks pretty good- a few thoughts:

1. The popper component shows up as a div inside the modal, which does some strange things (like add a scroll bar) on my screen. It would be nice if it could be show up in front of the modal, and not affect spacing, etc. in the modal. I think this would be just a CSS change.

3. The little "i" does look good in this instance, but I'm concerned what will happen if this pattern is used across all fields in the UI, especially on the EditPage. Maybe we can play around with having the popper show up when hovering over the label, but only after a ~1 second delay?

webapp/src/components/EditAccountSettingsModal.vue
```diff
@@ -131,6 +127,8 @@ export default {
       apiUrl: API_URL,
       apiKeyDisplayed: false,
       apiKey: null,
+      apiKeyHelpMessage:
+        "Lorem, ipsum dolor sit amet consectetur adipisicing elit. Minus adipisci quod earum maxime vero architecto molestiae ipsum est debitis saepe blanditiis repellat, libero numquam. Autem, qui? Libero tenetur deleniti numquam!Lorem, ipsum dolor sit amet consectetur adipisicing elit. Minus adipisci quod earum maxime vero architecto molestiae ipsum est debitis saepe blanditiis repellat, libero numquam. Autem, qui? Libero tenetur deleniti numquam!",
```
ml-evs, line 131:
```suggestion
        "You can use your API key via the datalab-api Python package, or pass it as an HTTP header "DATALAB-API-KEY" with the tool of your choice (e.g., curl).",
```

ml-evs (approved):
👍

PR #707, opened 2024-04-18
Recreate user index with new settings if already existing

ml-evs:
#693 updated the users index to fix an issue with uniqueness constraints on contact emails/identities.

This required some modification to the database index for users, which now has slightly different settings. Currently when an existing instance is upgraded, this index creation fails due to a conflicting name. This PR simply drops any existing collection if it is found, then recreates it (as is done for the FTS). We could roll this out more generally to other indices as and when we break them.


ml-evs:
pre-commit.ci fix

ml-evs:
pre-commit.ci autofix

PR #716, opened 2024-04-25
[WIP] Handle deployment settings in a new collection in the database

### LINKED ISSUE #704: Handle deployment settings in a new collection in the database
We have a number of settings on the frontend and backend that are currently set using environement variables. It would be nice for admins to be able to set many of these settings from the app itself. 

To do this, we will need a new database collection called "settings" (or similar), that holds a document with settings for the overall deployment, as well as documents for individual block types. Many of the settings in `resources.js` could be stored here, as well as things like API keys for llm models, configuration for login types, etc. These settings could be changeable by admins from the new admin panel that is in development.

BenjaminCharmes:
Closes #704

- Added a model setting with name, value and access_level (public or private)
- Added a setting table to the admin dashboard that display every setting set in the db
- Added a way to update these settings

Should we have a default value for certain parameters defined in the code? or
Have an empty field in the admin dashboard for parameters not yet defined? or
Just enter them manually in the database the first time?

### DIFF
--- a/pydatalab/pydatalab/models/utils.py
+++ b/pydatalab/pydatalab/models/utils.py
@@ -106,7 +106,8 @@ class PintType(str):
     def validate(self, v):
         q = self.Q(v)
         if not q.check(self._dimensions):
-            raise ValueError("Value {v} must have dimensions of mass, not {v.dimensions}")
+            raise ValueError(
+                "Value {v} must have dimensions of mass, not {v.dimensions}")
         return q
 
     @classmethod
@@ -280,3 +281,17 @@ class Constituent(BaseModel):
                     raise ValueError("Inline substance must have a name!")
                 return InlineSubstance(name=name, chemform=chemform)
         return v
+
+
+class AccessLevel(str, Enum):
+    PUBLIC = "public"
+    PRIVATE = "private"
+
+
+class Setting(BaseModel):
+
+    name: str
+
+    value: str
+
+    access_level: AccessLevel
--- a/pydatalab/pydatalab/routes/v0_1/admin.py
+++ b/pydatalab/pydatalab/routes/v0_1/admin.py
@@ -15,8 +15,8 @@ def check_authentication():
     if request.method == "OPTIONS":
         return
 
-    if not current_user.is_authenticated:
-        return jsonify({"error": "Unauthorized"}), 401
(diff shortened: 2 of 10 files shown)


BenjaminCharmes:
pre-commit.ci autofix

ml-evs:
pre-commit.ci autofix

pydatalab/pydatalab/models/utils.py
```diff
@@ -280,3 +280,16 @@ def coerce_reference(cls, v):
                     raise ValueError("Inline substance must have a name!")
                 return InlineSubstance(name=name, chemform=chemform)
         return v
+
+
+class AccessLevel(str, Enum):
```
ml-evs, line 285:
```suggestion
class AccessLevel(str, Enum):
   """Whether a setting is reported in the public API, e.g., settings that are 
   required by the UI, or whether it is only reported to admins, e.g., deployment
   level settings.
   
   
   """
```

ml-evs: Actually maybe these models are not necessary right now

BenjaminCharmes:
pre-commit.ci autofix

pydatalab/pydatalab/routes/v0_1/admin.py
```diff
@@ -101,3 +92,42 @@ def save_role(user_id):
...
+            jsonify(
+                {
+                    "status": "success",
+                    "message": "No update was performed",
+                }
+            ),
+            200,
+        )
+
+    return (jsonify({"status": "success"}), 200)
```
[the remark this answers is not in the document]
BenjaminCharmes: Done. For the moment, only the "LOGO_URL" and "HOMEPAGE_URL" settings work. So I also put a dropdown for "name" with only these two entries.

ml-evs:
I think much like #604, there's enough to tidy up here (and stuff that could subtly break deployments) that we should hold off until after the 0.4 release to merge this.

BenjaminCharmes:
As discussed during the datalab meeting and for future work:

- [] Should we remove the "Add a setting" button?
- [] Should the settings each have an entry in db or a collection by "acces_level" that includes all relative settings?
- [] Do we need to be able to delete a setting or just modify it?
- [] What settings should we put in the admin dashboard for the future? (only LOGO_URL and HOMEPAGE_URL work for now)

PR #718, opened 2024-04-25
[WIP] Upgrade whinchat into an agent

jdbocarsly:
This PR will add tool-calling functionality to whinchat, so that it can handle more complex tasks.


ml-evs:
I guess this is now stale? Please re-open if not

jdbocarsly:
No, this isn't stale - we should discuss how we want to proceed with the agent stuff, but will open for now

PR #719, opened 2024-04-27
Populating dynamic block-level documentation as UI tooltips

BenjaminCharmes:
Related to #705 & #667 

Add information for each block with popper.js on hover in the “Add a block” dropdown inside the EditPage.
The names in the dropdown are slightly different, as we now use those fetched directly from the models and not those in @resource.js (can be change if needed).

The design of the popper.js window is customizable, so don't hesitate to give me feedback.

Screenshot 2024-04-27 at 13 13 42

### DIFF
--- a/pydatalab/pydatalab/apps/nmr/blocks.py
+++ b/pydatalab/pydatalab/apps/nmr/blocks.py
@@ -17,7 +17,7 @@ class NMRBlock(DataBlock):
     name = "NMR"
     description = "A simple NMR block for visualizing 1D NMR data from Bruker projects."
 
-    accepted_file_extensions = ".zip"
+    accepted_file_extensions = (".zip",)
     defaults = {"process number": 1}
     _supports_collections = False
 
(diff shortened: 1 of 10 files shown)


BenjaminCharmes:
pre-commit.ci autofix

pydatalab/pydatalab/apps/nmr/blocks.py
```diff
@@ -17,7 +17,7 @@ class NMRBlock(DataBlock):
     name = "NMR"
     description = "A simple NMR block for visualizing 1D NMR data from Bruker projects."
 
-    accepted_file_extensions = ".zip"
+    accepted_file_extensions = (".zip",)
```
ml-evs, line 20: Nice catch, this should have been flagged by mypy but we're using a weird type hint on the base class (`Sequence[str]`) -- I'll push a commit that fixes it to (`tuple[str, ...]`) this PR and see if anything else comes out

ml-evs (commented):
Looking good, I'm going to test this out locally and push the tweaks mentioned and if all works, will go ahead and merge.

Thanks @BenjaminCharmes

webapp/src/views/EditPage.vue
```diff
@@ -200,6 +205,12 @@ export default {
         this.lastModified = formatDistanceToNow(save_date, { addSuffix: true });
       }
     },
+    async getBlocksInfo() {
+      if (!this.blocksInfos) {
+        let data = await getBlocksInfos();
+        this.$store.commit("setBlocksInfos", data);
```
ml-evs, line 211: I think we should do the store commit inside `server_fetch_utils` so that it can be reused elsewhere without needing the edit page loaded directly

webapp/src/store/index.js
```diff
@@ -263,6 +264,9 @@ export default createStore({
     setItemGraph(state, payload) {
       state.itemGraphData = payload;
     },
+    setBlocksInfos(state, blocksInfos) {
+      state.blocksInfos = blocksInfos;
```
ml-evs, line 268: I think it would also be good if we can store this as a dict of block type -> block info, so that each block component can easily make use of its own info

ml-evs (approved):
LGTM now @BenjaminCharmes! I'll have a follow up PR that uses some of this stuff to make at tooltip inside the block title too

PR #722, opened 2024-05-01
Bump Flask and Werkzeug to the latest releases on the v3 series

ml-evs:
Long overdue, but we it seems we can now ~painlessly~ update these packages.


jdbocarsly (approved):
Seems to work well for me locally! 
<sub><sup>now do bokeh</sup></sub>

Issue #710, opened 2024-04-19
Update API permissions based on user's `account_status`


ml-evs:
Now that #687 is merged, we need to do the following additional work:

- [x] Set a new user's account status to unverified in the case of ORCID & email registration, but allow GitHub through when an org allow list is provided
- [x] Prevent unverified users from accessing any non-public data, or uploading any data themselves, or generating API keys
- [x] Add tests for unverified users, and that permissions are updated when they are "verified"
- [x] Add more UI hints for unverified users (red dot next to their name in account settings), and the same hint for admins that a user has asked to be verified (this could also trigger an email send at some point). (PR #724)

PR #724, opened 2024-05-03
[WIP] Add notification-dot for user with unverified account_status

BenjaminCharmes:
Related to #710 

Add more UI hints for unverified users (red dot next to their name in account settings), and the same hint for admins that a user has asked to be verified (this could also trigger an email send at some point).


webapp/src/components/UserBubble.vue
```diff
@@ -1,21 +1,33 @@
...
-      md5(this.creator.contact_email || this.creator.display_name) +
-      '?d=' +
-      this.gravatar_style
-    "
-    class="avatar"
-    :width="size"
-    :height="size"
-    :title="this.creator.display_name"
-  />
+  <div class="bubble-wrapper">
```
ml-evs, line 2: I think we should make a separate component for the top bar "ThisUserBubble" vs the one used in the creators lists, that way we can also have custom components for admin vs normal user

ml-evs (commented):
Looking good so far, one minor adjustment

BenjaminCharmes:
pre-commit.ci autofix

ml-evs (approved):
Great work @BenjaminCharmes, thanks!

PR #727, opened 2024-05-04
Consider account status during auth and refactor API around blueprints

### LINKED ISSUE #710: Update API permissions based on user's `account_status`

ml-evs:
Closes #710 by refactoring all routes around blueprints and introducing some new decorators for permissions based on  `account_status`.

### DIFF
--- a/pydatalab/pydatalab/login.py
+++ b/pydatalab/pydatalab/login.py
@@ -10,7 +10,7 @@ from bson import ObjectId
 from flask_login import LoginManager, UserMixin
 
 from pydatalab.models import Person
-from pydatalab.models.people import Identity, IdentityType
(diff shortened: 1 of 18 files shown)


BenjaminCharmes (approved):

jdbocarsly (approved):
looks great!

PR #733, opened 2024-05-15
Make account status on registration be configurable by provider

ml-evs:
This PR updates our account registration workflow so that we can have the following behavior for our supported OAuth providers:

- If a GitHub allowed org is configured, nothing has changed, and a user can register and have their account activated automatically
- If `ORCID_AUTO_ACTIVATE_ACCOUNTS` is enabled, when someone registers via ORCID, their account is automatically activated. Otherwise, an admin must approve it.
- If `EMAIL_DOMAIN_ALLOW_LIST` is configured explicitly to `None`, then any address can automatically register an active account. Otherwise, if configured to `[]`, then no-one can register, and if configured to a set of domains, registration will require admin approval.

The main change here is that ORCID auth will now register an account in the default implementation, rather than relying on patches to specific deployments.

### DIFF
--- a/pydatalab/tests/server/test_auth.py
+++ b/pydatalab/tests/server/test_auth.py
@@ -7,8 +7,8 @@ def test_allow_emails():
     assert _check_email_domain("test@example.org", ["example.org", "example2.org"])
     assert _check_email_domain("test@subdomain.example.org", ["example.org", "example2.org"])
     assert _check_email_domain("test@subdomain.example.org", ["subdomain.example.org"])
-    assert _check_email_domain("test@example2.org", [])
-    assert not _check_email_domain("test@example2.org", None)
+    assert not _check_email_domain("test@example2.org", [])
+    assert _check_email_domain("test@example2.org", None)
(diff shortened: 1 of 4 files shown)


pydatalab/tests/server/test_auth.py
```diff
@@ -7,8 +7,8 @@ def test_allow_emails():
     assert _check_email_domain("test@example.org", ["example.org", "example2.org"])
     assert _check_email_domain("test@subdomain.example.org", ["example.org", "example2.org"])
     assert _check_email_domain("test@subdomain.example.org", ["subdomain.example.org"])
-    assert _check_email_domain("test@example2.org", [])
-    assert not _check_email_domain("test@example2.org", None)
+    assert not _check_email_domain("test@example2.org", [])
+    assert _check_email_domain("test@example2.org", None)
```
ml-evs, line 11: I would consider this a bug previously; this only affected the public deployment

BenjaminCharmes (commented):
Looks great !

jdbocarsly (approved):
👍

PR #735, opened 2024-05-15
Fix styling of StyledInput 

jdbocarsly:
Recent changes to StyledInput in #706 have broken the display on the StartingMaterialInformation page by removing the bootstrap form-control class. This PR reverts the changes.


jdbocarsly:
@BenjaminCharmes can you take a look at this, and make sure the inputs that you were using with this work with these changes? 

The issue is that the line:
```js
   :class="{ 'form-control': !readonly, 'form-control': readonly }"
```
tries to set the class to .form-control twice: once if `readonly` is true and once if `readonly` is not true. I  have reverted back to the correct behavior, but its possible it could break something you were doing.

jdbocarsly:
Closing this in favor of the simplified version of this at #751

Issue #732, opened 2024-05-14
Add gpt4o as a model option in whinchat


ml-evs:
Should be straightforward

ml-evs:
And now gemini I guess...

PR #736, opened 2024-05-15
Add gpt4o and other updated models

jdbocarsly:
addresses #732 by adding gpt-4o as an option in whinchat and setting it as the default. Also adds claude-3.5-sonnet and updates the other models.


jdbocarsly:
Screenshot 2024-05-15 at 11 19 29 AM

pydatalab/Pipfile.lock
```diff
@@ -108,11 +108,11 @@
         },
         "anthropic": {
             "hashes": [
-                "sha256:419a276eb20cfb7ddaac03c7e28e4e12df3ace71bcf33071a68c9a03c0dfcbdd",
```
ml-evs, line 111: What did you actually need to update for this PR to work? A minimum version should be added in the `Pipfile` too in that case (though I am also hoping #604 will sort this out sooner rather than later)

PR #737, opened 2024-05-16
Add block help to block title

ml-evs:
This PR extends #719 by adding an info hover to the block title itself, and uses the block info data to populate the `accepted_file_extensions` dynamically.

You can see the tweaks needed on each block for this, which emphasizes the need for a refactor like in #616, which makes a less generic base block that can do e.g., file drop downs.


ml-evs:
pre-commit autofix

ml-evs:
I will claim that this was reviewed in the meeting on Thursday, so will merge

Issue #634, opened 2024-03-08
Fix Chemical formula formatting of formulas such as Na2/3CoO2 and Ni(NO3)2.6H2O


jdbocarsly:

MJCliffe:
This might be the same issue, but if you put in the chemical formula section (NH4)2(SO4) it renders as NH$_4$SO$_4$ - this is pretty important for molecular stuff. Might be nice to have it also expand it to the empirical formula & molecular mass?

jdbocarsly:
Yes, this is part of the same issue. I've been meaning to fix this for a while, so will go on and try to come up with something that works nicely across a wider range of cases. For the mean time, the formula is actually just stored as a string in the database, so even though the display is incorrect, your data should be intact (you can see the actual data stored by clicking in the field in the UI)

ml-evs:
I think we should also have a fail-safe, if the entire formula cannot be captured by whatever regexp we conjure up for the formatting, it should just show the formula unformatted.

jdbocarsly:
> I think we should also have a fail-safe, if the entire formula cannot be captured by whatever regexp we conjure up for the formatting, it should just show the formula unformatted.
good idea

I also am thinking of all the ways we often express formulas, like Li1-xCoO2 if we aren't sure about x. It probably will be hard to properly handle everything people want to do, but we can certainly do a bit better.

MJCliffe:
This package might help: https://pypi.org/project/chemformula/#:~:text=Description-,ChemFormula%20is%20a%20Python%20class%20for%20working%20with%20chemical%20formulas,4%E2%8B%85H2O. 

Doesn't do variable occupancy - I guess it depends what is required.

ml-evs:
Trouble is this part of the code is JavaScript (though I'm sure a similar package exists) -- we could do the parsing on the server side but it would need another round-trip so probably want to avoid it

Issue #208, opened 2022-12-07
Allow more flexibility in chemical formula so it displays non-perfect formulas


jdbocarsly:

jdbocarsly:
Further discussion on #634

PR #743, opened 2024-05-17
Add simple component test for `ChemFormInput`

ml-evs:
Just testing this out for now, but otherwise could be useful for the development of #634 & #208


BenjaminCharmes:
I fixed the cypress tests on the components. It should work now. 

For this particular test, this still doesn't work because:
- By default, “editable” is set to false, so the input does not appear when cypress mounts the component.
- Chemical formulas don't format properly (Na3 stay Na3 and not Na<sub>3</sub>

ml-evs:
Hmmm, not sure if I broke this again @BenjaminCharmes with my rebase + changes! I was just trying to get it to also show the e2e test results in GitHub again, but now the tests are back to failing because the component is not displayed. Perhaps you could work some magic again when you get a moment?

PR #744, opened 2024-05-17
Add gravatar instructions and tweak account settings modal

ml-evs:
This PR tweaks the account settings UI to show the user's avatar (bigger), and dynamically reacts to them setting contact emails (which we should validate at some point). There's also instructions of how to register on gravatar, and some minor style tweaks


jdbocarsly (approved):
very nice

PR #750, opened 2024-05-27
Add `API_URL` as a `meta` tag in HTML header of all UI responses

### LINKED ISSUE #637: Add JSON response from web app when main page is requested with `Content-Type: application/json`
This would be very helpful for resolving the underlying API URL for users of the Python API. They may only "know" about e.g., `datalab.odbx.science` but the API will need to be passed `api.odbx.science`.

The web app should therefore have a JSON version of its root page that simply returns e.g., `{"api_url": "<datalab_api_url>"}` (and potentially some other metadata, like the description/owner of the datalab).

ml-evs:
This will allow the API package to smoothly redirect to the correct API URL (see https://github.com/datalab-org/datalab-api/commit/97a654a6e72e6d7f1cd2fbf766e9c526831e71ec).

Closes #637


ml-evs:
This is working well in the client, will merge

PR #782, opened 2024-06-19
Raman map

elbee99:
#781 Added basic 2D Raman mapping plotting


ml-evs:
Hi @elbee99, sorry for the delay on this, we've had so many changes in the meantime that I've transplanted your commits to another PR in #936. I'm happy to merge it and get it out there for feedback, but will add a little warning that appears in the UI about the image alignment until we can be sure!

PR #784, opened 2024-06-20
Improve sample table component

### LINKED ISSUE #254: Improve sample table component
We can use a prebaked table component that allows client-side filtering.

Desired features (please edit and expand):

- [x] All existing features of "old" sample table
- [ ] Pagination
    - [x] Client-side for now, but eventually server side too
- [x] Client-side search (free-text)
- [ ] Client-side filtering, e.g., categorical columns
- [x] Sortable columns
- [ ] Selectable/hideable columns client-side
- [x] Collapsible rows with more info
- [x] Selectable rows with actions available, e.g., delete, or add to collection
- [ ] Support for sorting/filtering derived data types like datetime
- [ ] Dynamic columns based on the API responses

BenjaminCharmes:
Close #254 

New sample table component using prime-vue


ml-evs (commented):
Looking good!

As we've discussed, let's:

- [x] Remove the column selector for now
- [x] Remove `date` from free text search
- [x] Add filter button for `type` column
- [ ] Comment about expandable row for now, until we can put real data in it
- [ ] See how easy it is to add clickable row for sample page

Minor things that can wait:

- [ ] Hiding sort arrows until hovered
- [ ] Move add item buttons to same level as deleted selected
- [ ] Think about other operations like "deleted selected" that could be grouped, e.g., "add to collection"

BenjaminCharmes:
Almost everything we've discussed should be working by now.
I'll look at adding features like "add to collection" but I think you can already review this PR when you have time.

webapp/src/components/SampleTablePrime.vue
```diff
@@ -0,0 +1,275 @@
...
+      if (
+        event.originalEvent.ctrlKey ||
+        event.originalEvent.metaKey ||
+        event.originalEvent.altKey
+      ) {
+        window.open(`/edit/${row.item_id}`, "_blank");
+      } else {
+        this.$router.push(`/edit/${row.item_id}`);
+      }
+    },
```
ml-evs, line 229: Empirically there is something strange going on here, when click on an ID (specifically) in the table, it opens in a new tab, and also tries to open it in the old tab (but fails with some JS error)

webapp/src/components/SampleTablePrime.vue
```diff
@@ -0,0 +1,275 @@
...
+          <InputText v-model="filterModel.value" type="text" />
+        </template>
+      </Column>
+      <Column field="creatorsList" header="Creators" sortable>
+        <template #body="slotProps">
+          <Creators :creators="slotProps.data.creators" />
+        </template>
+        <template>
+          <InputText v-model="filterModel.value" type="text" />
+        </template>
```
ml-evs, line 118: Just to note that the creators filter doesn't appear for me

webapp/src/components/SampleTablePrime.vue
```diff
@@ -0,0 +1,275 @@
...
+          <InputText v-model="filterModel.value" type="text" placeholder="Search by ID" />
+        </template>
+      </Column>
+      <Column field="type" header="Type" sortable>
+        <template>
+          <InputText v-model="filterModel.value" type="text" />
+        </template>
+        <template #filter="{ filterModel }">
+          <InputText v-model="filterModel.value" type="text" placeholder="Search by type" />
+        </template>
```
ml-evs, line 81: This also doesn't seem to do anything for me; compared to the one based on ID

ml-evs (commented):
Hi @BenjaminCharmes, sorry for the delay on reviewing this! We can catch up about it next week, but just left a few comments of what is empirically working/not working for me atm.

webapp/src/router/index.js
```diff
@@ -33,6 +34,7 @@ const routes = [
     component: Equipment,
   },
   { path: "/next/samples", name: "samples-next", alias: "/next", component: SamplesNext },
+  { path: "/prime/samples", name: "samples-prime", alias: "/prime", component: SamplePrime },
```
ml-evs, line 37: Let's use prime as the new next

webapp/src/components/SampleTablePrime.vue
```diff
@@ -0,0 +1,272 @@
...
+          <div class="button-left">
+            <button class="btn btn-default" @click="createItemModalIsOpen = true">
+              Add an item
+            </button>
+            <button class="btn btn-default ml-2" @click="batchCreateSampleModalIsOpen = true">
+              Add batch of samples
+            </button>
+          </div>
+          <div class="button-right d-flex">
+            <Button
```
ml-evs, line 42: Let's use the normal bootstrap button for this for now

BenjaminCharmes:
pre-commit.ci autofix

ml-evs (dismissed):
Awesome work @BenjaminCharmes! I'll commit my one tidying comment and merge this so we can play around with it.

We can then open a new issue to track feedback/style tweaks we want to make here before promoting it to the default table.

ml-evs (approved):

webapp/src/views/SamplePrime.vue
```diff
@@ -0,0 +1,46 @@
...
+</script>
+
+<style scoped>
+.fade {
+  opacity: 0;
+  transition: opacity 0.15s linear;
+}
+.fade.show {
+  opacity: 1;
+}
```
jdbocarsly, line 33: I don't think this is doing anything (these styles are use in Modal.vue, but I think they are copied to several of our pages, Equipment.vue, Samples.vue, etc... . Probably my mistake a while ago. It would be nice if our linter could flag unused styles)

jdbocarsly (approved):
looks good!

PR #797, opened 2024-07-11
Add ability to add batch of cells

### LINKED ISSUE #396: Suggestion: Allow creation of cells through batch UI
Suggested by @veronika-sedajova

jdbocarsly:
Currently, the "batch add" feature only allows for adding samples. This PR adds the ability to also add cells, with a template option for positive electrode, electroylte, and negative electrode.

- [x] Functionality for adding a batch of cell and choosing electrodes
- [x] Figure out how to cleanly label positive electrode, electrolyte, and negative electrode fields
- [x] Rename "Add a batch of samples" -> "Add a batch of items"
- [x] Rename sample -> item in BatchCreateSampleModal.vue
- [x] Turn on `taggable` in the `ItemSelect` components in the batch sample modal 
- [x] update e2e tests
- [x] add test(s)

Closes #396

### DIFF
--- a/pydatalab/pydatalab/routes/v0_1/items.py
+++ b/pydatalab/pydatalab/routes/v0_1/items.py
@@ -396,17 +396,16 @@ def _create_sample(
                 "negative_electrode",
                 "electrolyte",
             ):
-                if copied_doc.get(component):
-                    existing_consituent_ids = [
-                        constituent["item"].get("item_id", None)
-                        for constituent in copied_doc[component]
-                    ]
-                    copied_doc[component] += [
-                        constituent
-                        for constituent in sample_dict.get(component, [])
-                        if constituent["item"].get("item_id", None) is None
-                        or constituent["item"].get("item_id") not in existing_consituent_ids
-                    ]
+                existing_consituent_ids = [
+                    constituent["item"].get("item_id", None)
+                    for constituent in copied_doc[component]
+                ]
+                copied_doc[component] += [
+                    constituent
+                    for constituent in sample_dict.get(component, [])
+                    if constituent["item"].get("item_id", None) is None
+                    or constituent["item"].get("item_id") not in existing_consituent_ids
+                ]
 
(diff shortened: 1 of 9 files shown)


BenjaminCharmes (approved):
Perfect, everything works and looks nice. 👍
A small difference with "Add an item" is that Cell components are not written as optional in "Add batch of cells", If you want to add it.

jdbocarsly:
> A small difference with "Add an item" is that Cell components are not written as optional in "Add batch of cells", If you want to add it.

This is a good point, but I just tried adding in (optional) and it makes it a little bit messy, especially on small fields. I think that people will be able to figure it out pretty well, but if we get questions from users about this we can revisit in the future. 

In general, I think this component probably needs to be redesigned to be more flexible/powerful at some point, and also mobile-friendly.

pydatalab/pydatalab/routes/v0_1/items.py
```diff
@@ -396,17 +396,16 @@ def _create_sample(
...
+                existing_consituent_ids = [
+                    constituent["item"].get("item_id", None)
+                    for constituent in copied_doc[component]
+                ]
+                copied_doc[component] += [
+                    constituent
+                    for constituent in sample_dict.get(component, [])
+                    if constituent["item"].get("item_id", None) is None
+                    or constituent["item"].get("item_id") not in existing_consituent_ids
+                ]
```
ml-evs, line 408: I'm going to write and push a quick test for this

ml-evs (approved):
Looks great, thanks @jdbocarsly! Will push an additional test to this PR for the Python change, but otherwise this is ready to go.

I'm going to test it on the demo server with some legacy data immediately after merging, and if all is good I'll make a release.

pydatalab/pydatalab/routes/v0_1/items.py
ml-evs: Done in 2e29bc3

PR #801, opened 2024-07-11
Report runtime server config at `/info` and use this in UI

ml-evs:
to e.g., disable unconfigured login mechanisms, and more in future.

This PR also tidies up the login details slightly, preventing duplicated calls to `/get-current-user` when not rendering the account settings block.


BenjaminCharmes (approved):
Perfect! 👍

PR #805, opened 2024-07-12
Development docker, pre-commit and eslint updates & refactoring

### LINKED ISSUE #800: Run `yarn lint` in pre-commit
Now that we use the stricter settings turned on for eslint, we should have run `yarn lint` (or otherwise run eslint directly) in pre-commit so that it can do the auto-formatting for us. Currently we run only `prettier`.

ml-evs:
This PR:

- makes a few local docker dev tweaks (enables hot reloading, and pins docker-compose project name to allow for better git worktree use)
- tweaks the JSON schema generation hook, so that it only runs when models actually change (cf. #785)
- updates ESLint and babel, and refactors the config so that it can be used as a pre-commit hook
- updates several packages

Closes #800.


ml-evs:
Going to merge this so we can get some other stuff in, but think at least @jdbocarsly will want to quickly look over this

.pre-commit-config.yaml
```diff
@@ -31,10 +35,23 @@ repos:
...
+          - eslint-config-prettier@8.10.0
+          - eslint-plugin-prettier@5.1.3
+          - eslint-plugin-vue@8.0.3
+          - "@vue/cli-plugin-babel@5.0.8"
+          - "@vue/cli-plugin-eslint@5.0.8"
+          - "@vue/eslint-config-prettier@8.0.0"
+          - eslint-plugin-cypress@3.3.0
+          - "@babel/core@7.24.8"
+          - "@babel/eslint-parser@7.24.8"
+          - "@babel/plugin-transform-export-namespace-from@7.24.7"
```
ml-evs, line 54: It's a bit annoying that we have to do this, but alas this is the recommended approach.

One thing we might consider in the future is writing a little `invoke` script that grabs the installed versions of these packages from `yarn.lock` and updates them (replacing the equivalent `pre-commit autoupdate` functionality). However, as these rarely change I don't think its worth it for now.

.docker/app_dockerfile
```diff
@@ -6,7 +6,7 @@ WORKDIR /app
 
 EXPOSE 8081
 
-COPY webapp/package.json webapp/yarn.lock ./
+COPY webapp/package.json webapp/yarn.lock .
```
ml-evs, line 9:
```suggestion
COPY webapp/package.json webapp/yarn.lock ./
```

jdbocarsly (dismissed):
looks good- thanks for doing this!

ml-evs:
> Edit: actually, it seems eslint isn't running locally for me when I change .vue or .js files (and presumably others). It is simply skipping them, and if I force it to run with `pre-commit run --all-files`, then it passes even when it shouldn't. I think it is not correctly detecting the files it should, but I can't figure out why.

My final commits seem to fix this, going to merge and keep an eye on it.

Issue #803, opened 2024-07-12
QR codes and barcodes for samples


ml-evs:
cc @PeterKraus @jdbocarsly (and maybe @MJCliffe?)

We have often discussed labelling physical samples with QR codes/bar codes. This issue can be used to outline the work needed and then potential extensions.

- [ ] Resolve sample pages by refcode in the  API and UI

then either:

- [ ] We add a configurable "labelling" block that allows you to choose QR code size, style etc. for printing, and any extra info you want to be on the label
- [ ] or, we hardcode the label style at the level of the server, and always show that label on the item page. How it prints at different scales will be up to the user.

# Open questions

- How should permissions work in this case? We could embed a read-only token for the QR code, so if someone can scan your samples, then they should at least be able to see what it is/was.

# Extensions

## Link resolvers

How long do these barcodes need to last, and what info should we try to encode in them in case the datalab is gone/dead? I think there are two options:

- [ ] Stuff as much of the JSON data into the QR code itself, e.g., formula, hazards and so on, so that it at least can be resolved by a smart QR code scanner if datalab is dead. I don't know how smart a normal mobile browser scanner is, so this would need some experimenting.
- [ ] Devise a "datalab federation" link resolver, so that the QR code only points to e.g., `datalab.ly/concat:refcode`, then a simple server at datalab.ly resolves that `concat:` currently exists at `datalab.concatlab.eu`. This could allow us to have an interesting "off-ramp" where we statically host an entire datalab's JSON response to allow this resolution to work "forever".

## Printing labels from datalab

- [ ] Attach label printer to the datalab server itself (via tunnels or whatever), so that a "print label" button always goes to a printer in the lab, with some predefined sizes.
- [ ] I think @jdbocarsly was interested in the reverse, where the act of printing a label actually creates the stub datalab page and reserves the ID.

## Barcode scanning in datalab itself

- [ ] Given the read-only permissions above, it would be nice to use QR codes to also ingest data into datalab. e.g., you exchange samples with a collaborator, they can scan the QR code via the datalab browser (either uploading the image, or using the camera functionality), and then datalab can make the API call to the other person's datalab and ingest it, maintaining various backlinks. In this case, it could be very nice to even communicate to the "remote" datalab that the sample has been "received" and that it is now locked for editing until they rescan the QR code at original end (or alternatively, until they rescan it, any changes made remotely must be pushed back to the local datalab).

PeterKraus:
All good points.
[...]

PR #807, opened 2024-07-13
Add routes for resolving items by refcode

ml-evs:
First steps on #803 

~This PR adds `/items/by-refcode/<refcode>` and `/items/by-id/<id>` endpoints that essentially replace `/get_item_data/<item_id>`. I'm not entirely happy with the routes yet, and would like to find some examples of how other APIs deal with what are essentially multiple canonical IDs. Having `by-id` and `by-refcode` feels natural, but perhaps simply `/items/refcode/<refcode>` is more extensible and compatible with e.g. JSON:API to other fields in the future.~

Went with this in the end: having `/items/<refcode>` be the canonical URL is appealing, with the "legacy" endpoint being the only way to refer to items by item ID -- this would also allow us to promote refcode at some point to simply be the top-level `id` field in JSON:API parlance, which has some nice properties.

### DIFF
--- a/pydatalab/pydatalab/permissions.py
+++ b/pydatalab/pydatalab/permissions.py
@@ -1,6 +1,7 @@
 from functools import wraps
 from typing import Any, Dict
 
+from bson import ObjectId
 from flask import request
 from flask_login import current_user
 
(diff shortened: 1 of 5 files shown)


BenjaminCharmes (approved):
Looks great !

PR #808, opened 2024-07-13
Add QR code generation and scanning to UI

### LINKED ISSUE #803: QR codes and barcodes for samples

ml-evs:
Closes #803.

This PR:

- [x] Adds a QR code icon to all refcode displays, which, when clicked, opens a modal showing the QR code for that sample item, with a WIP print button that create a dummy window containing just the QR code.
- [x] The QR codes point to `{{ QR_CODE_RESOLVER_URL }}/items/<refcode>`, which can be configured. The refcode itself is always parseable from this URL, so datalab-native scanners (such as that described below) can resolve the item without needing to go via the resolver.
- [x] There is also a "Scan QR Code" button at the top of the sample table page (for now) which opens a camera stream and provides file upload for providing a datalab QR code.
- [x] API route argument is added for `redirect-to-ui`, where the `/items` endpoint can point itself to the UI, if configured. This is the default redirect used by the purl system.

Potential follow-ups:

- [ ] This scanner should do the following:
    - [ ] Detect if the sample is from this deployment, and if so, redirect to the edit page directly.
    - [ ] Follow the resolver link with JS and get the "real" underlying URL, then present it to the user to click on, if it believes it to be a datalab QR code.
    - [ ]  If it does not believe it to be a datalab QR code, do the same as above but with a giant warning.
- [ ] QR codes should also embed tokens that allow for access of non-private/safety data. This will need updates to the API too.
- [ ] The barcode print itself should also allow some info to be printed automatically.


webapp/src/resources.js
```diff
@@ -24,6 +24,11 @@ export const API_URL =
   process.env.VUE_APP_API_URL != null ? process.env.VUE_APP_API_URL : "http://localhost:5001";
 export const API_TOKEN = process.env.VUE_APP_API_TOKEN;
 
+export const QR_CODE_RESOLVER_URL =
+  process.env.VUE_APP_QR_CODE_RESOLVER_URL != null
+    ? process.env.VUE_APP_QR_CODE_RESOLVER_URL
+    : "https://purl.datalab-org.io";
```
ml-evs, line 30: I think this is the bit I'm still unsure about; think we need another option for whether the link resolver should be turned on; or even a server value for "is this server in the federation", and if not, just use the datalab's URL as the resolver

ml-evs (commented):
This this is ready when you are @jdbocarsly and @BenjaminCharmes -- I have one comment we can discuss regarding configuration (but probably that can be addressed in another PR before release)

BenjaminCharmes (approved):
It looks really good! 👏

jdbocarsly (approved):
looks good! excited to start using this and build on top of it

PR #809, opened 2024-07-15
Rebrand repo to the group-agnostic gh organisation

ml-evs:
Just updating GH links for the new organisation, even though the old one will continue to work.


.github/workflows/ci_automerge_dependabot.yml
```diff
@@ -1,30 +0,0 @@
-# Lifted from https://github.com/Materials-Consortia/optimade-python-tools/blob/master/.github/workflows/ci_automerge_dependabot.yml
```
ml-evs, line 1: This workflow is entirely defunct

PR #813, opened 2024-07-18
Add better layouts and configuration options to the Item Graph

jdbocarsly:
Todo: 
- [x] Add menu for graph configuration options
- [x] Add ability to ignore user-selected items and collections (use case: to avoid the graph clumping around trivial starting materials that are used in a large number of items)
- [x] Add option to show starting material names instead of ids
- [x] Add option to show sample/cell names instead of ids
- [x] Improve choice of graph algorithms to improve speed
- [x] Fix error `ctr is not defined` when the graph is generated, then returned to.
- [x] Make item and collection ignore dynamic (instead of regenerating the graph each time)
- [ ] ~Add button to not show any collections~ (deferred)
- [x] Dynamically change labels (instead of regenerating the graph)
- [ ] ~Some options for coloring/styling the nodes~ (deferred)
- [ ] ~pre-populate the ItemSelect with the highest rank starting materials, since these are the most likely to need to be removed~ (deferred)


ml-evs (approved):
This all looks great to me! Obviously hard to test in any meaningful way, as discussed we should move to moving more of this logic to the backend and testing for valid networks that way. I've deployed this on datalab.bocarslygroup.com and it works significantly better (like 10-100x faster) than our existing solution, so let's get it in!

PR #814, opened 2024-07-18
Upgrade MongoDB to v8

ml-evs:
Our MongoDB version is now out of support -- this PR updates the dockerized version, though some compatibility checks will need to be made with the tools for dumping/restoring data.

This is way overdue but needs to be handled carefully. Part of the delay is that MongoDB cannot be seamlessly upgraded by more than one version at a time... We should make some efforts to avoid accidental data loss from other people doing this upgrade, e.g., automatically backing up the server on startup.

Upgrading from MongoDB v3 -> v8 needs to be done carefully via `mongodump` and `mongorestore`. I'll add some docs for this, and probably some automation in the ansible playbooks to avoid nuking old databases.


.docker/mongo_dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM mongo:3.6-xenial
+FROM mongo:7.0-jammy
```
ml-evs, line 1:
```suggestion
FROM mongo:4.4-focal
```

.github/workflows/ci.yml
```diff
@@ -46,7 +46,7 @@ jobs:
 
     services:
       mongo:
-        image: mongo:3
+        image: mongo:7
```
ml-evs, line 49:
```suggestion
        image: mongo:4
```

PR #820, opened 2024-07-29
[WIP] Adding support for Palmsens and cyclic voltammetry

### LINKED ISSUE #462: Adding support for Palmsens
We (@gahorwitz and I) would like to support the Palmsens cyclers. They are bluetooth-connected so can be used directly inside a glovebox. Some more info:

- They have only one channel
- The software they require is called [PSTrace](https://www.palmsens.com/software/ps-trace/) (non-free) which also has a PythonSDK (semi-free, requires PSTrace to run).
- It supports various techniques, CV, EIS, etc.
- Currently people open their data in PSTrace as a "session" and then export the relevant parts to .csv or Excel files.

Will add more detail shortly

BenjaminCharmes:
Related to #462 

Thanks to @gahorwitz code from #502 this PR allows plotting in Echem block from .csv files of Palmsens cyclers.

Need to be discussed in more detail about specific needs to be added.


ml-evs:
superseded by #1696

PR #821, opened 2024-07-29
Fix logo link taking full screen width

### LINKED ISSUE #798: Disable logo banner homepage click

ml-evs:
Bit of an inline CSS hack, but closes #798


jdbocarsly (approved):
Works well! If you want it to be less of an inline CSS hack, you could always move these styles to two class, and then it would be a normal CSS hack... but it really doesn't make a difference IMO

PR #830, opened 2024-08-02
User select and sharing of sample page with creators

ml-evs:
This PR adds a user select and toggleable editable option to the creator list for items.

It should prevent the user from removing themselves, and should potentially prevent the user from removing other people (i.e. append-only share), or some more tortured logic where the "first" creator in the list takes precedence.

This also adds a new API route `PATCH /items/<refcode>/permissions` for this use case, with some tests.

Works similarly to the toggleable collection block, except that API call is made when the editing state is completed, and has a confirmation dialog.

Some outstanding things (for this PR or otherwise):

- [x] Error handling when the API update fails, e.g., if the creator tries to remove themselves. For me, the the error seems to get caught by the dev server before it gets back to the component to reset the value.
- [ ] Cypress testing... tricky since we're not ever logged in during the current e2e tests

### DIFF
--- a/webapp/src/components/Creators.vue
+++ b/webapp/src/components/Creators.vue
@@ -26,7 +26,7 @@ export default {
     },
     showNames: {
       type: Boolean,
-      default: false,
+      default: true,
       required: false,
     },
     showBubble: {
--- a/pydatalab/src/pydatalab/routes/v0_1/items.py
+++ b/pydatalab/src/pydatalab/routes/v0_1/items.py
@@ -205,13 +205,9 @@ def creators_lookup() -> Dict:
         "from": "users",
         "let": {"creator_ids": "$creator_ids"},
         "pipeline": [
-            {
-                "$match": {
-                    "$expr": {
-                        "$in": ["$_id", {"$ifNull": ["$$creator_ids", []]}],
-                    },
-                }
-            },
+            {"$match": {"$expr": {"$in": ["$_id", {"$ifNull": ["$$creator_ids", []]}]}}},
+            {"$addFields": {"__order": {"$indexOfArray": ["$$creator_ids", "$_id"]}}},
+            {"$sort": {"__order": 1}},
             {"$project": {"_id": 1, "display_name": 1, "contact_email": 1}},
         ],
         "as": "creators",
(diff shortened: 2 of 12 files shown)


webapp/src/components/Creators.vue
```diff
@@ -3,7 +3,7 @@
     <span v-if="showBubble">
       <UserBubble v-if="showBubble" :creator="creator" :size="size" />
     </span>
-    <span v-if="showNames && creator.display_name" class="display-name">
+    <span v-if="showNames && creator.display_name && creators.length === 1" class="display-name">
       {{ creator.display_name }}
       <span v-if="creator !== creators[creators.length - 1]">,</span>
```
jdbocarsly, line 8: It looks like now line 8 will never be reached, since the names will only be shown if only a single creator is there. If multiple creators are provided, and "showNames" is True, no names will be shown, which is confusing (assuming I'm reading this right)

I think it would be better to leave the functionality to show multiple creator names in. Then, cases where we want to show the name only if creators.length===1 can either be handled by adding another prop 'showNameIfOnlyOneCreator`, or just handled whenever creators is called:
```<Creators :creators="ItemCreators" :size="36" showNames="ItemCreators.length === 1"/>```

Comment (jdbocarsly) [reply webapp/src/components/Creators.vue:8]: This also allows us to do things like: `<Creators :creators="ItemCreators" :size="36" showNames="ItemCreators.length <= 3"/>

File: webapp/src/components/Creators.vue [hunk 2]
```diff
@@ -3,7 +3,7 @@
     <span v-if="showBubble">
       <UserBubble v-if="showBubble" :creator="creator" :size="size" />
     </span>
-    <span v-if="showNames && creator.display_name" class="display-name">
+    <span v-if="showNames && creator.display_name && creators.length === 1" class="display-name">
```
ml-evs, line 6:
```suggestion
    <span v-if="showNames && creator.display_name" class="display-name">
```

ml-evs:
This has been deployed on @PeterKraus's deployment for a few weeks now and seems to be working well (pending future enhancements to also apply it to collections). Since all I've done since then is add data-testid's to get the e2e tests working, I'm going to take responsibility and merge this now...

PR #831, opened 2024-08-04
Rework the `/delete-sample` route and enable soft deletion of items

ml-evs:
By default, delete will now simply set a flag `_deleted` on the database entry, so that it can be restored in future. This is important for e.g., labelled samples and maintaining old relationships to samples.


ml-evs:
Have tried to rebase this too, but needs more time in the oven. Will revisit for the next release cycle.

PR #844, opened 2024-08-19
Update Pipfile.lock

ml-evs:
c.f. #604.

The lockfile here and #604 should be the same (modulo one additional day of released deps), which should be verified before merging #604.


ml-evs:
As hoped, this PR, which preceded #604 and used the legacy `pipenv` approach for updating deps, does not make any changes beyond updating deps that have been released since #604. This can be considered closed by #604.

PR #850, opened 2024-08-20
Fix docs formatting and update versioning scheme

ml-evs:
The setup-tools-scm version was not being picked up outside of docker as the `.git` folder is one level up. This PR fixes that, adds a simple test, and also makes some docs formatting fixes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,6 +59,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -60,6 +64,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
```
ml-evs, line 65:
```suggestion
      - uses: actions/checkout@v4
        with:
          fetch-depth: 0
```

PR #857, opened 2024-08-22
Refactor login components

### LINKED ISSUE #768: Login/register button should not render until the `/get-current-user/` call has completed
We should replace the login button with a spinner until the `/get-current-user` call to the API has finished, and have better state if it fails.

Similarly, I often see two calls to `/get-current-user` being made when the page first opens, we should hunt these down and make sure the second one simply waits for the store to be populated with user info. Once stored once, we should probably also use the store directly without making another call to `/get-current-user`.

ml-evs:
Closes #768.

Also adds a patch that helps development cfgs by allowing non-fatal vue/prettier errors to appear as warnings.

### DIFF
--- a/webapp/src/components/LoginDetails.vue
+++ b/webapp/src/components/LoginDetails.vue
@@ -1,108 +1,66 @@
 <template>
-  <div class="row justify-content-center pt-3">
-    <GetEmailModal v-model="emailModalIsOpen" />
-    <template v-if="user != null">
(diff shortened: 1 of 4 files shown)


BenjaminCharmes (approved):
Looks good !

The only small difference, I don't know if it's intentional, is that the dropdown remains open when you click on “Account Settings”, even after closing the modal.

PR #862, opened 2024-08-30
Improving prime Datatable

BenjaminCharmes:


webapp/src/components/SampleTablePrime.vue
```diff
@@ -10,7 +10,7 @@
       selection-mode="multiple"
       :value="samples"
       paginator
-      :rows="10"
+      :rows="20"
       :rows-per-page-options="[10, 20, 50, 100]"
```
BenjaminCharmes, line 14: Hey @jdbocarsly, as requested yesterday, this is where you can easily modify the number of items per row, or the different choices offered in the dropdown. Don't hesitate to let me know if you want me to change anything.

webapp/cypress/component/NavbarTest.cy.jsx
```diff
@@ -116,7 +116,7 @@ describe("Navbar", () => {
     cy.url().should("include", "/starting-materials");
 
     cy.contains("Equipment").click();
-    cy.url().should("include", "/equipment");
+    cy.url().should("include", "/equipments");
```
ml-evs, line 119:
```suggestion
    cy.url().should("include", "/equipment");
```

I think the original is correct, though the conventions of pluralising API routes do get confusing for words that are self-plural...

webapp/src/components/AddToCollectionModal.vue
```diff
@@ -70,7 +75,13 @@ export default {
           await addItemsToCollection(collectionId, refcodes);
         }
         this.$emit("itemsUpdated");
-
+        if (this.itemsSelected.some((item) => item.type === "samples")) {
+          getSampleList();
+        } else if (this.itemsSelected.some((item) => item.type === "startingMaterials")) {
+          getStartingMaterialList();
+        } else if (this.itemsSelected.some((item) => item.type === "equipments")) {
```
ml-evs, line 82:
```suggestion
        } else if (this.itemsSelected.some((item) => item.type === "equipment")) {
```

webapp/src/components/CollectionsTable.vue
```diff
@@ -0,0 +1,42 @@
+<template>
+  <DynamicDataTable
+    :columns="collectionColumn"
+    :data="collections"
```
ml-evs, line 4: This looks great, though I have a mild suggestion that we should retain the same component name as previously (so that the git commit history is retained on the file, essentially)

webapp/src/components/DynamicButtonDataTable.vue
```diff
@@ -0,0 +1,201 @@
...
+      <button
+        v-if="dataType === 'startingMaterials' && editableInventory"
+        class="btn btn-default"
+        :disabled="itemsSelected.length === 0"
+        @click="$emit('open-create-item-modal')"
+      >
+        Add a starting material
+      </button>
+      <button
+        v-if="dataType === 'equipments'"
```
ml-evs, line 41:
```suggestion
        v-if="dataType === 'equipment'"
```

You get the idea, I think all of the `equipments` should be changed to `equipment` throughout.

ml-evs (commented):
Looking really good @BenjaminCharmes! Great to see how much old code we can get rid of without breaking anything! 

Minor initial comments: I think we should retain the old component names just for consistency (of the git history more than anything). Plus, the name `equipment` should be kept throughout to match the underlying model.

Sadly I'm still having issues with my local set up where the components within the table are not being used -- I'm still using node 18 locally, so perhaps this is just an a "me problem" but I'd like to at least find the reason before we merge this. Perhaps the minimum node version needs to be updated in `package.json` too.

I'll give this a more thorough review tomorrow and we can discuss on Friday.

webapp/src/components/CollectionsTable.vue
BenjaminCharmes: Do you mean “PrimeDataTable”?

ml-evs: No, sorry, I mean that `CollectionsTable.vue` should remain `CollectionTable.vue` -- I can make these tweaks

jdbocarsly (changes requested):
A few notes from today's chat:

1. The "select all" button is slow with lots of samples/inventory. Can we change the select button to only select the shown samples as a workaround? 
1. Each individual checkbox has to be clicked precisely to be clicked. Can we change the @ click even to be on the cell, rather than on the checkbox
1. Is there a way to make the filter symbol filled in when a filter is on?
1. ~~"Add a collection" from the CollectionSelect is great, but the spinner seems to be broken as it never goes away even once the collection is created~~
1. When you insert something into a collection, and then search for the same collection again, it gives you the option to create the collection again, and then causes an error. 
1. seems to be possible to "double add" something to a collection that was already in that collection, which causes an error when you try to access the page "relationships and collection mismatch"
1. Tests for checking whether components are rendered in the table
1. Test for adding to inventory when editable_inventory is turned on etc

pydatalab/pydatalab/routes/v0_1/collections.py
```diff
@@ -359,6 +359,8 @@ def add_items_to_collection(collection_id):
                     "relation": None,
                     "type": "collections",
                     "immutable_id": ObjectId(collection["_id"]),
+                    "item_id": None,
+                    "refcode": None,
```
ml-evs, line 363: Perhaps we should use the actual relationships model here? I can't use the suggestion but it would look something like

```
TypedRelationship(description="Is a member of", type="collections", immutable_id=ObjectId(collection["_id"]))
```

then all of the `None` fields will be set automatically. Perhaps we even want to use a specialisation of this for collection relationships in future, that sets the type and ID for you

ml-evs (approved):
:rock: :rocket:

jdbocarsly (approved):
looks great!

PR #865, opened 2024-08-31
Switch Python package to src-layout

ml-evs:
Switching to `src`-layout is increasingly helpful for plugin development and avoids misleading imports. The main knock-on effects are related to deployment, as well as some minor tweaks needed to launch the Python server.


BenjaminCharmes (dismissed):
Looks great ! 👍

PR #882, opened 2024-09-19
Attempt to enable code coverage for cypress

ml-evs:
Experimenting with cypress code coverage, doubt this will work for both e2e and component testing as simply as this...


ml-evs:
Not worth the hassle for now

PR #884, opened 2024-09-24
Add component tests for DataTable Prime components

### LINKED ISSUE #794: Component testing
I made some initial attempts at component testing in #743 but could not get the incantation right. It would be good to begin adding component tests every time we refactor a component, starting with the chemical formula.

Note to self to discuss this with @BenjaminCharmes soon ;)

### LINKED ISSUE #889: Initial issues to resolve with new data table
- [x] The "select all" button is slow with lots of samples/inventory. Can we change the select button to only select the shown samples as a workaround? 
- [x] Each individual checkbox has to be clicked precisely to be clicked. Can we change the @ click even to be on the cell, rather than on the checkbox
- [x] #1012
- [ ] ~Test for adding to inventory when editable_inventory is turned on etc~ (attempt to remove this anyway)
- [x] Check whether add to collection select on create new item modal still works
- [ ] ~Consider future refactor around how components are specified in the table, i.e., whether they can be mapped from the datalab server schema~ (delay for later refactor)
- [x] Wrapping of long columns
- [x] #973
- [x] #992 

1. ~~Is there a way to make the filter symbol filled in when a filter is on?~~
1. ~~"Add a collection" from the CollectionSelect is great, but the spinner seems to be broken as it never goes away even once the collection is created~~
1. ~~When you insert something into a collection, and then search for the same collection again, it gives you the option to create the collection again, and then causes an error.~~
1. ~~seems to be possible to "double add" something to a collection that was already in that collection, which causes an error when you try to access the page "relationships and collection mismatch"~~

_Originally posted by @jdbocarsly in https://github.com/datalab-org/datalab/pull/862#pullrequestreview-2318695346_

### LINKED ISSUE #1012: Tests for checking whether components are rendered in the table

BenjaminCharmes:
Closes #1012 
Closes #794 
Closes #889 

Add component test for:

- SampleTable
- CollectionTable
- EquipmentTable
- StartingMaterialTable

Add e2e test for:

- startingMaterial (Only for EDITABLE_INVENTORY false, test that the button 'Add a starting material' isn't displayed)


BenjaminCharmes:
I remove the starting_materials e2e test for now because the EDITABLE_INVENTORY is false by default. So we need to decide if we only test the starting_material page when we can't edit the inventory, or if we should test it when it's true (or both?)

ml-evs (approved):
LGTM! Let's park the `EDITABLE_INVENTORY` stuff as I think it will be refactored away soon anyway. Always useful to have more tests in any form!

PR #885, opened 2024-09-24
Add dynamic data table on collection page

### LINKED ISSUE #815: Collection table is currently broken
We've noticed that the `/next` sample tables (which are about to be replaced anyway) are currently broken in deployments, probably as a result of the recent linting and JS package overhaul in #805.

These should be replaced anyway by @BenjaminCharmes's work in #254 / #784, but we should make sure some simple tests get added soon.

### LINKED ISSUE #891: Collections table incorrect link to collection edit page
The current table assumes that all edit pages are under `/edit/`, this is currently not true for collections (should it be? -- will then need to decide which component to load), where it should be configured to be `/collections`.

ml-evs:
Closes #815
Closes #891.

Replaces the next table on collections with the new dynamic table, and:

- [x] Customises the edit page links on a per-table basis to allow it to work with the current collections routes (closes #891)
- [x] Allows you to disable the buttons and top search bar

### DIFF
--- a/webapp/src/components/CollectionTable.vue
+++ b/webapp/src/components/CollectionTable.vue
@@ -4,6 +4,7 @@
     :data="collections"
     :data-type="'collections'"
     :global-filter-fields="['collection_id', 'title', 'creatorsList']"
+    :edit-page-route-prefix="'collections'"
   />
 </template>
 
--- a/webapp/src/components/DynamicDataTable.vue
+++ b/webapp/src/components/DynamicDataTable.vue
@@ -43,7 +43,7 @@
       <template #empty> No data found. </template>
       <template #loading> Loading data. Please wait. </template>
 
-      <Column class="checkbox" selection-mode="multiple"></Column>
+      <Column v-if="showButtons" class="checkbox" selection-mode="multiple"></Column>
 
       <!-- <Column expander style="width: 5rem" /> -->
       <Column
--- a/webapp/src/components/DynamicDataTable.vue
+++ b/webapp/src/components/DynamicDataTable.vue
@@ -223,6 +223,14 @@ export default {
   methods: {
     goToEditPage(event) {
       const row = event.data;
+      let row_id = null;
+
+      // Check if the row has an item ID, otherwise default to collection ID
+      if (!row.item_id && row.collection_id) {
+        row_id = row.collection_id;
+      } else {
+        row_id = row.item_id;
+      }
 
       if (event.originalEvent.target.classList.contains("checkbox")) {
         return null;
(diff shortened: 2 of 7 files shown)


webapp/src/components/CollectionTable.vue
```diff
@@ -4,6 +4,7 @@
     :data="collections"
     :data-type="'collections'"
     :global-filter-fields="['collection_id', 'title', 'creatorsList']"
+    :edit-page-route-prefix="collections"
```
BenjaminCharmes, line 7: Should be a string ?

BenjaminCharmes, line 7: Should be a string ?

```suggestion
    :edit-page-route-prefix="'collections'"
```

webapp/src/components/DynamicDataTable.vue
```diff
@@ -222,9 +228,9 @@ export default {
         event.originalEvent.metaKey ||
         event.originalEvent.altKey
       ) {
-        window.open(`/edit/${row.item_id}`, "_blank");
+        window.open(`/${this.editPageRoutePrefix}/${row.item_id}`, "_blank");
```
BenjaminCharmes, line 231: Missing in props ?

```suggestion
props: {
  editPageRoutePrefix: {
    type: String,
    required: true,
  },
},```

File: webapp/src/components/DynamicDataTable.vue [hunk 3]
```diff
@@ -222,9 +228,9 @@ export default {
         event.originalEvent.metaKey ||
         event.originalEvent.altKey
       ) {
-        window.open(`/edit/${row.item_id}`, "_blank");
```
BenjaminCharmes, line 225: `editPageRoutePrefix` is missing inside `Props` ?

webapp/src/components/CollectionTable.vue
BenjaminCharmes, line 7: This should be a string and not an object ?

BenjaminCharmes (commented):
Looks really good. Just a few comments or things to modify maybe?

ml-evs:
Thanks Ben, think I've fixed everything

> Perhaps we should either remove the checkboxes column inside CollectionInformation, or add a button to delete multiple items from the collection?

I've removed the checkboxes for now, though we might think about a delete button that removes from just a collection, good idea!

> And a small eslint warning inside `CollectionInformation.vue` because "data" should be above "computed".

Just raised #895 for this

BenjaminCharmes (approved):
Very great ! :rocket:

ml-evs:
Think this is ready again when you are @BenjaminCharmes! I added a test to catch the case we discussed (also our first for the collections table/page generally).

BenjaminCharmes (approved):
It look really great, nice one ! 🚀

ml-evs:
> It look really great, nice one ! 🚀

Thanks, I'm going to merge this now but make a mental note that the tests locally might not work every time

PR #888, opened 2024-09-25
Fix some issues in the datatable

BenjaminCharmes:
- [x] The functionality of adding to a collection from the 'Create New Item' modal has been verified and is working as intended
- [x] The issue of requiring each individual checkbox to be clicked precisely has been resolved by changing the click event to be triggered on the entire cell instead of just the checkbox
- [x] The "select all" button now only selects the visible samples (should be much faster with large inventories)
- [x] The validation for collection_id has been implemented in the CollectionSelect component to ensure that a new collection is only created when a valid collection_id is provided

Addresses #889

### DIFF
--- a/webapp/src/components/CollectionSelect.vue
+++ b/webapp/src/components/CollectionSelect.vue
@@ -75,10 +75,12 @@ export default {
       },
     },
     collectionOrNewCollection() {
+      const valueSafe = Array.isArray(this.value) ? this.value : [];
+
       if (
         this.searchQuery &&
         !this.collections.some((item) => item.collection_id === this.searchQuery) &&
-        !this.value.some((item) => item.collection_id === this.searchQuery)
+        !valueSafe.some((item) => item.collection_id === this.searchQuery)
       ) {
         return [
           ...this.collections,
--- a/webapp/src/components/CollectionSelect.vue
+++ b/webapp/src/components/CollectionSelect.vue
@@ -9,7 +9,8 @@
     @search="debouncedAsyncSearch"
   >
     <template #no-options="{ searching }">
-      <span v-if="searching"> Collection already selected </span>
+      <span v-if="validationMessage" class="form-error">{{ validationMessage }}</span>
+      <span v-else-if="searching"> Collection already selected </span>
       <span v-else class="empty-search"> Search for a collection... </span>
     </template>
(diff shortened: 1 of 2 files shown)


webapp/src/components/CollectionSelect.vue
```diff
@@ -119,23 +126,47 @@ export default {
...
+            this.value = [
+              ...this.value.filter((item) => item.collection_id !== null),
+              {
+                collection_id: collection_id,
+              },
+            ];
+          }
+        } catch (error) {
+          console.error("Error:", error);
+          alert("An error occurred while creating the collection. Please try again.");
```
ml-evs, line 155:
```suggestion
          alert("An error occurred while creating the collection. Please check that your desired collection ID is valid.");
```

Just a minor tweak; I got this message when I tried an invalid collection ID that we weren't catching, other case might be that someone else made the collection in the meantime, so "try again" probably doesn't help as they will use the same input

webapp/src/components/CollectionSelect.vue
```diff
@@ -119,23 +126,47 @@ export default {
         loading(false);
       }, debounceTime);
     },
+    IDValidationMessage() {
+      if (this.searchQuery && !/^[a-zA-Z0-9_-]+$/.test(this.searchQuery)) {
+        return "ID can only contain alphanumeric characters, dashes ('-'), and underscores ('_').";
+      }
```
ml-evs, line 132:
```suggestion
      }
      if (/^[._-]/.test(this.searchQuery) | /[._-]$/.test(this.searchQuery)) {
        return "ID cannot start or end with punctuation";
      }
```

Just another case of the item ID validation that was missing. We actually now repeat this code 3/4 times in different modals (with the same typos in the error message, fixed here...), so we should wrap this up into a single utility function for IDs (perhaps with the ability to disable some via bool toggles depending on context). Perhaps you could throw this together after @BenjaminCharmes?

ml-evs (changes requested):
This looks great @BenjaminCharmes! Did something get fixed with the wrapping of long names too? It seems to work nicely for me now

I just had a couple of really minor suggestions below; I think I will commit these myself and then merge -- perhaps you could tidy up the ID validation in JS since we duplicate it in a few places now?

ml-evs (approved):
:+1:

Issue #852, opened 2024-08-20
Deprecate and remove `pipenv`


ml-evs:
Now that we have essentially removed `pipenv`, we should plan to deprecate the residual `Pipfile` and `Pipfile.lock` that remain for convenience. This should be done at the next-next minor version (v0.6.0) so people have time to migrate any hacky setups they have with `pipenv` right now. It's a bit hard to formally deprecate it, but perhaps an intermediate case would be to move the files somewhere new (whilst retaining them in the repo) before deleting in v0.6.0 along with the documentation sections.

PR #899, opened 2024-10-01
Update deps lockfiles and use `uv sync` in CI

ml-evs:
This PR resyncs the requirements files and `uv.lock`, and updates `Pipfile.lock`, with the aim for better allowing dependabot to deal with things.

Continues the process of #852


ml-evs:
I think these changes are all straightforward, so I'm going to merge this and let dependabot do its thing overnight.

PR #902, opened 2024-10-02
Add dropdown with actions buttons for selected items inside DataTable

### LINKED ISSUE #897: Make "delete selected", "add to collection" buttons on the dynamic data page all under one easy-to-extend dropdown menu
Currently we have separate buttons for "delete selected", and "add to collection" that appear on the dynamic data table pages when items are checked. As we add more advanced sample management options (like "print labels"), it would be nice to have all of these under a single dropdown.

BenjaminCharmes:
Closes #897 

This PR refactors the "Add to Collection" and "Delete Selected" actions into a PrimeVue SplitButton dropdown.
Also add a small fix to make cypress e2e tests work with dropdown.

### DIFF
--- a/webapp/src/components/DynamicButtonDataTable.vue
+++ b/webapp/src/components/DynamicButtonDataTable.vue
@@ -47,7 +47,7 @@
     <div class="button-right d-flex">
       <div class="dropdown">
         <button
-          id="selectedDropdown"
+          data-testid="selected-dropdown"
           class="btn btn-default dropdown-toggle"
           type="button"
           data-toggle="dropdown"
@@ -66,18 +66,24 @@
         >
           <a
             v-if="itemsSelected.length !== 0 && dataType !== 'collections'"
+            data-testid="add-to-collection-button"
             class="dropdown-item"
             @click="handleAddToCollection"
           >
             Add to collection
           </a>
-          <a v-if="itemsSelected.length !== 0" class="dropdown-item" @click="confirmDeletion">
+          <a
+            v-if="itemsSelected.length !== 0"
+            data-testid="delete-selected-button"
+            class="dropdown-item"
+            @click="confirmDeletion"
+          >
             Delete selected
           </a>
         </div>
       </div>
 
-      <IconField class="">
+      <IconField>
         <InputIcon>
           <i class="pi pi-search"></i>
         </InputIcon>
@@ -152,6 +158,13 @@ export default {
       localFilters: { ...this.filters },
       isSelectedDropdownVisible: false,
     };
+  },
+  watch: {
+    itemsSelected(newVal) {
+      if (newVal.length === 0) {
+        this.isSelectedDropdownVisible = false;
+      }
+    },
   },
   methods: {
     confirmDeletion() {
--- a/webapp/src/components/DynamicButtonDataTable.vue
+++ b/webapp/src/components/DynamicButtonDataTable.vue
@@ -24,52 +24,67 @@
       </button>
       <button
         v-if="dataType === 'collections'"
-        class="btn btn-default"
+        class="btn btn-default ml-2"
         @click="$emit('open-create-collection-modal')"
       >
         Create new collection
       </button>
       <button
         v-if="dataType === 'startingMaterials' && editableInventory"
-        class="btn btn-default"
+        class="btn btn-default ml-2"
         @click="$emit('open-create-item-modal')"
       >
         Add a starting material
       </button>
       <button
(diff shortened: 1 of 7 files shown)


BenjaminCharmes:
Hey @jdbocarsly, 
Here's a first draft to see what it might look like. Don't hesitate to tell me if you want me to modify the css style, or the text ... 

Screenshot 2024-10-02 at 12 01 48

webapp/src/components/DynamicButtonDataTable.vue
```diff
@@ -45,23 +45,15 @@
...
-      <button
-        v-if="itemsSelected.length > 0 && dataType != 'collections'"
-        class="btn btn-default"
-        :disabled="itemsSelected.length === 0"
-        @click="$emit('open-add-to-collection-modal')"
-      >
-        Add to collection
-      </button>
-      <button
+      <SplitButton
```
ml-evs, line 48: As we discussed, we should probably just use the bootstrap dropdown

jdbocarsly: agree, that's probably a bit lighter weight and more consistent in style (though the primevue one also looks nice)

ml-evs: We also couldn't come up with a good label for the button, but perhaps:

- disabling the button when nothing is selected
- labelling it with "Selected..." when nothing is selected
- label it with "n Selected" when n items are selected

BenjaminCharmes: All done! Let me know if you want to change the label.

webapp/src/components/DynamicButtonDataTable.vue
```diff
@@ -24,44 +24,64 @@
...
       <button
         v-if="dataType === 'equipment'"
-        class="btn btn-default"
+        class="btn btn-default ml-2"
         @click="$emit('open-create-equipment-modal')"
       >
         Add an item
       </button>
     </div>
     <div class="button-right d-flex">
```
jdbocarsly, line 47: I was confused about where this `button-right` class is defined, but I now see it seems to be on the DynamicButtonDataTable.vue

Ideally, we would use `<style scoped>` on each component and the styles from one component that way wouldn't affect the others. There is a way to use "deep" selectors on a scoped component which is occasionally needed, but in general having the styles defined in the component where they are used is best.

webapp/src/components/DynamicButtonDataTable.vue
jdbocarsly, line 1: Sorry for the pedantic comment, but can you rename this to: `DynamicDataTableButtons.vue` . This is based on the component naming convention for components that are tightly coupled to other components:
https://v2.vuejs.org/v2/style-guide/?redirect=true#Tightly-coupled-component-names-strongly-recommended

webapp/src/components/DynamicButtonDataTable.vue
```diff
@@ -24,44 +24,64 @@
...
+          data-testid="selected-dropdown"
+          class="btn btn-default dropdown-toggle"
+          type="button"
+          data-toggle="dropdown"
+          aria-haspopup="true"
+          aria-expanded="false"
+          :disabled="itemsSelected.length === 0"
+          @click="isSelectedDropdownVisible = !isSelectedDropdownVisible"
+        >
+          {{ itemsSelected.length > 0 ? `${itemsSelected.length} selected ` : "Selected ... " }}
```
jdbocarsly, line 59:
```suggestion
          {{ itemsSelected.length > 0 ? `${itemsSelected.length} selected... ` : "Selected ... " }}
```

I like this working, but think we should keeping the elippses on when items are selected so it will read "n selected...", showing that you can click on that button for more options.

jdbocarsly (changes requested):
Styling looks great! just a few minor changes

webapp/src/components/DynamicButtonDataTable.vue
BenjaminCharmes: Yeah, I must have forgotten to move it when the buttons didn't have their own components yet, thansks!

webapp/src/components/DynamicButtonDataTable.vue
BenjaminCharmes: Makes sense, thanks !

ml-evs (dismissed):
Looks great to me, thanks @BenjaminCharmes! It looks like you've incorporated all Josh's changes so I'm happy to merge once after adding one tiny change

ml-evs (approved):

PR #905, opened 2024-10-02
Add support for XRD files that have indeterminate headers (e.g., diffrac.eva export)

### LINKED ISSUE #903: Support for Bruker's diffrac.eva .xy files
They tend to contain metadata headers (with no distinguishing comment character) that we should parse and extract to the block description. Right now the files cannot be parsed at all, so we may also consider a more robust general `.xy` parser that attempts to skip rows until things work.

(cc @jmas5)

ml-evs:
Closes #903 by trying to load xy/xye files iteratively, ignoring any lines that do not fit the expect number of floats (and crashing out if the header is more than 10,000 lines).

cc @jmas5

Probably one for @jdbocarsly to check with some of his group's files to make sure nothing breaks (I added test cases, but you never know...)


pydatalab/pydatalab/apps/xrd/blocks.py
```diff
@@ -39,11 +39,35 @@ def load_pattern(
...
 
-        elif ext == ".xy":
-            df = pd.read_csv(location, sep=r"\s+", names=["twotheta", "intensity"])
-
         else:
-            df = pd.read_csv(location, sep=r"\s+", names=["twotheta", "intensity", "error"])
+            if ext == ".xye":
+                columns = ["twotheta", "intensity", "error"]
+            else:
+                columns = ["twotheta", "intensity"]
```
jdbocarsly, line 46: This will cause a problem with the I11 data files from previous beamlines, which have ".dat" extension but are 3 column xye. Though overall this gets messy because other data manufacturers use .dat to mean totally different formats unfortuantely

pydatalab/pydatalab/apps/xrd/blocks.py
```diff
@@ -39,11 +39,35 @@ def load_pattern(
...
+
+            # Try to parse the file by incrementing skiprows until all lines can be cast to np.float64
+            skiprows: int = 0
+            # Set arbitrary limit to avoid infinite loop; a header of 10,000 lines is unlikely to be useful
+            while skiprows < 10_000:
+                try:
+                    df = pd.read_csv(
+                        location, sep=r"\s+", names=columns, dtype=np.float64, skiprows=skiprows
```
jdbocarsly, line 54: Would it be easier to just use `comment="'"`?  

Annoyingly, some xy/xye-type files might use other comment characters like "#", though I don't believe any of the types we currently support do that...

jdbocarsly (changes requested):

pydatalab/pydatalab/apps/xrd/blocks.py
jdbocarsly: quick fix is just to do `if ext in (".xye", ".dat")`, which I guess should at least work with the file types we are currently working with

pydatalab/pydatalab/apps/xrd/blocks.py
ml-evs: That's the problem, the comment has to be a single char rather than a list. I thought about some other ways of processing this manually but I think this is the most robust generally -- also lets us extract the header as potential metadata though I admit I don't like the hardcoded limit of 10k

pydatalab/pydatalab/apps/xrd/blocks.py
ml-evs: Another option is to try detect how many columns there are if the filetype is not explicit

ml-evs: But thanks for the heads up about .dat...

pydatalab/pydatalab/apps/xrd/blocks.py
jdbocarsly: What about that idea you had of reading the files backwards?

pydatalab/pydatalab/apps/xrd/blocks.py
ml-evs: Actually this is straightforward, we can just always try to read as 3 columns and the `error`s will be set to NaN if missing

pydatalab/pydatalab/apps/xrd/blocks.py
ml-evs: I thought about trying to add it to pandas directly (via C) but I think its overkill, and way less efficient than this way if the header is small

pydatalab/pydatalab/apps/xrd/blocks.py
ml-evs: (right now we don't use the errors anywhere, but once we have a schema  we can just `null` out any NaNs)

PR #906, opened 2024-10-03
Extend version numbers in About page

### LINKED ISSUE #713: Extend version numbers in app about page
The about page/info endpoints need to be able to handle arbitrary strings for versions, so we can include e.g., release tags.

BenjaminCharmes:
Closes #713


ml-evs:
I'd like the version to be entirely dynamic and not reliant on someone committing the new version, we might need to look around a bit for packages that let us do this a la setuptools-scm

ml-evs (approved):
I wasn't happy with any libraries or the functionality of `npm version from-git` (which assumes you want to commit the tag, and doesn't let you control many things), so I've instead pushed a little node script that grabs the git version and sets it during build via the `VUE_APP_GIT_VERSION` env var (which can also be used to override it in cases where the `.git` info is not available). This basically fits the approach we take for Python now too.

The only weirdness is that the version in `package.json` (`0.0.0-git`) should now basically be ignored, and is just used as a fallback when the version has not been set by any other source.

PR #915, opened 2024-10-08
Add pictogram UI components for hazard codes and use them for inventory

### LINKED ISSUE #657: Better handling of hazards for starting materials
We have the ability to specify hazard codes in the starting material GUI, but no validation or special handling at the moment. We should add a nice UI showing pictograms.

BenjaminCharmes:
Closes #657 

Add a better way of handling hazards for starting materials, and display their pictograms in the UI.

@ml-evs edits:

- [x] Use toggleable Hazard component
- [x] Rearrange starting material page
- [x] Fix item graph for single starting material
- [x] Adjust datetime styling when no date provided


BenjaminCharmes:
- First version : Using a primevue multiselect where you can select one or more of the 9 GHS codes, and update the db with a string containing the GHS code(s).

https://github.com/user-attachments/assets/df1aeb12-ba91-438e-9522-a7e50b8a9460

- Second version (more versatile): Clicking on the ‘+’ icon opens a modal in which you can enter hazard information (basically, you can write anything), and it will then retrieve all the H codes and transcribe them into the corresponding GHS pictrograms. and store in the db everything that has been entered in the modal.

https://github.com/user-attachments/assets/52870ecf-912b-46a8-ad26-508f271e8e88

jdbocarsly:
First of all-- awesome! I'm excited about this. 

Sorry to ask, but can you try this with` vue-select` instead of the primevue multiselect? We use vue-select for all the other multiselects in the app (item, collection, etc.). You can use "taggable" to allow you to add multiple and I think it would be more visually and UI consistent to stick with the same component.

webapp/src/components/ToggleableGHSHazardInformation.vue
```diff
@@ -0,0 +1,82 @@
+<template>
+  <div ref="outerdiv" class="h-100 form-group clickable" @click="isEditing = !isEditing">
+    <label id="startmat-hazards" class="clickable">
+      GHS Hazard Codes
+      <font-awesome-icon id="edit-icon" class="pl-1" icon="pen" size="xs" :fade="isEditing" />
+    </label>
+    <div>
+      <GHSHazardInformation
```
ml-evs, line 8: If no hazards, maybe add placeholder text saying no hazards recorded

ml-evs:
Combining @BenjaminCharmes's changes with mine (plus my other tweaks to the starting material page), here's the final video of hazard handling as it stands. I'm going to add some placeholder text that says "No hazards registered" plus a description of what to type, but mostly this will be synced from an external source (and Ben's code grabs any combination of H-codes inside the free text field).

https://github.com/user-attachments/assets/e8b7ff82-609e-4c72-b81f-d137be6f1b3d

webapp/src/components/SelectGHSHazardInformation.vue
jdbocarsly, line 1: Is this component used anywhere? I did like the idea of having a multiselect for this (though the free text field is more flexible when working with other platforms, like cheminventory)

webapp/src/components/ToggleableGHSHazardInformation.vue
jdbocarsly, line 1: This works really well, but I think we need to actually display the H codes at the same time of the pictograms. So, rather than have it be toggleable, just keep the text field below (or above) the pictograms. The reason is because the pictograms are not unique, and not every H code has a pictogram. So, the pictograms give a quick view of safety and handling, but don't tell the full story.

jdbocarsly (changes requested):
Looks and works great, but I think before merging, the toggleable functionality should be changed to show both H codes and pictograms at the same time, since pictograms aren't sufficient to describe the safety and handling conditions. Once that small change is made, we can merge and build on from there.

webapp/src/components/SelectGHSHazardInformation.vue
ml-evs: No, as with the modal, but I left them in case we want them in the future. Perhaps I'll remove them in my commit that changes the toggle, and then we can just remember they are here in the history...

ml-evs:
Final version for this iteration: show the full hazard notes all the time, its not toggleable at all, and it updates live as the user types in.

2025-04-04-122257_808x230_scrot

PR #919, opened 2024-10-09
Fix css issues in DynamicDataTable

BenjaminCharmes:
Related to #889

Fix css issues applied to primevue components after adding 'scoped' to styles
Also removes padding in the datatable


ml-evs (approved):
LGTM!

PR #921, opened 2024-10-11
Set up custom primevue theme

BenjaminCharmes:
Setting up a custom theme for primevue instead of applying global css to primevue components 

- Set transition time to 0.05s (default: 0.2s)
- Set the padding of datatable headers and cells to 0.5rem (default: offset between header and bodycell)
- Hides the sorting icon until the header is hovered
- Make the text in the cells appear on a single line (nowrap and overflow hidden)
- Make the filter icon change color when a filter is active
- Checkboxes color grey

### DIFF
--- a/webapp/src/components/DynamicDataTable.vue
+++ b/webapp/src/components/DynamicDataTable.vue
@@ -236,6 +236,15 @@ export default {
 
       if (event.originalEvent.target.classList.contains("checkbox")) {
         return null;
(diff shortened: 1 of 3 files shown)


BenjaminCharmes:
pre-commit.ci autofix

jdbocarsly (approved):
Looks great!! This will be really helpful to have, and its nice that it cleans up the css on the datatable component as well

ml-evs (approved):
Looks (and feels) great at my end, thanks @BenjaminCharmes!

PR #925, opened 2024-10-12
Ensure that new collection `immutable_id` is returned upon creation

ml-evs:
Previously, when creating a collection the `immutable_id` is returned as `None`, as we do not set the database-assigned ID back to the creation model. As collection relationships work from this ID, it is very helpful to return it, so that another request to `get_collection` does not need to be made to get the ID (see https://github.com/datalab-org/datalab-api/pull/35) for the required Python API workaround for this).

### DIFF
--- a/pydatalab/tests/server/test_samples.py
+++ b/pydatalab/tests/server/test_samples.py
@@ -485,12 +485,14 @@ def test_create_collections(client, default_collection):
     assert response.json["data"]["collection_id"] == "test_collection"
     assert response.json["data"]["title"] == "My Test Collection"
     assert response.json["data"]["num_items"] == 0
+    assert response.json["data"]["immutable_id"]
 
     response = client.get("/collections")
     assert response.status_code == 200
     assert len(response.json["data"]) == 1
     assert response.json["data"][0]["collection_id"] == "test_collection"
     assert response.json["data"][0]["title"] == "My Test Collection"
+    assert response.json["data"]["immutable_id"]
     assert response.status_code == 200
 
     # Create a collection with initial items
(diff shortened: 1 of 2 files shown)


pydatalab/tests/server/test_samples.py
```diff
@@ -485,12 +485,14 @@ def test_create_collections(client, default_collection):
...
     assert response.json["data"]["title"] == "My Test Collection"
     assert response.json["data"]["num_items"] == 0
+    assert response.json["data"]["immutable_id"]
 
     response = client.get("/collections")
     assert response.status_code == 200
     assert len(response.json["data"]) == 1
     assert response.json["data"][0]["collection_id"] == "test_collection"
     assert response.json["data"][0]["title"] == "My Test Collection"
+    assert response.json["data"]["immutable_id"]
```
ml-evs, line 495:
```suggestion
    assert response.json["data"][0]["immutable_id"]
```

pydatalab/tests/server/test_samples.py
```diff
@@ -485,12 +485,14 @@ def test_create_collections(client, default_collection):
...
     assert response.json["data"]["title"] == "My Test Collection"
     assert response.json["data"]["num_items"] == 0
+    assert response.json["data"]["immutable_id"]
 
     response = client.get("/collections")
     assert response.status_code == 200
     assert len(response.json["data"]) == 1
     assert response.json["data"][0]["collection_id"] == "test_collection"
     assert response.json["data"][0]["title"] == "My Test Collection"
+    assert response.json["data"][0]["immutable_id"]
```
ml-evs, line 495:
```suggestion
```

BenjaminCharmes (approved):
LGTM 👍

PR #931, opened 2024-10-15
Better handle case where block in database is not implemented by app/API

ml-evs:
Should remove the annoying warning when doing block development.

Also prevents reloading the block info endpoint for every edit page.


BenjaminCharmes (approved):
LGTM 👍 🚀

Issue #543, opened 2024-01-19
Make use of pymongo sessions for improved atomicity


ml-evs:
We have a few cases that use a database lookup to decide whether to make a modification, or a case where something is inserted in one step, then updated later.

As well as refactoring generally for improved atomicity, we should use pymongo sessions to ensure that the database is locked between these steps, as this can mess up backup integrity and breakages when running with multiple workers.

PR #933, opened 2024-10-15
Update collection deletion endpoint to also unlink items

### LINKED ISSUE #932: Items are not unlinked when underlying collection is deleted
The `DELETE /collections/<collection_id>` endpoint does not unlink items that were in that collection, and can instead accumulate a load of dead collection IDs.

This was obscured by the fact that the API never returns these dead IDs when returning item data, as it performs an aggregation to get live collections, though it would still be nice to tidy this up. I'm not sure it is worth the hassle of a full suggested migration, though it might be worth baking it in as a default database clean-up task in the future.

ml-evs:
Closes #932.

This also provides a template for #543, which we should really use everywhere throughout the API for routes that make multiple database calls


BenjaminCharmes (approved):
Works perfectly 👍

Issue #895, opened 2024-09-30
pre-commit uses development ESLint rules rather than prod


ml-evs:
As noticed by @BenjaminCharmes, the pre-commit ESLint seems to be using the dev rules rather than prod, so some fixable issues are not being fixed automatically. It seems quite hard to set this manually and none of the workarounds I tried seemed to work; instead we might revert the eslint config changes (i.e., treat most things as errors again) but figure out how to get `yarn serve` to be more forgiving itself.

PR #934, opened 2024-10-15
Fix eslint warning in CollectionInformation

BenjaminCharmes:
I have this warning every day on console, so I needed to fix it!


ml-evs:
Indeed! I was hoping to use this as an example to figure out why `pre-commit` won't fix it for us (see https://github.com/datalab-org/datalab/issues/895) -- though we can also just fix it and have a fake PR that tries to reintroduce it or similar errors.

ml-evs (approved):
For the sake of our collective sanity ;)

Issue #920, opened 2024-10-10
Report item schemas in API and use them to dynamically create UIs


ml-evs:
General idea: we want to dynamically report the item types that a datalab supports. These might be defined outside of this repository at the deployment level, e.g., a datalab instance defines a new custom field on top of our `Sample` type. The server should then report the JSONSchema for the `Sample` type at runtime on the server.

This will involve 3 main steps:

- [x] #940
- [ ] #952
- [ ] Then actually extend our pydantic models with ways to hint about desired UI. This ties in with custom types needing custom components, e.g., date time. [Add UI hints metadata system to Pydantic models #1401
]([https://github.com/datalab-org/datalab/pull/1401])
- [ ] Document all this!

Points that come to mind:

- determining/reporting type hierarchy
- versioning of schemas
- generally need to think about how we support custom schemas in datalab, of which this is a important part but not the whole story
- this approach, if successful, could also then be used for blocks

PR #941, opened 2024-10-17
Add API endpoint for reporting schemas

### LINKED ISSUE #940: Add an API endpoints for reporting schemas
Add an API /info/types & /info/types/sample endpoint for reporting schemas, starting with the JSON schema representation of the pydantic model model.schema_json() -- this could already be used by the Python API.

BenjaminCharmes:
Related to #920 
Closes #940

### DIFF
--- a/pydatalab/src/pydatalab/routes/v0_1/info.py
+++ b/pydatalab/src/pydatalab/routes/v0_1/info.py
@@ -11,7 +11,8 @@ from pydantic import AnyUrl, BaseModel, Field, validator
 from pydatalab import __version__
 from pydatalab.blocks import BLOCK_TYPES
 from pydatalab.config import CONFIG, FEATURE_FLAGS, FeatureFlags
-from pydatalab.models import Person
+from pydatalab.models import Collection, Person
+from pydatalab.models.items import Item
 from pydatalab.mongo import flask_mongo
 
 from ._version import __api_version__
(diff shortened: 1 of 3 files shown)


ml-evs:
Just marked this as ready for-review so I can remember to take a look tomorrow!

pydatalab/src/pydatalab/routes/v0_1/info.py
```diff
@@ -143,3 +144,36 @@ def list_block_types():
...
+
+
+def get_all_models():
+    return Item.__subclasses__()
+
+
+@INFO.route("/info/types", methods=["GET"])
+def list_supported_types_schemas():
+    """Returns a dictionary of supported item types and their schemas."""
+    schemas = {cls.__name__.lower(): cls.schema() for cls in get_all_models()}
```
ml-evs, line 156: Very minor, but I think I would use the item's `type` here as it should be the ground truth. This might need a bit of manipulation, but I think you can get it from e.g. `cls.schema()["properties"]["type"]["default"]` or something similar

ml-evs: also I think here I would actually just list the types themselves, rather than providing their full schema. Then each schema can be individually resolved under `/info/types/<x>`

pydatalab/src/pydatalab/routes/v0_1/info.py
```diff
@@ -143,3 +144,36 @@ def list_block_types():
...
+for model_class in get_all_models():
+    model_name = model_class.__name__.lower()
+
+    def make_route(model_class):
+        @INFO.route(
+            f"/info/types/{model_name}", methods=["GET"], endpoint=f"get_{model_name}_schema"
+        )
+        def get_model_schema():
+            """Returns the JSON schema for the model."""
+            return jsonify(model_class.schema())
```
ml-evs, line 171: It's just convention, but nesting this response under the keys `data->schema` would be good, then we can add any extra metadata or external links we want. We might want to introduce other fields for custom schemas, like `data->source`, or even just a simple list of properties (easier to access than within the JSONschema itself). It might be benefiical to grab the default metadata from the other info endpoint here.

webapp/src/components/SampleTable.vue
```diff
@@ -49,13 +50,24 @@ export default {
...
     this.getSamples();
+    await this.getSampleSchema();
   },
   methods: {
     getSamples() {
       getSampleList();
     },
+    async getSampleSchema() {
+      try {
+        this.sampleSchema = await getSchema("sample");
```
ml-evs, line 63:
```suggestion
        this.sampleSchema = await getSchema("samples");
```

Following my comment above this would then be pluralised

ml-evs (changes requested):
Hi @BenjaminCharmes, some initial minor comments -- I don't think these change much how this gets used in the UI, beyond tweaking the endpoint names. Any other optional fields I mentioned can be left out for now, but it's probably good to standardize on the `model.type` rather than the Python class name.

As discussed, let's try to get this merged and then I'll try stick a custom schema into this process and see what breaks

pydatalab/src/pydatalab/routes/v0_1/info.py
```diff
@@ -143,3 +144,50 @@ def list_block_types():
...
+    return Item.__subclasses__()
+
+
+@INFO.route("/info/types", methods=["GET"])
+def list_supported_types():
+    """Returns a list of supported schemas."""
+    types = [cls.schema()["properties"]["type"]["default"] for cls in get_all_items_models()]
+    types.append(Collection.schema()["properties"]["type"]["default"])
+
+    return jsonify(types)
```
ml-evs, line 159:
```suggestion
    return jsonify(types)
```

This should also be encapsulated in a JSON:API response under a `data` key.

ml-evs (commented):
Couple of minor comments that I will address myself before merging, will also quickly add a test. One thing we'll want to do is cache the schemas on startup, rather than regenerate them on the route each time

PR #942, opened 2024-10-17
Hotfix for file-overwriting from uppy

### LINKED ISSUE #939: `stored_files` name clash leads to leakage between samples
I've just run into quite a nasty issue with the `stored_files` field in our store. When you go to a sample that has a file called `x.csv` uploaded to it, then visit another sample and try to upload a file called `x.csv`, the stored files cache is not cleared properly and it treats this as a duplicate to be overwritten. Most worryingly, if you accept that the file should be replaced, it uploads it to the *original* sample, i.e., the first one you clicked on.

Proposed fix:
- properly clear the store when loading new files
- stop matching files on filenames anyway, since many instruments produce files with the same name for different experiments. The API shouldn't care that you have two files with the same name either way, so neither should the UI. 

This might be related to the #636 and #500 (which I actually thought we had fixed...).

I'll try to take a look at this fairly urgently, although it is an unlikely enough scenario to come up in "production".

ml-evs:
Do not consider filenames when uploading a file via uppy, just show the alert

Closes #939

### DIFF
--- a/webapp/src/file_upload.js
+++ b/webapp/src/file_upload.js
@@ -57,9 +57,8 @@ export default function setupUppy(item_id, trigger_selector, reactive_file_list)
     for (const file_id in reactive_file_list) {
       // console.log(`evaluating ${reactive_file_list[file_id].name} vs ${file.name}`)
       if (reactive_file_list[file_id].name == file.name) {
-        matching_file_id = file_id;
         alert(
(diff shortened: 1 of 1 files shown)


ml-evs:
Just a note before merging that the original behaviour (but working) should be reintroduced in #943.

PR #943, opened 2024-10-17
Rework of the files utilities and endpoints for both app and API

ml-evs:
This PR reworks some of the file upload handling in the app and API, particularly how file metadata is used in the vuex store, and where the file update methods get called. The API remains unchanged beyond some minor refactoring for clarity (and extra defensiveness on permissions).


ml-evs:
I'm happy that this works well with @BenjaminCharmes's additions, but we currently don't have tests for e.g, removing a file from a sample via the app, so perhaps this could be added at some point.

PR #945, opened 2024-10-20
Add scheduled `uv lock -U` to replace dependabot

ml-evs:
dependabot is not being very useful right now. Instead, I propose a simple wf that runs `uv lock -U` each week and makes a PR with any changes, similar to a grouped dependabot update PR that should always lead to latest compatible versions.

We could then adapt this in the future with updates for newer (potentially breaking) versions of other deps.

If this works, we can disable dependabot for Python for all but security updates (which I don't believe `uv` has explicit support for), and leave it on only for `npm`.

### DIFF
--- a/.github/workflows/dependable-bot.yml
+++ b/.github/workflows/dependable-bot.yml
@@ -1,0 +1,45 @@
+name: Automatic `uv` dependency upgrades
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


ml-evs:
Merging just to test it out

PR #950, opened 2024-10-20
Add `uv-lock` to pre-commit

ml-evs:
This will be helpful for e.g., dependabot PRs that update `pyproject.toml`, but don't know how to update the `uv.lock`, as we can just ask pre-commit CI to fix them, with the CI still running with locked deps every time.


ml-evs:
Nevermind, this is fraught with issues:

- uv needs to install packages which are not allowed at runtime in free pre-commit.ci
- uv in pre-commit ends up nuking another installed local version's cache

PR #953, opened 2024-10-21
Use schemas to dynamically display UI

### LINKED ISSUE #952: Use this information in the UI: e.g., when generating dynamic data table for a type, or when generating item pages. At least initially, this could just give a way to list custom fields that are missing from the edit page with generic components. e.g., if JSON schema type is a string: show it in a textarea, or if its a date, show it as a date. i.e., type-aware components. This might involve extending JSON schema types with some internal type-system that includes e.g., dates.

BenjaminCharmes:
Closes #952 

First draft (not finished yet) maybe it's not what you had in mind.
To discuss !


ml-evs (commented):
Looking good so far -- as discussed, please update this PR with the final changes from #941 

Then:

- let's add this as a `/next` edit page alongside the current one
- see how much hardcoding you can remove at this point, but don't worry too much, e.g., this edit page should not use the `SampleInformation` style component, but instead should have some vague sections that you can put your fields into
- once this is working, we can try it with entirely custom schemas and figure out what kind of hints we need to put in the schemas themselves around sizing/ordering/styling of fields as components.

PR #957, opened 2024-10-22
Remove `vue/cli` as dependency and add missing polyfills

ml-evs:
`vue/cli` is not needed after the initial project scaffold, but as it is in maintenance mode, it leads to many dependency security issues for us. This PR removes it entirely from the repo, along with all of the extra deps it brings.

Note, this is different from the `vue/cli-service`, which is also deprecated, but simply provides some wrapper scripts for e.g., cypress, eslint, build, serve etc.

Because Javascript, removing vue/cli also breaks a load of webpack 5 polyfills, because vue/cli was already polyfilling the update from webpack 4 to 5 for us magically. So I've had to add those back in explicitly, with `process` and `buffer`


BenjaminCharmes (approved):
Everything works fine for me !

PR #970, opened 2024-10-26
Add timezone consistency across deployments & UI

### LINKED ISSUE #440: Server `last_modified` times depend on server timezone
Even though #439 resolves #329, the actual timezone used by the server determines the `last_modified` timestamp (which is UTC on the deployment server anyway). This should also use UTC.

ml-evs:
This PR enforces UTC for all datetimes stored by datalab, via:

- updating the underlying model so that any missing timezone from a user/script is interpreted as UTC. This is done at the validator level, so some inconsistencies might arise in old data that has not previously been stored with a timezone, but the API at least will always validate and fix it this now. I explicitly added the timezones into remote fs checking so that we don't have to worry about mongo version compatibility.
- The app code can now assume that a missing timezone really does mean UTC, and all future timestamps should have the `00:00` UTC offset suffix.
- App code now has to have a custom computed setter/getter for datetime objects, so that we can cast the format from HTML `<input format='datetime-local'>` into the full isoformat stamp.
- all internal datetimes now use UTC explicitly, with enabling of the ruff DTZ/flake8-datetimez linting rules

Closes #440 

Assorted fixes:

- [x] the `IsoformatDateTime` wrapper was also setting a bad JSON Schema type, for some reason -- this is removed
- [x] The `File` model wasn't using the extended datetime format for `time_added`. 

Relatedly (from HN today...): https://ssoready.com/blog/engineering/truths-programmers-timezones/

### DIFF
--- a/webapp/src/views/CollectionPage.vue
+++ b/webapp/src/views/CollectionPage.vue
@@ -144,9 +144,7 @@ export default {
       if (item_date == null) {
         this.lastModified = "Unknown";
       } else {
-        // API dates are in UTC but missing Z suffix
-        const save_date = new Date(item_date + "Z");
-        this.lastModified = formatDistanceToNow(save_date, { addSuffix: true });
+        this.lastModified = formatDistanceToNow(new Date(item_date), { addSuffix: true });
       }
     },
   },
--- a/webapp/src/views/EditPage.vue
+++ b/webapp/src/views/EditPage.vue
@@ -310,9 +310,7 @@ export default {
       if (item_date == null) {
         this.lastModified = "Unknown";
       } else {
-        // API dates are in UTC but missing Z suffix
-        const save_date = new Date(item_date + "Z");
-        this.lastModified = formatDistanceToNow(save_date, { addSuffix: true });
+        this.lastModified = formatDistanceToNow(new Date(item_date), { addSuffix: true });
       }
     },
   },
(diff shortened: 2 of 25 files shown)


webapp/src/views/CollectionPage.vue
```diff
@@ -144,9 +144,7 @@ export default {
       if (item_date == null) {
         this.lastModified = "Unknown";
       } else {
-        // API dates are in UTC but missing Z suffix
-        const save_date = new Date(item_date + "Z");
-        this.lastModified = formatDistanceToNow(save_date, { addSuffix: true });
+        this.lastModified = formatDistanceToNow(item_date, { addSuffix: true });
```
ml-evs, line 147:
```suggestion
        this.lastModified = formatDistanceToNow(new Date(item_date), { addSuffix: true });
```

webapp/src/views/EditPage.vue
```diff
@@ -310,9 +310,7 @@ export default {
       if (item_date == null) {
         this.lastModified = "Unknown";
       } else {
-        // API dates are in UTC but missing Z suffix
-        const save_date = new Date(item_date + "Z");
-        this.lastModified = formatDistanceToNow(save_date, { addSuffix: true });
+        this.lastModified = formatDistanceToNow(item_date, { addSuffix: true });
```
ml-evs, line 313:
```suggestion
        this.lastModified = formatDistanceToNow(new Date(item_date), { addSuffix: true });
```

ml-evs (commented):
This should fix the issue you were seeing @BenjaminCharmes, was just some overzealous refactoring

PR #974, opened 2024-10-29
Enhanced filtering UI on fields with components

### LINKED ISSUE #973: Enhanced filtering UI on fields with components
Enhanced filtering UI on fields with components, like Creators and Collections

BenjaminCharmes:
Related to #889
Closes #973 

Enhanced filtering for Creators and Collections column

### DIFF
--- a/webapp/src/components/CollectionTable.vue
+++ b/webapp/src/components/CollectionTable.vue
@@ -19,7 +19,7 @@ export default {
       collectionColumn: [
         { field: "collection_id", header: "ID", body: "FormattedCollectionName", filter: true },
         { field: "title", header: "Title" },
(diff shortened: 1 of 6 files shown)


jdbocarsly:
This looks nice, a few comments/questions:
1. Can we show the names in the table by default (perhaps just hiding them if there is more than one creator)?
2. What do the "starts with", "ends with", etc. filters do if the options are a checkbox, rather than a free text?

ml-evs (commented):
- let's think about making type a categorical filter too, after pulling your API schemas

ml-evs (approved):
From what I've seen, this works really nicely now! I'll merge at this point and we can circle back to the other comments later.

PR #977, opened 2024-10-31
Stagger fortnightly scheduled uv/yarn updates

ml-evs:
I think it makes sense to stagger the Python/JS ecosystem updates, so if we have any issues in the week after merging one, we can be fairly sure it was the cause. This PR staggers the updates so that Python updates on Wednesdays in the 1st and 3rd week of a month, and JS in the 2nd and 4th.


BenjaminCharmes (approved):
LGTM 👍

PR #979, opened 2024-11-01
Fix default `MAX_CONTENT_LENGTH` setting to 10 GB

### LINKED ISSUE #978: `MAX_CONTENT_SIZE` config is not using the correct units
The default value is meant to be 100 GB (perhaps this should be revised anyway...) but it seems that the config value is being interpreted as much lower (presumably 100 MB). This should be:

- tested, by making sure you cannot upload a file when it is set too low (i.e., in the test suite, set it to 5 MB or something and try to upload our largest file)
- the default value should be changed to ~10 GB

https://github.com/datalab-org/datalab/blob/b38c99f404acff7622653b2e873070ba324e3fdd/pydatalab/src/pydatalab/config.py#L233-L240

ml-evs:
Closes #978

### DIFF
--- a/pydatalab/tests/server/conftest.py
+++ b/pydatalab/tests/server/conftest.py
@@ -82,6 +82,9 @@ def app_config(tmp_path_factory):
         "EMAIL_DOMAIN_ALLOW_LIST": ["example.org", "ml-evs.science"],
         "MAIL_DEBUG": True,
         "MAIL_SUPPRESS_SEND": True,
+        # Set to 10 MB to check that larger files fail; this should be larger than all of our example files.
+        # Elsewhere, we can generate an artificial large file to check that it fails.
+        "MAX_CONTENT_LENGTH": 10 * 1000**2,
     }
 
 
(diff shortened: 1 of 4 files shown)


BenjaminCharmes (approved):
LGTM 👍

pydatalab/tests/server/conftest.py
```diff
@@ -82,6 +82,9 @@ def app_config(tmp_path_factory):
         "EMAIL_DOMAIN_ALLOW_LIST": ["example.org", "ml-evs.science"],
         "MAIL_DEBUG": True,
         "MAIL_SUPPRESS_SEND": True,
+        # Set to 10 MB to check that larger files fail; this should be larger than all of our example files.
+        # Elsewhere, we can generate an artificial large file to check that it fails.
+        "MAX_CONTENT_LENGTH": 10 * 1024**2,
```
ml-evs, line 87:
```suggestion
        "MAX_CONTENT_LENGTH": 10 * 1000**2,
```

PR #981, opened 2024-11-01
Fix and streamline attribute lookup in MediaBlock

### LINKED ISSUE #980: Media block no longer works after store changes
Looks like https://github.com/datalab-org/datalab/pull/943 broke the media block. I'll make a PR that adds a fix and hopefully tests it with cypress now that we can check blocks.

ml-evs:
Closes #980

### DIFF
--- a/webapp/cypress.config.ts
+++ b/webapp/cypress.config.ts
@@ -5,8 +5,6 @@ export default defineConfig({
   e2e: {
     baseUrl: "http://localhost:8080",
     apiUrl: "http://localhost:5001",
-    experimentalMemoryManagement: true,
-    numTestsKeptInMemory: 0,
     defaultCommandTimeout: 10000,
   },
   component: {
--- a/webapp/cypress/e2e/editPage.cy.js
+++ b/webapp/cypress/e2e/editPage.cy.js
@@ -275,4 +275,21 @@ describe("Edit Page", () => {
     cy.contains("label", "X axis").should("exist");
     cy.contains("label", "Y axis").should("exist");
   });
+
+  it("Uploads a fake PNG image, make a Media block and checks that the image is shown", () => {
+    let test_fname = "test_image.png";
+    cy.createTestPNG(test_fname);
+    cy.uploadFileViaAPI("editable_sample", test_fname);
+
+    cy.get('[data-testid="search-input"]').type("editable_sample");
+    cy.findByText("editable_sample").click();
+
+    cy.findByText("Add a block").click();
+    cy.get(".dropdown-menu").findByText("Media").click();
+    cy.findAllByText("Select a file:").eq(1).should("exist");
+    cy.get("select.file-select-dropdown").eq(1).select(test_fname);
+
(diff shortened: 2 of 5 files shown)


webapp/cypress.config.ts
```diff
@@ -5,8 +5,7 @@ export default defineConfig({
   e2e: {
     baseUrl: "http://localhost:8080",
     apiUrl: "http://localhost:5001",
-    experimentalMemoryManagement: true,
-    numTestsKeptInMemory: 0,
+    numTestsKeptInMemory: 1,
```
ml-evs, line 8: Actually the default is 50 here, so I'm just going to remove our config. It's very useful to have snapshots locally

docker-compose.yml
```diff
@@ -51,6 +51,8 @@ services:
     restart: unless-stopped
     networks:
       - backend
+    ports:
+      - "27017:27017"
```
ml-evs, line 55:
```suggestion
```

webapp/cypress/e2e/editPage.cy.js
```diff
@@ -275,4 +275,21 @@ describe("Edit Page", () => {
...
+    cy.get('[data-testid="search-input"]').type("editable_sample");
+    cy.findByText("editable_sample").click();
+
+    cy.findByText("Add a block").click();
+    cy.get(".dropdown-menu").findByText("Media").click();
+    cy.findAllByText("Select a file:").eq(1).should("exist");
+    cy.get("select.file-select-dropdown").eq(1).select(test_fname);
+
+    // Check that the img with id "media-block-img" is present
+    cy.get("img#media-block-img").should("exist");
```
jdbocarsly, line 293:
```suggestion
    cy.get("img#[data-testid=media-block-img]").should("have.length", 1);
```

See comment below about id. Also would be good in this case I think to check that it has length 1, since the `.get` would pick up multiple nodes with that same `data-testid` if multiple media blocks were on the page. We had something similar happen a little while ago when all the tests broke

webapp/src/components/datablocks/MediaBlock.vue
```diff
@@ -8,7 +8,7 @@
       class="mb-3"
       update-block-on-change
     />
-    <img v-if="isPhoto" :src="media_url" class="img-fluid mx-auto" />
+    <img v-if="isPhoto" id="media-block-img" :src="media_url" class="img-fluid mx-auto" />
```
jdbocarsly, line 11:
```suggestion
    <img v-if="isPhoto" data-testid="media-block-img" :src="media_url" class="img-fluid mx-auto" />
```


Assigning an `id` should be avoided here since `id`s are supposed to be unique across the whole DOM. When two (or more) Media Blocks are added to an Edit page, the images would have the same `id` and that could cause problems. Probably best thing to do here is use a testing-specific attibute, `data-testid="media-block-img"`

jdbocarsly (changes requested):
small change on the testing, but works great for me! Thanks

PR #984, opened 2024-11-04
LoginDetails alignment when a user is unverified

BenjaminCharmes:
Fix the alignment of LoginDetails when an unverified user is logged in


ml-evs (changes requested):
We should use data-testid here, as suggested

ml-evs (approved):
Works well, thanks!

PR #988, opened 2024-11-05
Naively update eslint and plugins, remove some more vue-cli-plugins

ml-evs:


ml-evs:
Putting this on hold as this is non-trivial for stupid JS ecosystem reasons. The eslint config file format has completely changed and they haven't released a migration script that works for us yet. I'm tempted to bin all of our JS linting and start again with vite or likewise at some point.

PR #990, opened 2024-11-05
Chatblock deployment fix: make sure LLM API keys are passed from .env to environ

ml-evs:
Previously, LLM API keys were automatically set as env vars by pipenv as part of the build process. In this PR, the procedure is passed onto flask itself.


ml-evs:
pre-commit.ci autofix

PR #993, opened 2024-11-06
Initial implementation of manual ngram-based search in MongoDB

### LINKED ISSUE #679: Free text search could be improved
Due to limitations of our full text index, you often have to type the full item ID before results appear for your search. We should try to set up an n-gram based index instead.

cc @elbee99

ml-evs:
Closes #679 -- MongoDB text indexes tokenize using whitespace and punctuation only. This PR investigates whether we can build a manual ngram index, so when searching for refcode `ABCDEF`, you get results if you only ask for `ABC`, `BCD` etc. 

This is done by making a separate collection called `item_fts` that is used only for this kind of search, by storing `immutable_id`, `type` and `ngrams` for all items, with an index over `ngrams`. Lookup is then done by ngrammifying the query string and doing array lookup and ordering by the number of matches.

Will have to fiddle around to see:
- what the optimal value of `N` is, and whether we need to do all N+1-grams up to a fixed range
- how expensive this is for realistic deployment sizes
- whether it might be better to try an edit-distance based approach for some fields

### DIFF
--- a/pydatalab/src/pydatalab/main.py
+++ b/pydatalab/src/pydatalab/main.py
@@ -206,6 +206,7 @@ def create_app(
         extension.init_app(app)
 
     pydatalab.mongo.create_default_indices()
+    pydatalab.mongo.create_ngram_item_index()
 
     if CONFIG.FILE_DIRECTORY is not None:
         pathlib.Path(CONFIG.FILE_DIRECTORY).mkdir(parents=False, exist_ok=True)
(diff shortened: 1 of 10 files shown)


ml-evs:
I think this is ready for review now, though we should not make it the default (yet) until we can test scaling and quality of search results.

pydatalab/src/pydatalab/main.py
```diff
@@ -206,6 +206,7 @@ def create_app(
         extension.init_app(app)
 
     pydatalab.mongo.create_default_indices()
+    pydatalab.mongo.create_ngram_item_index()
```
ml-evs, line 209: This should only be run on one of the API processes, or there should be a lock

PR #1001, opened 2024-11-12
Add docker build var to spoof API version

ml-evs:
Right now, if the version of datalab is being installed from a git submodule, or a worktree, it is impossible to set the API version. For the app, we can override  this with the `VUE_APP_GIT_VERSION`. For the API, we have to use the `SETUPTOOLS_SCM_PRETEND_SCM_VERSION`. This PR passes that value through the build system to make it available when `uv` installs.


ml-evs:
I've tested this on yellowhammer, so will merge when the final tests pass again

PR #1004, opened 2024-11-13
Fix vue-loader errors

BenjaminCharmes:
Add missing `<tbody>` or `<tr>` inside `components/BatchCreateItemModal.vue`, `components/StatisticsTable.vue` & `views/About.vue`


jdbocarsly (approved):
Looks great! Thanks for doing this

PR #1006, opened 2024-11-14
Tweak git version spoofing in docker config, and test it 

ml-evs:
- Remove annoying warning by removing default value for spoof version in docker-compose
- Add version spoof to CI to check that it does not introduce build failure


ml-evs:
This is very minor, don't think it kills our docker caching in CI (though impossible to say -- no performance decrease anyway) -- will merge

PR #1010, opened 2024-11-19
Allow both search and filter to be specified simultaneously in entry table

### LINKED ISSUE #992: Item table search bar does nothing when filter is applied
@BenjaminCharmes and I noticed that the search bar above the table doesn't work if a column filter is already applied. This should be fixable as both are part of prime, but if not, then I am not against just disabling the search bar after a filter.

BenjaminCharmes:
Closes #992


ml-evs (approved):

The CI failures are just annoying red herrings from cypress, as the tests are triggered again when you mark a PR as ready (required for our auto-deps update PRs) -- cypress cloud decides not to return a positive status as it thinks the tests have already run.

PR #1011, opened 2024-11-19
Dynamic table tweaks: filter by item type enum & resizable columns

ml-evs:
This PR attempts to turn on resizable columns and set the starting size with `"fit"`, addressing #889,

It also then tries to use a custom enum filtering for item types. For now, I just pull the values that the user has in the table, but ideally in the future it would query the server for the API schemas to see which types are available globally.

This doesn't quite work yet, the filtering does not seem to be applied correctly for types. Could oyu maybe take a look @BenjaminCharmes?

### DIFF
--- a/webapp/src/components/DynamicDataTable.vue
+++ b/webapp/src/components/DynamicDataTable.vue
@@ -17,6 +17,8 @@
       filter-display="menu"
       :global-filter-fields="globalFilterFields"
       removable-sort
+      column-resize-mode="fit"
+      resizable-columns
       sort-mode="multiple"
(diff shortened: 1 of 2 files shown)


ml-evs:
Thanks @BenjaminCharmes, your change works nicely! I have reintroduced the generic filter (that was being used by ID too).

If its not too fiddly, it would be good to also disable `MATCH ALL` as It doesn't really make sense for types.

ml-evs:
Thanks for the help @BenjaminCharmes, this works nicely for me now!

PR #1013, opened 2024-11-22
Improve DataTable Pagination Handling with Vuex

### LINKED ISSUE #983: Items per page selector resets on viewing sample
In the web ui, the "Items per page" selector resets to 20 regardless of the previous choice after viewing a sample.

BenjaminCharmes:
Closes #983

Store the current page and the number of rows of each page in the vuex store

### DIFF
--- a/webapp/cypress/component/CollectionTableTest.cy.jsx
+++ b/webapp/cypress/component/CollectionTableTest.cy.jsx
@@ -9,6 +9,12 @@ describe("CollectionTable Component Tests", () => {
     store = createStore({
       state() {
         return {
+          datatablePaginationSettings: {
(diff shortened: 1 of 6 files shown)


ml-evs (approved):
Works really well for me! Resets on refresh but persists page number/size when pressing "back" from a sample.

For now, when you click "home" after a sample, I guess that counts as a refresh and the store is cleared. I think resetting on refresh is fine, but I wonder if we can retain the store when someone clicks home? I've actually just noticed why this is happening (we don't use `router-link` for the edit page navbar) -- I'll make a follow-up PR.

ml-evs (approved):

PR #1026, opened 2024-12-02
Switch from vue-cli to vite

### LINKED ISSUE #1005: Switch our webapp build tool from vue-cli to vite
As `vue-cli` is in maintenance mode, we should upgrade to `vite`.

BenjaminCharmes:
Closes #1005

First attempt to switch from vue-cli to vite and a few things to discuss:

- I had to use `cryptojs` instead of `crypto` (`MD5(value).toString();` instead of `crypto.createHash("md5").update(value).digest("hex");`)

- I had to use `import.meta.env.BASE_URL` instead of `process.env.BASE_URL`

- I remove `.browserslistrc` and `babel.config.js`, not useful with vite ?

- `vite 6.x.x` not supported by `cypress` ? (warning message)

- In `vite.config.js`, I use `extensions: [".vue", ".js", ".json"]` to not rename all the imports in all the files, but it could be better to take the time to do it ? (e.g. `import Navbar from "@/components/Navbar";` to `import Navbar from "@/components/Navbar.vue";`

- Had to rename env variables from `VUE_` to `VITE_` to make it work without having to import all of them. Should we keep it that way (and also rename them where ever it's needed, e.g. github?) or just import them in the vite.config.js ?

To do: 

- [ ] Rewrite the scripts in `package.json` ?
- [x] Need to fix `Bokeh`.
Just need to downgrade to Bokeh 2.4.0 
- [x] `import "tinymce/plugins/hr";` not working
Downgrade tinymce to v5
- [x] Fix docker -> Fix cypress e2e test in github workflow


ml-evs (commented):
Comments for myself:

- [ ] check docker build version info with vite
- [ ] think about migrating env vars too

CHANGELOG.md
ml-evs, line 1: Were these changes done automatically by vite? Feels weird that it would break out of the directory it should be in

webapp/cypress/component/UserBubbleLoginTest.cy.jsx
```diff
@@ -1,10 +1,10 @@
 import { createStore } from "vuex";
 import UserBubbleLogin from "@/components/UserBubbleLogin.vue";
 import NotificationDot from "@/components/NotificationDot.vue";
-import crypto from "crypto";
+import MD5 from "crypto-js/md5";
```
ml-evs, line 4: Think we need to be careful here too, I had a lot of fun getting the crypto-browserify stuff to work recently. The crypto-js library itself is deprecated (see README here: https://www.npmjs.com/package/crypto-js) -- we may need another solution

webapp/index.html
```diff
@@ -1,11 +1,14 @@
...
+<!DOCTYPE html>
 <html lang="">
   <head>
     <meta charset="utf-8" />
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <meta name="viewport" content="width=device-width,initial-scale=1.0" />
-    <link rel="icon" href="<%= BASE_URL %>favicon.ico" />
-    <title><%= htmlWebpackPlugin.options.title %></title>
+    <link rel="icon" href="/favicon.ico" />
+    <title>Default Title</title>
```
ml-evs, line 8:
```suggestion
    <title>datalab</title>
```

May as well just use datalab here too, or can this be removed entirely?

webapp/package.json
```diff
@@ -3,76 +3,59 @@
...
     "primeicons": "^7.0.0",
-    "primevue": "^4.0.0",
+    "primevue": "^4.2.4",
     "process": "^0.11.10",
-    "qrcode-vue3": "^1.6.8",
-    "serve": "^14.2.1",
+    "qrcode-vue3": "^1.7.1",
+    "redis": "^4.7.0",
+    "serve": "^14.2.4",
     "stream-browserify": "^3.0.0",
```
ml-evs, line 45: We may not need stream-browserify anymore

webapp/vue.config.js
```diff
@@ -1,43 +0,0 @@
...
-        process: "process/browser",
-      }),
-    ];
-  },
-  chainWebpack: (config) => {
-    config.plugin("html").tap((args) => {
-      args[0].title = process.env.VUE_APP_WEBSITE_TITLE || "datalab";
-      args[0].meta = {
-        x_datalab_api_url: process.env.VUE_APP_API_URL,
-      };
```
ml-evs, line 39: This is used by the client code quite a lot, we should make sure that the equivalent is done with vite too (I couldn't see it above), should be doable in the same way as the website title

ml-evs (changes requested):
Thanks @BenjaminCharmes, great to remove 4000 lines from our lockfile, though we might need to review some of the deps upgrades carefully to make sure they still work. Some initial comments below

ml-evs (commented):
Two things left:

- fix whatever is going on with `cypress/support/e2e.js` (maybe just try removing it)
- at some point, update our package.json so that it launches the dev server for you before running tests, as vue-cli did

ml-evs:
I've pushed a bit further on this in the last few commits, though its really unsatisfactory still. The component tests rae failing with `Cannot access [Component] before initialization`. With the help of Claude, I've tried to address this with two changes:

- Added more node.js polyfills for Vite
   - Configured `vite-plugin-node-polyfills` in both Vite and Cypress configs to support the `crypto` module in browser environments, since the UserBubble test was failing at least partially for this reason
- Resolved a potential circular dependency in resources.js:
   - Moved/refactored `GRAVATAR_STYLE` (which has never been changed) from resources.js to prevent initialization order issues
   - The bottom line is that Vite's ESM-based module system is stricter about circular dependencies than vue-cli was, and component tests are particularly sensitive to these initialization order issues

Of the remaining failures, the `FormattedItemName` test fails for the same reason here, that `resources.js` is imported. Then all of the dynamic table tests fail, presumably for the same reason... We might need to think about refactoring `resources.js` generally.

ml-evs:
See e.g. 337b718c for the fix required to `FormattedItemName` to allow this component test to pass. I don't think we necessarily want to do this change...

ml-evs:
I've unpicked this a bit in a new PR at #1873. Thankfully things we changed around gravatar, crypto etc. have made this migration a bit easier (and thankfully Claude is now smarter!)

PR #1028, opened 2024-12-03
Upgrade to Bootstrap v5

### LINKED ISSUE #32: Update to Bootstrap 5
The webapp is currently using Bootstrap 4 without a JS interaction library, since Bootstrap 4 JS relies on Jquery which we do not want to use in the app. BootstrapVue would solve this, but it does not support Vue3. 

Upgrading to Bootstrap 5 will solve this issue, as it uses pure JS for its interactions. We should also be able to choose which js to include (if any).

### LINKED ISSUE #620: Better mobile support
Again, from discussions with @veronika-sedajova! I tried out datalab on my phone with some pretty big echem data. It actually performs remarkably well (comparable speed to the desktop browser given how much is done server-side) but some of the UI gets pretty messed up in narrow screens (e.g., the top bar on the edit page).

Testing this would be a whole can of worms but we might be able to get 80% of the way there by just adjusting bootstrap classes to allow people to view things on their phones.

BenjaminCharmes:
Closes #32:

- New `<button />` class used to close `Modal`
- Replace `ml-2` (doesn't exist in Bootstrap v5) by `gap-2` (d-flex) for `DynamicDataTableButtons` (could also use `me-2` from bootstrap instead) (Bootstrap left and right replaced by start and end)
- I had `color: black` to FormattedItemName because `badge light` display text in white now
- Edit every Modal (Item, Batch, Collection, Equipment ...) to stay the same
- Edit every datablock (and also Bokeh block with querySelector)
- and more ...

Closes #620:

- CompactConstituentTable.vue is now responsive.
- The EditPage and CollectionPage navbar now use a dropdown on small screen
- DynamicDataTabeButtons also use a dropdown on small screen
- and more ...

Cypress:
- Improving tests using Modal and vue-select
- Add a few utility functions
- Lot of refactoring
- There are still a lot of `cy.wait()` that could (should?) be removed (I haven't removed them, waiting feedback).


webapp/package.json
```diff
@@ -23,7 +23,8 @@
     "@uppy/webcam": "^1.8.4",
     "@uppy/xhr-upload": "^1.6.10",
     "@vueuse/components": "^10.7.2",
-    "bootstrap": "^4.5.3",
+    "boostrap": "^2.0.0",
```
ml-evs, line 26: :eye: think this one needs to be removed! Thankfully this is one where npm have sat on the name to prevent malicious use

ml-evs (changes requested):
I've built and played around with the new version and only have a few minor comments of mine (and we can go through yours) -- just a comment below about the deps!

BenjaminCharmes:
Everything works! 🎉 I still need to refactor the e2e tests a bit (I added a lot of functions) and clean things up, and then the PR will be ready.

ml-evs:
I'm going to collect general comments as I explore the UI without trying to be mega picky, and without referencing any particular changes. These may also be issues present in the existing UI. I'll tag them as :red_circle: (needs fixing), :yellow_circle: (needs discussion) and :green_circle: (probably fine or an existing issue, just noting the change).

- [ ] :green_circle:  `button-left` on the DynamicDataTable seems to fill to a different height to `button-right` (38px vs 42px for me). There is now uniformity between e.g., the search bar and the multi-select, which wasn't the case before (i.e., a good thing!) 
- [ ] :yellow_circle: / :green_circle: The `button-left` and `button-right` groups still do not scale very nicely in small windows 
- [ ] :green_circle: item badges have slightly more padding now (not necessarily bad but might need to be reduced slightly as it decreases the font size)
- [ ] :red_circle: submit button in modal has new (worse, IMO) colour -- black on rich blue rather than white on a darkish teal
- [ ] :red_circle: Router links (e.g., About | Samples | Collections) are now underlined even when they are not selected
- [ ] :red_circle: / :yellow_circle: Bokeh plots in some blocks like XRD are now much wider by default (maybe too large), though previously I think this was hardcoded on the Python side which is also not optimal
- [ ] :yellow_circle: The margin for the block names in the add a block menu is quite a lot larger
- [ ] :red_circle: Synthesis information tables now reactive (great!) but the unit column is now too large and it gets broken by samples with long names as it no longer wraps (see screenshot) -- this actually breaks all the rows above it too.
    2025-05-16-155345_1367x250_scrot

BenjaminCharmes:
I've made most of the changes asked, and awaiting discussion on the others.

ml-evs (approved):
Great work @BenjaminCharmes, this is looking really nice from my side! I'm going to approve this now, but we should hold off merging it until the 0.5.3 release is out. Then this can be the first PR merged for the 0.6.0 series.

PR #1054, opened 2025-02-21
Add overlay while deleting multiple items

### LINKED ISSUE #944: Deleting multiple items should have accompanying loading state
Right now, if you select (say) 100 items and delete them, you sit and watch as they trickle away from the sample table one at a time.

We should add a loading state that a) prevents the user from clicking/interrupting the process, and b) prevents the table from refreshing until all API calls have been made.

BenjaminCharmes:
Closes #944

Add an overlay while deleting multiple items to prevents the user from interrupting the process.


be-smith (approved):
lgtm. Tested locally works well :)

PR #1055, opened 2025-02-25
Do not update 'files' reference when saving item

ml-evs:
I accidentally depended on an item having `files` saved in the db, where actually it should only be a field constructed for the API response. This crept through as calling `save-item` from the app included this field, and as it fits the pydantic model, it was being saved in the database.

We still need a better way of distinguishing such fields between the database and API, by adding special validators to the base entry types and annotating them as such in our models, but this just patches this problem for now.


ml-evs:
Actually, converting this to a draft to add proper tests and a more general solution at a later date.

PR #1059, opened 2025-03-04
Add concept of block events and some design notes

ml-evs:
This PR adds the concept of "events" that can be registered on block types; essentially, server-side functions that can be triggered separately from parsing/ingestion for things like setting parameters. It also introduces the handling of particular events raised by Bokeh plots in the base data block in Vue.

This is all described in a new design document for this current version of datablocks, also committed in this PR.

I think eventually we can add several lifecycle hooks to the base block for handling certain things (remake block from scratch etc) that can be reused by derived block types, though I haven't put any effort into doing this now. I'm already using this functionality in a custom block that sets some parameters based on user feedback that requires reanalysing all block data (essentially a slider setting some offset time parameter).

For review, I recommend reading the design doc in the rtd preview at https://the-datalab--1059.org.readthedocs.build/en/1059/design/blocks/.

There's still a few open questions about how errors in events should be handled, and what methods should be triggered afterwards (mostly due to the fact we always call the `to_web()` block method in the update-block route.


ml-evs:
Just pinging @BenjaminCharmes and @jdbocarsly for review on this one!

BenjaminCharmes (approved):
LGTM 🚀 very promising!
XRD Block works perfectly fine.

pydatalab/src/pydatalab/apps/xrd/blocks.py
```diff
@@ -27,9 +27,19 @@ class XRDBlock(DataBlock):
     def plot_functions(self):
         return (self.generate_xrd_plot,)
 
-    @classmethod
+    @event
+    def set_wavelength(self, wavelength: float | None):
```
jdbocarsly, line 31: Can you show how this gets implemented in the XRD block before we merge? That would be a good way to test that everything is working on the front end

webapp/src/components/datablocks/DataBlockBase.vue
```diff
@@ -207,6 +207,10 @@ export default {
         this.contentMaxHeight = "none";
       }
     });
+    document.addEventListener("bokehStateUpdate", this.handleBokehEvent);
```
jdbocarsly, line 210: If this adds a listener to the overall document, will we have multiple listeners if we have multiple blocks on a page? This could cause the handler to be called multiple times

you should be able to add this listener to a only the relevant node by assigning the outer div  ref="thisBlock", then only add the listener to:
``` this.$refs.thisBlock.addEventListener("bokehStateUpdate", this.handleBokehEvent);```

**Comment by ml-evs** [comment]
Made some fixes post-discussion with @jdbocarsly. Known limitations still:

- [ ] The event handler currently triggers per block on the page, with the wrong data. We need to find a way to scope the handler, and then dispatch it correctly to the right block. Ideally this would be encapsulated in a function so that we don't have too much fiddly boilerplate in the Bokeh callbacks.
- [ ] We should probably rename `bokehStateUpdate`
- [ ] The whole block currently updates; it would be nice to be able to also do partial updates

File: webapp/src/components/datablocks/DataBlockBase.vue (hunk 2 above)
Comment (ml-evs) [reply webapp/src/components/datablocks/DataBlockBase.vue]: Still struggling a lot with this... following the approach with $refs, it seems that the block events bubble up to the `block-list-item` defined in the `EditPage` and skip the `DatablockBase`.

Comment (ml-evs) [reply webapp/src/components/datablocks/DataBlockBase.vue]: Perhaps the easiest resolution is adding a `ref` to the `DataBlockBase` and then directly calling:

```
<template>
 <DataBlockBase :item_id="item_id" :block_id="block_id" ref="xrdBlockBase">
</template>
...
this.$refs.xrdBlockBase.handleBokehEvent(event);
```

ml-evs: The ref approach was a bit of a non-starter, instead passing block ID and having each event reject if not matching is a good compromise (and only scales badly with number of blocks on a page)

BenjaminCharmes (approved):
LGTM, everything works great!

PR #1062, opened 2025-03-11
Add concept of user groups

ml-evs:
This PR adds the concept of groups of users, and the corresponding API routes and UIs. The aim of groups is to be able to provide blanket permissions to subsets of users. Some statements:

- Users are created without any group.
- Currently only admins can create groups and assign users to them, though it is already possible to list users as admins for a group (and in future they will control access).

Still to-do:

- [x] Use these in permissions lookups
- [ ] Allow users to assign samples as visible to groups on creation and edit
- [ ] Default permissions for groups

### DIFF
--- a/webapp/src/components/AdminDisplay.vue
+++ b/webapp/src/components/AdminDisplay.vue
@@ -6,7 +6,7 @@
     <template v-if="selectedItem === 'Access Tokens'">
       <TokenTable />
     </template>
-    <template v-if="selectedItem === 'Groups'"> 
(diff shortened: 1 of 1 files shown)


BenjaminCharmes:
Users can now see items if they:
1. Created the item (existing)
2. Are a member of a group the item is shared with (new)
3. Are an admin of a group the item is shared with (new)

- Users can only share items with groups they belong to or administer

I've also updated the UI:

- In the administrator panel, you can now create, update or delete groups.
- On an item page, you can share this item with the groups you belong to.

PR #1087, opened 2025-04-03
Fix plugin loading not triggering on server startup

### LINKED ISSUE #1084: In situ block not loaded correctly/at all
In #1040, I forgot to actually load the plugin for in situ to get it to run on server startup. We also need to fix the name in the `resources.js` so that the custom Vue code can be used for it.

BenjaminCharmes:
Closes #1084 

`insitu-nmr` plugin is correctly loaded on server startup. 
Updates plugin name in `resources.js` and `NMRInsituBlock.vue`

[WIP]:
- Need to understand why these changes make tests failed

### DIFF
--- a/pydatalab/src/pydatalab/blocks/__init__.py
+++ b/pydatalab/src/pydatalab/blocks/__init__.py
@@ -40,15 +40,16 @@ def load_block_plugins():
         if not issubclass(block, DataBlock):
             raise ValueError(f"Plugin {block} must be a subclass of DataBlock")
 
-        if block.blocktype in BLOCK_TYPES:
-            raise ValueError(f"Block type {block.blocktype} is already defined")
-
-        BLOCK_TYPES[block.blocktype] = block
-        BLOCKS.append(block)
         block_plugins[block.blocktype] = block
+
+        if block.blocktype not in BLOCK_TYPES:
+            BLOCK_TYPES[block.blocktype] = block
+            BLOCKS.append(block)
 
     return block_plugins
 
+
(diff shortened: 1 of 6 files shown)


pydatalab/src/pydatalab/blocks/__init__.py
```diff
@@ -40,13 +40,12 @@ def load_block_plugins():
...
-        if block.blocktype in BLOCK_TYPES:
-            raise ValueError(f"Block type {block.blocktype} is already defined")
-
-        BLOCK_TYPES[block.blocktype] = block
-        BLOCKS.append(block)
         block_plugins[block.blocktype] = block
 
+        if block.blocktype not in BLOCK_TYPES:
+            BLOCK_TYPES[block.blocktype] = block
+            BLOCKS.append(block)
```
ml-evs, line 47:
```suggestion
            BLOCKS.append(block)
            block_plugins[block.blocktype] = block

File: pydatalab/src/pydatalab/blocks/__init__.py [hunk 2]
```diff
@@ -40,16 +40,18 @@ def load_block_plugins():
...
+            block_plugins[block.blocktype] = block
```
ml-evs, line 48:
```suggestion
```

ml-evs:
Found the issue with the tests! The docker container does not have `setuptools` installed, whereas all our local environments and presumably the CI have it installed by default, so I've added it as a runtime dependency too.

ml-evs (approved):

PR #1098, opened 2025-04-04
Add regex-based alternative to full-text search

### LINKED ISSUE #679: Free text search could be improved

ml-evs:
After noodling around in #993, I think the conclusion is that its not worth maintaining a potentially large separate custom index for FTS that works with only a few characters. Instead, as a bit of a prototype, this PR adds a simple regex-based search that can triggered in the item select by prefixing the query with "%". This seems to work fairly well at our typical database sizes, but might need some optimisations. For now it also only uses the fields that are currently part of the free-text index, but this could be expanded (and we can add real indices over these fields if performance is bad).

Closes #679?


ml-evs:
Going to merge this as a bit of a hidden feature, want to test it out on some real deployments before making it the default

PR #1099, opened 2025-04-04
Make media block resizable

### LINKED ISSUE #1093: Media block resize
The default size for the image in the media block is sometimes a bit unwieldy - would be useful to be able to manually resize sometimes.

jdbocarsly:
Closes #1093


ml-evs (approved):
Resize is a bit janky but still as massive improvement, thanks!

PR #1101, opened 2025-04-08
Added synthesis information to starting materials and made component collapsible

be-smith:
Changed synthesis information to be added to starting materials as well as samples.

If the synthesis info is empty it will start collapsed upon loading the page - otherwise it will start expanded.


pydatalab/src/pydatalab/models/starting_materials.py
```diff
@@ -66,6 +66,12 @@ class StartingMaterial(Item):
 
     comment: Optional[str] = Field(alias="Comments")
 
+    synthesis_constituents: List[Constituent] = Field([])
```
ml-evs, line 69: Minor, but as discussed I would refactor this with a `HasSynthesisInfo` trait that can then be used by both samples and starting materials, though we can discuss

webapp/src/components/SynthesisInformation.vue
```diff
@@ -86,11 +154,78 @@ export default {
...
+
+.content-container {
+  overflow: hidden;
+  max-height: none;
+  transition: max-height 0.4s ease-in-out;
+}
+
+.content-container.collapsed {
+  max-height: 0;
+}
```
ml-evs, line 230: I guess most of this style is not needed and was just taken from the collapsible data block?

webapp/src/components/SynthesisInformation.vue
```diff
@@ -1,20 +1,35 @@
...
-    <label class="mr-2 pb-2">Synthesis Information</label>
-    <div class="card component-card">
-      <div class="card-body pt-2 pb-0 mb-0 pl-5">
-        <CompactConstituentTable
-          id="synthesis-table"
-          v-model="constituents"
-          :types-to-query="['samples', 'starting_materials']"
-        />
+  <div :id="block_id" class="data-block">
+    <div class="datablock-header collapsible" :class="{ expanded: isExpanded }">
```
ml-evs, line 3:
```suggestion
    <div class="datablock-header collapsible" :class="{ expanded: isExpanded }">
```

some of the labelling here is likely the issue too, make sure the css classes are appropriately named (see comment below)

ml-evs (commented):
I imagine the test failures are related to the "hidden by default" for this component, see e.g.:

https://cloud.cypress.io/projects/4kqx5i/runs/3060/test-results/69349704-c474-4b9f-a10f-36a43a97b5ef/replay?att=1&ts=1744117996161.6&utm_campaign=view+Test+Replay&utm_medium=failed&utm_source=github which shows that it is failing to find the search box when first loading a page (should be as easy as editing the test to find the heading, clicking it to uncollapse before looking for the search).

2025-04-08-143437_519x410_scrot

It might be worth figuring out how to get the cypress tests running locally on your end (should be as simple as `yarn run cypress run --e2e`... but often isn't). Otherwise, the cypress cloud can be very helpful in debugging too (lets you replay the tests in the browser even if they run in the CI).

webapp/src/components/SynthesisInformation.vue
be-smith: I think I have removed the spurious ones now - a lot was left over from the data block or previous approaches to expanding and collapsing the block

pydatalab/src/pydatalab/models/samples.py
```diff
@@ -1,24 +1,25 @@
...
 
     type: str = Field("samples", const="samples", pattern="^samples$")
 
     chemform: Optional[str] = Field(example=["Na3P", "LiNiO2@C"])
     """A string representation of the chemical formula or composition associated with this sample."""
 
-    synthesis_constituents: List[Constituent] = Field([])
-    """A list of references to constituent materials giving the amount and relevant inlined details of consituent items."""
+    # synthesis_constituents: List[Constituent] = Field([])
+    # """A list of references to constituent materials giving the amount and relevant inlined details of consituent items."""
```
ml-evs, line 19:
```suggestion
```

pydatalab/src/pydatalab/models/samples.py
```diff
@@ -1,24 +1,25 @@
...
 
-    synthesis_description: Optional[str]
-    """Free-text details of the procedure applied to synthesise the sample"""
+    # synthesis_description: Optional[str]
+    # """Free-text details of the procedure applied to synthesise the sample"""
```
ml-evs, line 22:
```suggestion
```

pydatalab/src/pydatalab/models/traits.py
```diff
@@ -78,3 +78,24 @@ def add_missing_collection_relationships(cls, values):
...
+
+class HasChemInfo:
+    smile: Optional[str] = Field(None)
+    """A SMILES string representation of the chemical structure associated with this sample."""
+    inchi: Optional[str] = Field(None)
+    """An InChI string representation of the chemical structure associated with this sample."""
+    inchi_key: Optional[str] = Field(None)
+    """An InChI key representation of the chemical structure associated with this sample."""
+    """A unique key derived from the InChI string."""
+    chemform: Optional[str] = Field(None)
```
ml-evs, line 101: Probably this is meant for a different PR/branch?

ml-evs (changes requested):
Looking good @be-smith, just a few final tweaks

ml-evs:
Quick note in case I forget in the next 16 minutes -- we should add the starting material info to the add item modal

ml-evs (approved):
Awesome work, thanks @be-smith!

PR #1105, opened 2025-04-09
Add search link to CAS Common Chemistry when CAS provided

ml-evs:
Looks like this:

https://github.com/user-attachments/assets/e433e3b8-17f0-4560-be1e-bf6734584bdd


jdbocarsly (approved):
looks good! Excited for when we can query an API and pull this info into datalab itself

ml-evs:
> looks good! Excited for when we can query an API and pull this info into datalab itself

I looked into this again and signed up at https://www.cas.org/services/commonchemistry-api but might need to be careful that we're not crossing into commercial use

PR #1106, opened 2025-04-09
Multiselect filtering for "no values" in nblocks column

### LINKED ISSUE #1104: Possibility of filtering multiselect for "no values"
Now that we have block filtering in the table (https://github.com/datalab-org/datalab/pull/1080), it seems like we should be able to match for "give me all samples with no blocks". Probably this is a feature we disabled for other filters, that we could reintroduce here (cc @BenjaminCharmes).

BenjaminCharmes:
Closes #1104

### DIFF
--- a/webapp/src/components/DynamicDataTable.vue
+++ b/webapp/src/components/DynamicDataTable.vue
@@ -371,13 +371,15 @@ export default {
     uniqueBlockTypes() {
       const itemsWithBlocks = this.data.filter((item) => item.blocks && item.blocks.length > 0);
 
-      return Array.from(
-        new Map(
-          itemsWithBlocks
-            .flatMap((item) => item.blocks)
-            .map((block) => [block.title, { title: block.title }]),
-        ).values(),
+      const blockTypesMap = new Map(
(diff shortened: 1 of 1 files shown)


jdbocarsly:
I haven't tried on this branch, but I was also noticing that "match any" seems to always match samples with 0 blocks. Is that easy to fix?

BenjaminCharmes:
Thank for the review!

- I can't reproduce the behaviour Josh describes (Either on main or on this branch)
- I, also, can't reproduce the first behaviour Matt describes with "match all -> no blocks" 
- For the second behaviour you describe, it's probably because the box is empty (and doesn't have ‘0’) when there are no blocks. I can look for a way to change that

I'm going to take a closer look at all this!

ml-evs:
Realised what is causing this behaviour: when you copy a sample that has blocks, the block is copied identically and then the API summary doesn't include it even though `nblocks: 1` -- weird bug we need to fix

ml-evs (approved):
Works well, now that we figured out that weird bug in the API!

PR #1107, opened 2025-04-09
Fixed calling _format_ftir_plot method named when called to show data

be-smith:
Bug due to me renaming the _formati_ftir_plot method but not updating when it's actually called in the block itself


ml-evs (approved):
Oops, this indeed highlights the need to refactor some of our tests to properly test block usage. Thanks!

PR #1109, opened 2025-04-09
Adjust block and file counter style

ml-evs:
Tone down the red circles and replace it with an underline, to attempt to indicate that you can still hover

2025-04-10-003308_1285x403_scrot


BenjaminCharmes (approved):
👍

PR #1110, opened 2025-04-10
Remove datatable-sort-badge when only one column is sorted

### LINKED ISSUE #1108: Sample table: Don't display "1" when only one column is sorted
Can we tweak the sample table (and other fancy tables) so that the sort order indicator [the green "(1)"] doesn't show up if only one column is sorted? The reason is that it adds extra horizontal space into the column and causes the columns to have to resize whenever one is selected for sorting. I think its only needed if someone selected multiple columns and wants to see the order of sorting. 

Image

BenjaminCharmes:
Closes #1108 

The numbers always appear when you're in `sort-mode="multiple"` and prime-vue only removes them when you're in `sort-mode="single"`. This allows the user to know that they can sort several columns at once.

I guess we can find a way of removing the "1" when only one column is sorted, but I couldn't find anything in the prime-vue docs, or even using the custom theme we have in datalab (the two columns have exactly the same css classes), so I had to add another custom css class.

It's a first attempt to see what it could be.

### DIFF
--- a/webapp/src/components/DynamicDataTable.vue
+++ b/webapp/src/components/DynamicDataTable.vue
@@ -480,11 +480,9 @@ export default {
   methods: {
     onSort(event) {
       const sortedColumns = event.multiSortMeta || [];
-      console.log(sortedColumns);
 
       this.$nextTick(() => {
         const tableHeaders = document.querySelectorAll(".p-datatable-sortable-column");
-        console.log(tableHeaders);
 
         tableHeaders.forEach((header) => {
           if (sortedColumns.length === 1) {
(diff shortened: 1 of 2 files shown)


BenjaminCharmes:
> Having some trouble testing this locally, it seems to work well for the ID column, but I'm getting errors when trying to filter with any of the enum columns (e.g., item type, creators). Does this work for you locally @BenjaminCharmes?

I've just tried again and everything seems to be working fine for me

Screenshot 2025-04-23 at 09 30 53
-
Screenshot 2025-04-23 at 09 31 06

ml-evs (approved):
Thanks @BenjaminCharmes, this does seem to work well! I think it would be a good idea to add a couple more tests for the data table component tests for search, sort and filtering (also with up to date example data too with block types, nfiles). I'll raise a separate issue to track this!

PR #1123, opened 2025-04-14
Default to opening a new tab when clicking on the sample table

### LINKED ISSUE #1114: Default to modified click when opening a sample
Used to be the case, but now it opens in the current tab

BenjaminCharmes:
Closes #1114


ml-evs (approved):
This LGTM, thanks @BenjaminCharmes! I guess all we lose out on here is the ctrl + click for multi-select, but I didn't actually know we could do this anyway... I think this is a more sensible default and it solves #1114.

PR #1124, opened 2025-04-14
Use localStorage for persistent column hiding in Datatable

### LINKED ISSUE #1113: Persistence of user settings for table
- [ ] Want searches to persist after clicking on a sample

BenjaminCharmes:
Closes #1113 & #1135

Use localStorage in DataTable to store:

- which column are displayed/hide
- width of each column

### DIFF
--- a/webapp/src/components/DynamicDataTable.vue
+++ b/webapp/src/components/DynamicDataTable.vue
@@ -686,7 +686,11 @@ export default {
     },
     onToggleColumns(value) {
       this.$nextTick(() => {
-        this.selectedColumns = [...value];
+        this.selectedColumns = [...value].sort((a, b) => {
+          const indexA = this.availableColumns.findIndex((col) => col.field === a.field);
+          const indexB = this.availableColumns.findIndex((col) => col.field === b.field);
+          return indexA - indexB;
+        });
       });
     },
     onStateSave(state) {
--- a/webapp/src/components/DynamicDataTable.vue
+++ b/webapp/src/components/DynamicDataTable.vue
@@ -5,6 +5,7 @@
     </div>
 
     <DataTable
+      ref="datatable"
       v-model:filters="filters"
       v-model:selection="itemsSelected"
       v-model:select-all="allSelected"
@@ -27,6 +28,7 @@
       @select-all-change="onSelectAllChange"
       @page="onPageChange"
       @sort="onSort"
+      @column-resize-end="onColumnResize"
     >
       <!-- v-model:expandedRows="expandedRows" -->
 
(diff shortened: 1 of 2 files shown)


ml-evs (approved):
I'm not sure search value is the first one we should implement here (you can go to a sample without losing the table), but selected columns/width would be great!

ml-evs (changes requested):
didn't mean to accept...

BenjaminCharmes:
> I'm not sure search value is the first one we should implement here (you can go to a sample without losing the table), but selected columns/width would be great!

I must have misunderstood the issue #1113 
I'll try to add the size of the columns, and then, add the hide/display columns as soon as #1118 is merged.

BenjaminCharmes:
It was way easier than I expected, because primevue already have a localStorage state management for the datatable.
Perhaps this could also be used to display/hide columns.

Do we want to be able to move the columns as well, or do they always have to stay in the same place?

ml-evs:
> It was way easier than I expected, because primevue already have a localStorage state management for the datatable. Perhaps this could also be used to display/hide columns.

Great!
 
> Do we want to be able to move the columns as well, or do they always have to stay in the same place?

For now, I think stay in the same place...

jdbocarsly:
Does this one also save filters? I think that would be helpful, since sometimes you set up a complex series of filters and then lose it

ml-evs:
> Does this one also save filters? I think that would be helpful, since sometimes you set up a complex series of filters and then lose it

I really don't expect people to use filters this way, at least now they won't lose things when they click on a sample, but I think I'd be surprised as a user if filters persisted over sessions?

ml-evs (commented):
Functionality-wise, this seems to be working well @BenjaminCharmes! We can discuss in a mo, but perhaps we should get some usability stuff in alongside this PR:

- [ ] resize columns once user changes them (I know we spoke about this a bit, might be awkward to do...)
- [ ] ability for underlying components e.g., `SampleTable` to specify default columns that are not necessarily the entire range (c.f. #1163) -- if the selected columns stored locally cannot be loaded, or are empty, we can defer to this default
- [ ] Currently if the user logs out, this local storage persists... maybe we should wipe local storage when they log out, or use session storage?

BenjaminCharmes:
I was able to remove all the information from the prime-vue state and keep only what I wanted, and even add the names of the columns displayed, so it's even simpler. And I can keep the state without putting the filter, item selected ... in the localStorage.
I'll take some time, probably in another PR if you want to merge this one, for the other issues.

ml-evs (approved):
Looks good, thanks @BenjaminCharmes! Let's not forget to raise any more issues about stuff we mentioned here for future PRs. I think the main point to think about is how to deal with this local storage on multi-user systems, making sure it gets cleared or cached elsewhere when the user logs out, and that the user can only ever use their own stored data even if another user's is present in the browser cache. As discussed, just for column hiding, I don't think we need to worry.

PR #1128, opened 2025-04-16
Add CIF->PXRD calculator and refactor XRD block to better show multiple patterns

### LINKED ISSUE #1120: Ability to upload CIF references to XRD block

ml-evs:
Closes #1120 by adding a [matador](https://github.com/ml-evs/matador)-based PXRD calculator for CIF files.

Bonus: record peak information in database where possible, and update `FileSelectDropdown` to allow for "All compatible files" option to be turned on.

### DIFF
--- a/pydatalab/src/pydatalab/apps/xrd/utils.py
+++ b/pydatalab/src/pydatalab/apps/xrd/utils.py
@@ -192,3 +192,29 @@ def parse_rasx_zip(filename: str) -> pd.DataFrame:
             "intensity": xrd_data["intensity"],
         }
     )
+
+
+def compute_cif_pxrd(filename: str, wavelength: float) -> tuple[pd.DataFrame, dict]:
+    """Parses a CIF file and returns a pandas DataFrame with columns
+    twotheta and intensity.
+
+    Parameters:
+        filename: The file to parse.
+
+    """
+    from matador.fingerprints.pxrd import PXRD
+    from matador.scrapers.cif_scraper import cif2dict
+
+    structure, success = cif2dict(filename)
+    if not success:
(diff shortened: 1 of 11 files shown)


pydatalab/src/pydatalab/apps/xrd/utils.py
```diff
@@ -186,3 +186,31 @@ def parse_rasx_zip(filename: str) -> pd.DataFrame:
             "intensity": xrd_data["intensity"],
         }
     )
+
+
+def compute_cif_pxrd(filename: str, wavelength: float) -> tuple[pd.DataFrame, dict]:
+    """Parses a CIF file and returns a pandas DataFrame with columns
+    twotheta and intensity.
```
ml-evs, line 193:
```suggestion
    twotheta and intensity, and a dictionary with peak information.
```

ml-evs:
Note to self: expose raw data in this PR

ml-evs:
> Note to self: expose raw data in this PR

Actually going to push this back to a later PR so that we can avoid holding this up

jdbocarsly:
Seems to work for me!  We can iterate on the UX with ability to turn on/off patterns, offset them, display ticks etc. But its a really useful thing to have in there. 

When choosing all files, a warning comes up (even though it plots fine). Also, the second warning is very frequently on our data sets since they have points at 0 and the log calculation throws the warning. Can you suppress that? It gets annoying.
Screenshot 2025-04-28 at 11 14 54 PM

ml-evs:
Not totally awful if you limit it to a few patterns:

2025-05-23-214621_740x620_scrot

Think we need to think generally about making better use of space for plots like this though (see if we can't make the plot resizable), and use the whitespace outside for things like the legend.

Issue #658, opened 2024-03-22
Handling of CAS for starting materials (and potentially samples)


jdbocarsly:
Currently we keep track of CAS numbers for chemicals in the inventory, but don't do anything special for them. We should integrate features to:

- Add features to automatically populate fields like chemical formula, chemical names, and hazards based on CAS. This can be done with pubchem's API, or potentially a different service. 
- Automatically find and attach a SDS (generally in pdf format) to a sample based on CAS>
- Make the CAS field in `StartingMaterialInformation.vue` a clickable link that takes the user to the CAS registry page (pubchem, chemspider are other options.)

Also discussed in comments in #642

PR #1129, opened 2025-04-17
Add PDF support in MediaBlock using an iframe

ml-evs:
This was way easier than I expected just using an iframe (rather than having to bundle e.g., PDF.js and the like). Will be useful for e.g., #658


BenjaminCharmes (approved):
works perfectly 👍

PR #1133, opened 2025-04-18
UV-vis block

be-smith:
Added a block for plotting UV-Vis data. Currently works on .Raw8.txt files from the grey group UV-Vis spectrometer. Requires two files minimum for plotting, a reference file and the sample file - but multiple samples can be plotted at the same time if more than two files are supplied.

Absorbance is calculated using the standard approach, where I0 is the reference intensity and I the sample intensity.
A = log(I0/I)

Included is a vue component for selecting and ordering multiple files - could be useful for other blocks (Like echem if files need to be stitched together)

Screenshot 2025-04-18 at 18 50 52

Feedback on this component specifically would be nice - was a bit of a vibe coding session.

Example data not uploaded yet as waiting on permission from the owner.


pydatalab/example_data/UV-Vis/1908047U1_0000.Raw8(1).TXT
ml-evs, line 1: Can we get rid of the (1) in these file names? It's asking for trouble... I think right now these are stripped in datalab. Also, in this case, Finder (?) has decided to add the (1) to the file extension...

pydatalab/src/pydatalab/apps/uvvis/__init__.py
```diff
@@ -0,0 +1,170 @@
...
+from bokeh.models import HoverTool
+
+from pydatalab.blocks.base import DataBlock
+from pydatalab.bokeh_plots import DATALAB_BOKEH_THEME, selectable_axes_plot
+from pydatalab.file_utils import get_file_info_by_id
+from pydatalab.logger import LOGGER
+
+
+class UVVisBlock(DataBlock):
+    accepted_file_extensions = (".txt", ".TXT")
```
ml-evs, line 16:
```suggestion
    accepted_file_extensions = (".txt",)
```

These should already be case-insensitive (all comparisons should be done case-insensitively)

Alternatively, you might consider specialising it really to just raw8 files (would need to test that "double extensions" like this work fine, but they should)

```suggestion
    accepted_file_extensions = (".Raw8.txt", )
```

pydatalab/src/pydatalab/apps/uvvis/__init__.py
```diff
@@ -0,0 +1,170 @@
...
+    def parse_uvvis_txt(self, filename: Path) -> pd.DataFrame:
+        """
+        Parses a UV-Vis .txt file into a pandas DataFrame
+        Args:
+            filename (Path): Path to the .txt file
+        Returns:
+            pd.DataFrame: DataFrame containing the UV-Vis data with columns for wavelength and absorbance
+        """
+        # Read the file, skipping the first 7 rows and using the first row as header
+        data = pd.read_csv(filename, sep=r";", skiprows=7, header=None)
```
ml-evs, line 39: Definitely always 7? Alternatively we could look for the blank lines around the header that these files all seem to have

pydatalab/src/pydatalab/apps/uvvis/__init__.py
```diff
@@ -0,0 +1,170 @@
...
+            LOGGER.warning("No file set in the DataBlock - selected_file_order")
+            return
+
+        else:
+            file_info = []
+            for file_id in self.data["selected_file_order"]:
+                file_info.append(get_file_info_by_id((file_id), update_if_live=True))
+            # Check if the file is in the accepted file extensions
+            for file in file_info:
+                ext = os.path.splitext(file["location"].split("/")[-1])[-1].lower()
```
ml-evs, line 150: This is maybe the bit that doesn't compare work with ext "raw8.txt" -- again in future we should abstract this away, but perhaps for this particular case you can also check the "-2"th extension

pydatalab/tests/apps/test_uvvis.py
```diff
@@ -0,0 +1,64 @@
...
+    for f in data_files:
+        if f.suffix in UVVisBlock.accepted_file_extensions:
+            logging.info(f"Loading file: {f.name}")
+            if f.name == "1908047U1_0000.Raw8(1).TXT":
+                logging.info(f"Loading reference file: {f.name}")
+                reference_df = UVVisBlock.parse_uvvis_txt(f)
+            elif f.name == "1908047U1_0001.Raw8(1).txt" or f.name == "1908047U1_0060.Raw8.txt":
+                sample_dfs.append(UVVisBlock.parse_uvvis_txt(f))
+    if reference_df is None:
+        raise ValueError("Reference file '1908047U1_0000.Raw8(1).TXT' not found.")
```
ml-evs, line 41: I would factor this out into a separate test, no need to use the fixture with data files here if not all data files are equal (though you could keep the simple parsing test running uniformly on all of them)

ml-evs (changes requested):
Some initial comments on the Python side, will try out the full block soon!

be-smith:
So I have changed the extensions for the block to Raw8.txt

Dealing with multiple suffixes is a bit janky, the following seems to work but has to be called often in the code or in tests as far as I can tell to deal with upper and lower cases matching. This is robust for if there is only one extension like .txt - so could be the standard going forwards. Let me know if there is a better way of doing this!

`
ext = "".join(Path(file["location"]).suffixes).lower()
`
`
if ext not in {ext.lower() for ext in self.accepted_file_extensions}:
`
`do something`

webapp/src/components/FileMultiSelectDropdown.vue
```diff
@@ -0,0 +1,350 @@
...
+        this.extensions.length === 0 || (this.extensions.length === 1 && this.extensions[0] === "");
+
+      return allIds.filter((file_id) => {
+        const file = this.all_files_map.get(file_id);
+        if (!file) return false; // Skip if file details not found
+
+        if (showAll) return true; // Show all if extensions are empty or just ""
+
+        const filename = file.name?.toLowerCase() || "";
+        return this.extensions.some((extension) => filename.endsWith(extension.toLowerCase()));
```
ml-evs, line 169: I guess I was more worried about file selection on the UI side e.g., this bit, but you already have a custom component for it that works for UV Vis this way. I'll try to get around to reviewing the UI code tomorrow!

webapp/src/resources.js
```diff
@@ -71,6 +72,7 @@ export const blockTypes = {
   ms: { description: "Mass Spectrometry", component: MassSpecBlock, name: "Mass Spectrometry" },
   chat: { description: "Virtual assistant", component: ChatBlock, name: "Virtual Assistant" },
   ftir: { description: "FTIR", component: BokehBlock, name: "FTIR" },
+  uvvis: { description: "UV-Vis", component: UVVisBlock, name: "UV-Vis" },
```
ml-evs, line 75:
```suggestion
  "uv-vis": { description: "UV-Vis", component: UVVisBlock, name: "UV-Vis" },
```

Can we use the hyphenated version everywhere please? (both in the Python and JS)

ml-evs (approved):
This looks good from my side now, thanks @be-smith! I'm just going to ask that @BenjaminCharmes tries it out and reviews the JS multi-select more finely before we merge.

BenjaminCharmes (commented):
Can't remove this comment, wanted to approve!

BenjaminCharmes (approved):
LGTM, everything works very well 👌

Maybe just a comment on CSS: a few classes are useless because they are already bootstrap classes: .active, .disabled,:focus. It may be slightly different here (as in the case of :focus, which adds a blue border) But I'm not sure we want to add too much custom CSS to keep it uniform everywhere.

PR #1146, opened 2025-04-23
Add searching, filtering and sorting test in Datatable

### LINKED ISSUE #1139: Missing tests for searching, filtering and sorting in the data table
We need some tests for the above, as I think a recent primevue (4.1 -> 4.3 for primevue/core) update in #1003 may have broken some features (see sub issues) Locally I'm observing:

- [ ] errors when trying to filter on enum value columns,
- [ ] cannot clear filters once they are created (no clear button)
- [ ] We could simply revert the updates, or try to fix these with the new version.

BenjaminCharmes:
Closes #1139

I may have missed some things to test, so don't hesitate to tell me and I'll add them!

### DIFF
--- a/webapp/cypress/component/SampleTableTest.cy.jsx
+++ b/webapp/cypress/component/SampleTableTest.cy.jsx
@@ -28,19 +28,79 @@ describe("SampleTable Component Tests", () => {
               name: "Sample 1",
               chemform: "H2O",
               date: "2023-09-01T12:34:56Z",
-              collections: ["collection1"],
+              collections: [{ collection_id: "collection1" }],
               creators: [{ display_name: "Creator 1" }],
               nblocks: 1,
+              nfiles: 1,
+              blocks: [{ title: "NMR" }],
+            },
+            {
+              item_id: "sample2",
+              type: "samples",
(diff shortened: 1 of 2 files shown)


webapp/cypress/component/SampleTableTest.cy.jsx
```diff
@@ -28,19 +28,79 @@ describe("SampleTable Component Tests", () => {
...
+              item_id: "sample2",
+              type: "samples",
+              name: "Sample 2",
+              chemform: "H2O",
+              date: "2023-09-02T12:34:56Z",
+              collections: [{ collection_id: "collection2" }],
+              creators: [{ display_name: "Creator 2" }],
+              nblocks: 2,
+              nfiles: 2,
+              blocks: [{ title: "NMR" }, { title: "insitu" }],
```
ml-evs, line 47: I think these should perhaps be looking for `blocktype: 'nmr'` rather than the title since the titles are editable by users, but maybe that's for another PR...

ml-evs (approved):
Looks really good, thanks @BenjaminCharmes -- will see if this fixes my local issues too!

PR #1160, opened 2025-04-30
Bes/uv vis extra

be-smith:
Added watchers to fileselectdropdown and filemultiselect to check if the available files change - and if the file(s) selected are removed it removes them from the modelvalue - and therefore the block


ml-evs (approved):
Works well, thanks @be-smith (and sorry for the delay!)

PR #1167, opened 2025-05-02
Assorted fixes for starting materials in collections

### LINKED ISSUE #790: Collections cannot contain starting materials
This is a consequence of the UI summary page, but now with the API it is possible to put any item in a collection, so we should reflect this in the API response.

ml-evs:
Closes #790

Some fixes to allow starting materials to appear in collections in the UI, plus some assorted fixes to the item graph to better show starting materials.

- [x] Refactor summary endpoint so that starting materials can be requested when called from collection table
- [x] Add a default search index of `date` for all items and use it by default when returning inventory (rather than implicitly using `ctime`
- [x] Update the item graph endpoint so ignore collections by default, which are a frequent cause of errors
- [x] Some tweaks for logging and cleaning up MongoDB connections


BenjaminCharmes (approved):
Works very well !

PR #1170, opened 2025-05-05
Improving column management in Datatable

### LINKED ISSUE #1172: Components should be able to specify default columns
- [ ] ability for underlying components e.g., `SampleTable` to specify default columns that are not necessarily the entire range (c.f. #1163) -- if the selected columns stored locally cannot be loaded, or are empty, we can defer to this default and/or let the user reset the table

_Originally posted by @ml-evs in https://github.com/datalab-org/datalab/pull/1124#pullrequestreview-2811801254_

BenjaminCharmes:
Closes #1172

- Add a property in Datatable columns allowing certain columns to be hidden by default
- Add a button to reset the Datatable


webapp/src/components/DynamicDataTableButtons.vue
```diff
@@ -270,6 +279,11 @@ export default {
     columnLabel(option) {
       return option.label || option.header || option.field;
     },
+    resetTable() {
+      if (window.confirm("Are you sure you want to completely reset the Datatable?")) {
```
ml-evs, line 283:
```suggestion
      if (window.confirm("Are you sure you want to reset your preferences (visible columns, widths) for this table?")) {
```

Can we make this slightly clearer?

ml-evs (commented):
Thanks for this @BenjaminCharmes, functionality works very well!

Just a minor suggestion below, then one general comment.

For my particular setup (font sizes, pixel density etc.), the refresh button causes the left-most buttons to collapse one another (as it does when you make the default window too narrow anyway). I know this is probably going to be better resolved in #1028, but I wonder what options we have to tweak this here.

Possible ideas:

- [ ] Hide the disabled `Selected` option until items are actually selected (which is what it did at some point)
- [ ] Add a "settings" button that hides the column dropdown and refresh button under a separate dropdown (maybe overkill). Not sure about this
- [ ] Explicitly make two rows of buttons so they don't collide, somehow the "add/scan buttons" at the top and the others on the right (again, back to how it was ages ago) Also not sure about this
- [ ] Simply try narrowing the search box so that we get away with it for one more PR!

Issue #1159, opened 2025-04-30
Only emit warnings when an unknown block type is encountered


ml-evs:
This currently breaks the dev server which can be annoying when switching between unfinished/merged block types. We should just have a fallback that says "block not implemented on this server" or the like.

ml-evs:
Importantly, to extend the above, the block data should not be nuked (or touched by the UI) in case this is simply a config bug and the block will reappear on the server in the future.

PR #1188, opened 2025-05-16
Refactor app block loading to prevent optional dependencies from breaking dev builds

### LINKED ISSUE #1178: Move app blocks so that base data block is importable without extra deps

ml-evs:
Closes #1178 by moving all block loading into the apps module.

This still feels like a bit of a temporary solution, ideally this could be refactored away at some point with better hierarchy between apps and blocks, but for now this is the way of doing it with minimal server import changes (otherwise we'd have to move the base data block and change every block and plugin to use a new location of base data block). The overall problem stems from the fact that the base block is implemented in `pydatalab.blocks.base` and we want to import it, and many child classes of it, from the top-level `pydatalab.blocks`. This PR simply moves the loading stage into `pydatalab.apps`.

Also makes pymongo a core dep.

Relatedly we should also worry about https://github.com/datalab-org/datalab/issues/1159 -- providing a proper error when an unknown block is encountered

NB/ as this changes the internal datalab API, we should reserve this for the next minor release (also hopefully soon).

### DIFF
--- a/pydatalab/pyproject.toml
+++ b/pydatalab/pyproject.toml
@@ -27,6 +27,7 @@ dependencies = [
     "pydantic[email, dotenv] < 2.0",
     "pint ~= 0.24",
     "pandas[excel] ~= 2.2",
+    "pymongo ~= 4.7, < 4.11",
 ]
 
 [project.urls]
@@ -46,7 +47,6 @@ version_scheme = "post-release"
 
 [project.optional-dependencies]
 server = [
-    "pymongo ~= 4.7, < 4.11",
     "Flask ~= 3.0",
     "Flask-Login ~= 0.6",
     "Flask-Cors ~= 5.0",
--- a/pydatalab/src/pydatalab/apps/__init__.py
+++ b/pydatalab/src/pydatalab/apps/__init__.py
@@ -1,2 +1,125 @@
-# This import is required to prevent circular imports for application-specific blocks
-from pydatalab.blocks.base import DataBlock  # noqa
+"""This module provides a convenience wrapper for loading all
+'app' blocks, which may or may not be available.
+
+"""
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    # This import is required to prevent circular imports for application-specific blocks
+    from pydatalab.blocks.base import DataBlock  # noqa
+
+from pydatalab.blocks import COMMON_BLOCKS
+
+
+def _check_error(e):
+    if "circular" in str(e):
+        raise ImportError(e) from e
+
+
+def load_app_blocks():
+    app_blocks: list[type["DataBlock"]] = []
+
+    try:
+        # A dummy block that is used to check that bad blocks do not break the import
+        from pydatalab.apps._canary import CanaryBlock
+
+        app_blocks.append(CanaryBlock)
+    except ImportError as e:
+        _check_error(e)
+
+    try:
+        from pydatalab.apps.chat import ChatBlock
+
+        app_blocks.append(ChatBlock)
+    except ImportError as e:
+        _check_error(e)
+
+    try:
+        from pydatalab.apps.echem import CycleBlock
+
+        app_blocks.append(CycleBlock)
+    except ImportError as e:
+        _check_error(e)
+
+    try:
+        from pydatalab.apps.eis import EISBlock
+
+        app_blocks.append(EISBlock)
+    except ImportError as e:
+        _check_error(e)
+
+    try:
+        from pydatalab.apps.ftir import FTIRBlock
+
+        app_blocks.append(FTIRBlock)
+    except ImportError as e:
+        _check_error(e)
+
+    try:
+        from pydatalab.apps.nmr import NMRBlock
+
+        app_blocks.append(NMRBlock)
+    except ImportError as e:
+        _check_error(e)
+
+    try:
+        from pydatalab.apps.raman import RamanBlock
+
+        app_blocks.append(RamanBlock)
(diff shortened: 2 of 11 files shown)


be-smith:
Still have a dependency on bson:

Seems to come from importing LOGGER in 

/datalab-app-plugin-insitu/.venv/lib/python3.11/site-packages/pydatalab/logger.py


Traceback (most recent call last):
  File "/Users/ben/datalab-app-plugin-insitu/src/datalab_app_plugin_insitu/testing_process_data.py", line 4, in <module>
    from pydatalab.apps.uvvis import UVVisBlock
  File "/Users/ben/datalab-app-plugin-insitu/.venv/lib/python3.11/site-packages/pydatalab/apps/__init__.py", line 12, in <module>
    from pydatalab.blocks import COMMON_BLOCKS
  File "/Users/ben/datalab-app-plugin-insitu/.venv/lib/python3.11/site-packages/pydatalab/blocks/__init__.py", line 2, in <module>
    from pydatalab.blocks.base import DataBlock
  File "/Users/ben/datalab-app-plugin-insitu/.venv/lib/python3.11/site-packages/pydatalab/blocks/base.py", line 5, in <module>
    from pydatalab.logger import LOGGER
  File "/Users/ben/datalab-app-plugin-insitu/.venv/lib/python3.11/site-packages/pydatalab/logger.py", line 102, in <module>
    LOGGER = setup_log()
             ^^^^^^^^^^^
  File "/Users/ben/datalab-app-plugin-insitu/.venv/lib/python3.11/site-packages/pydatalab/logger.py", line 58, in setup_log
    from pydatalab.config import CONFIG
  File "/Users/ben/datalab-app-plugin-insitu/.venv/lib/python3.11/site-packages/pydatalab/config.py", line 19, in <module>
    from pydatalab.models import Person
  File "/Users/ben/datalab-app-plugin-insitu/.venv/lib/python3.11/site-packages/pydatalab/models/__init__.py", line 5, in <module>
    from pydatalab.models.cells import Cell
  File "/Users/ben/datalab-app-plugin-insitu/.venv/lib/python3.11/site-packages/pydatalab/models/cells.py", line 6, in <module>
    from pydatalab.models.entries import EntryReference
  File "/Users/ben/datalab-app-plugin-insitu/.venv/lib/python3.11/site-packages/pydatalab/models/entries.py", line 6, in <module>
    from pydatalab.models.relationships import TypedRelationship
  File "/Users/ben/datalab-app-plugin-insitu/.venv/lib/python3.11/site-packages/pydatalab/models/relationships.py", line 6, in <module>
    from pydatalab.models.utils import (
  File "/Users/ben/datalab-app-plugin-insitu/.venv/lib/python3.11/site-packages/pydatalab/models/utils.py", line 9, in <module>
    from bson.objectid import ObjectId
ModuleNotFoundError: No module named 'bson'

be-smith:
Also have similar issues still with importing 

`get_file_info_by_id`

Traceback (most recent call last):
  File "/Users/ben/datalab-app-plugin-insitu/src/datalab_app_plugin_insitu/testing_process_data.py", line 11, in <module>
    from pydatalab.apps.uvvis import UVVisBlock
  File "/Users/ben/datalab-app-plugin-insitu/.venv/lib/python3.11/site-packages/pydatalab/apps/uvvis/__init__.py", line 11, in <module>
    from pydatalab.file_utils import get_file_info_by_id
  File "/Users/ben/datalab-app-plugin-insitu/.venv/lib/python3.11/site-packages/pydatalab/file_utils.py", line 11, in <module>
    from werkzeug.datastructures import FileStorage
ModuleNotFoundError: No module named 'werkzeug'

Could maybe move this so you only import it when you use it so can still access the methods etc of the UVVisBlock

ml-evs:
> Also have similar issues still with importing
> 
> `get_file_info_by_id`
> 
> Traceback (most recent call last): File "/Users/ben/datalab-app-plugin-insitu/src/datalab_app_plugin_insitu/testing_process_data.py", line 11, in from pydatalab.apps.uvvis import UVVisBlock File "/Users/ben/datalab-app-plugin-insitu/.venv/lib/python3.11/site-packages/pydatalab/apps/uvvis/**init**.py", line 11, in from pydatalab.file_utils import get_file_info_by_id File "/Users/ben/datalab-app-plugin-insitu/.venv/lib/python3.11/site-packages/pydatalab/file_utils.py", line 11, in from werkzeug.datastructures import FileStorage ModuleNotFoundError: No module named 'werkzeug'
> 
> Could maybe move this so you only import it when you use it so can still access the methods etc of the UVVisBlock

Oh right, its being triggered by you importing the UVVisBlock, I thought this was just generally triggered when using the in situ block. In that case, the uv vis block does indeed need those dependencies to be installed and I don't think its worth refactoring too much around it (we'd have to change every block to solve the general problem). So for the sake of your development I would just install the datalab apps extra too `uv pip install '.[apps,server]'` from your datalab repo with your in situ environment activated.

In the NMR/echem case, it looks like we're just using the utilities developed for NMR/echem rather than the blocks themselves. Essentially we need to try to have a base block that can enable the child blocks to be agnostic to the database unless they really need to use it, and even then have that be an optional route.

ml-evs:
At least now this PR means you don't need all the extra app dependencies installed (langchain etc) -- might be able to get away with just adding pymongo and letting `get_file_info_by_id` fail silently if used without a database.

ml-evs:
Still not 100% happy with this approach (its a bit nasty to add new blocks, though this will become increasingly rare now we have plugins), but going to merge for now to stop holding up plugin development.

PR #1189, opened 2025-05-16
Fix DataTable select-all when sorting is applied

### LINKED ISSUE #1186: Data table select-all does not work when sort has been applied
To reproduce: sort data table by ID (or whatever) and try to select-all -- it will say 20 are selected, but those 20 are actually on a different page. Not sure if this has been introduced by recent changes, or has been around for a while. Probably one for @BenjaminCharmes?

BenjaminCharmes:
Closes #1186

### DIFF
--- a/webapp/src/components/DynamicDataTable.vue
+++ b/webapp/src/components/DynamicDataTable.vue
@@ -529,6 +529,10 @@ export default {
           }
         });
       });
+
+      this.$nextTick(() => {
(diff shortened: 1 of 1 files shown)


ml-evs (approved):
Works nicely, thakns @BenjaminCharmes! It would be nice if primevue triggered these things automatically on `$nextTick` when needed, but perhaps that is wishful thinking!

PR #1195, opened 2025-05-18
Set DataTable loading state when loading from store

### LINKED ISSUE #1194: DataTable never visually shows loading state
For a particularly slow call to `/samples` (e.g., an admin who can see everything), the table defaults to the `#empty` template rather than `#loading`. It seems like we need to manually control the loading state to trigger this; seems simple enough with `:loading = "data === null"`, so will make a PR.

ml-evs:
Closes #1194.

@BenjaminCharmes could you take a look at rolling this out across all item types? I don't think we necessarily need to make everything async for this work, but it should be straightforward to default to `null` in the store when things are loading and `[]` when they are actually empty.

As a side issue, the visual template for loading isn't particularly nice (grey overlay across entire component, with white text... perhaps this could be improved here too).

### DIFF
--- a/webapp/src/components/DynamicDataTable.vue
+++ b/webapp/src/components/DynamicDataTable.vue
@@ -24,6 +24,7 @@
       sort-mode="multiple"
       state-storage="local"
       :state-key="`datatable-state-${dataType}`"
+      :loading="data === null"
       @state-restore="onStateRestore"
       @state-save="onStateSave"
       @filter="onFilter"
@@ -56,8 +57,8 @@
           @reset-table="handleResetTable"
         />
       </template>
-      <template #empty> No data found. </template>
       <template #loading> Loading data. Please wait. </template>
+      <template #empty> No entries found. </template>
 
       <Column v-if="showButtons" class="checkbox" selection-mode="multiple"></Column>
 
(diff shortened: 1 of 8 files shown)


webapp/src/components/DynamicDataTable.vue
```diff
@@ -56,8 +57,20 @@
...
+      <template #loading>
+        <div class="card text-center">
+          <div class="card-body">
+            <font-awesome-icon
+              :icon="['fa', 'sync']"
+              class="fa-2x text-primary mb-2"
+              :spin="true"
+              aria-label="loading"
+            />
+            <p class="mb-0 font-weight-medium">Loading data. Please wait...</p>
```
ml-evs, line 69:
```suggestion
            <p class="mb-0 font-weight-medium">Loading entries, please wait...</p>
```

webapp/src/components/SampleTable.vue
```diff
@@ -88,8 +91,8 @@ export default {
     this.getSamples();
   },
   methods: {
-    getSamples() {
-      getSampleList();
+    async getSamples() {
+      await getSampleList();
```
ml-evs, line 95:
```suggestion
    getSamples() {
      getSampleList();
```

I think we can remove this now

ml-evs (commented):
Works nicely, thanks @BenjaminCharmes! Just going to make these two tweaks and then merge.

PR #1200, opened 2025-05-19
Improvements to docs rendering for models and datablocks

ml-evs:
Improves docs formatting and removes now-defunct notes, also updates to the latest mkdocs/mkdocs-materials/mkdocstrings, which should have better pydantic support going forward.


ml-evs:
Example before: 

image


And after:

image

Issue #610, opened 2024-02-21
Add JCAMP block


ml-evs:
JCAMP would be another "simple" format to support for plotting, though I'm not sure if anyone in the Grey group uses it currently.

PR #1205, opened 2025-05-20
Add support for JCAMP-DX files in NMR block

ml-evs:
This PR enables JCAMP-DX (`.jdx`, `.dx`) files to be read by the NMR block. It also updates the documentation for the NMR block slightly, and refactors the block into combined functions for loading either Bruker/JCAMP data.

Related to #610 -- we could easily add support for NMR given nmrglue's existing functionality for JCAMP-DX, but a general JCAMP block might be too complicated.


BenjaminCharmes (approved):
Everything working well! 

Two small comments:

1. Always seeing "NMRBlock raised warning: JCAMP-DX key without value: $RELAX" when using JCAMP-DC example_data

2. There is some measurement details missing when using JCAMP-DC Files

ml-evs:
> Everything working well! 

Thanks!

> 1. Always seeing "NMRBlock raised warning: JCAMP-DX key without value: $RELAX" when using JCAMP-DC example_data

Yeah, this is an nmrglue warning, I think it's probably something slightly off specification with the example data. 

> 2. There is some measurement details missing when using JCAMP-DC Files

Yup, I've extracted all the things that are required to be in the file, but we could also try to reverse engineer some of the custom extension keys `$xyz` that e.g. bruker save into these files.

PR #1207, opened 2025-05-20
Delete large keys (e.g., bokeh_plot_data) when sending and before receiving updated block data

### LINKED ISSUE #1206: Incorrect plot shown when switching between "good" and "bad" file
Currently bokeh plot data seems to be cached in the store when going from a block operating on a good file and switching that block to a file that is unable to be plotted. Simple fix: always invalidated bokeh_plot_data when updating a block from the server.

ml-evs:
Closes #1206 by zeroing the store data for plots when receiving a new version (in case the new version does not contain a plot).

Likewise, drops bokeh plot data (and other large fields) when sending information back to the API about the block (should be significantly faster), as these are not used anywhere at the moment. In future, this should be standardized by properly nesting block fields rather than polluting the top-level namespace.

### DIFF
--- a/webapp/src/server_fetch_utils.js
+++ b/webapp/src/server_fetch_utils.js
@@ -485,8 +485,14 @@ export async function getCollectionData(collection_id) {
 }
 
 export async function updateBlockFromServer(item_id, block_id, block_data, saveToDatabase = true) {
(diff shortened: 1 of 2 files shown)


BenjaminCharmes (approved):
LGTM

PR #1209, opened 2025-05-20
Use `tool.uv.sources` to specify in situ plugin git dependency in pyproject

### LINKED ISSUE #1208: Cannot release to PyPI with direct git dependency on in situ plugin
Turns out PyPI doesn't allow git dependencies (should have guessed). I think are options are the following:

- release the plugin to PyPI (doable soon, probably), should be done if `insitu-nmr` remains a "core plugin"
- remove the explicit dependency here -- it will still be used by servers if it is found to be installed, but will potentially need a new local mechanism for installing it (e.g., a requirements file like `plugins.txt` or something)

ml-evs:
Closes #1208.

Also fixes the minimum setuptools version, and properly passes the README to the PyPI build process.

### DIFF
--- a/pydatalab/pyproject.toml
+++ b/pydatalab/pyproject.toml
@@ -35,7 +35,7 @@ documentation = "https://docs.datalab-org.io"
 changelog = "https://github.com/datalab-org/datalab/releases"
 
 [build-system]
-requires = ["setuptools >= 62", "setuptools_scm ~= 8.1", "wheel"]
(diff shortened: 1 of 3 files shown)


pydatalab/pyproject.toml
```diff
@@ -72,9 +73,13 @@ apps = [
     "rosettasciio ~= 0.3, < 0.4",
     # TGA
     "python-dateutil ~= 2.9",
+]
+
+insitu = [
```
ml-evs, line 78:
```suggestion
app-plugins-git = [
```

PR #1213, opened 2025-05-22
Limit batch interface to 100 items in UI and 10,000 via API

### LINKED ISSUE #1070: Limit bad interface to reasonable number (100?)
We should cap the batch interface out at some limit to avoid locking up the UI.

BenjaminCharmes:
Closes #1070 

window.alert() will be replaced with new composant from #1212

### DIFF
--- a/pydatalab/src/pydatalab/config.py
+++ b/pydatalab/src/pydatalab/config.py
@@ -240,7 +240,7 @@ its importance when deploying a datalab instance.""",
     )
 
     MAX_BATCH_CREATE_SIZE: int = Field(
-        100, description="Maximum number of items that can be created in a single batch operation."
+        10_000, description="Maximum number of items that can be created in a single batch operation."
     )
 
     BACKUP_STRATEGIES: dict[str, BackupStrategy] | None = Field(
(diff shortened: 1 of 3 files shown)


pydatalab/src/pydatalab/config.py
```diff
@@ -239,6 +239,10 @@ class ServerConfig(BaseSettings):
 its importance when deploying a datalab instance.""",
     )
 
+    MAX_BATCH_CREATE_SIZE: int = Field(
+        100, description="Maximum number of items that can be created in a single batch operation."
```
ml-evs, line 243: From the API side we can be much more lax, e.g., 10,000... just in case that is necessary. The main problem is that the UI is too slow for it atm, but it is a good idea to add an upper limit for the API anyway.

```suggestion
        10_000, description="Maximum number of items that can be created in a single batch operation."
```

ml-evs (changes requested):
Thanks for having a go at this @BenjaminCharmes!

I think the overall problem is more that once you type in 1000, the UI makes 1000 rows which is mega slow, so we need the validation to happen before that step (apologies if this code does indeed do that, I can't test locally right now). I think due to the way we're binding the value from the `input`, it ignores the already set `max`, so it might just be a case of preventing the rows from being rendered until the number in the box is valid.

ml-evs (approved):
Works well, thanks @BenjaminCharmes!

PR #1220, opened 2025-05-24
Add persistent per-item access tokens that can be used in QR codes

### LINKED ISSUE #1053: Give read-only access to any QR code links
QR code links should be accessible without login to datalab, with read-only access. Probably will need the following approach:

- [ ] datalab API stores a new kind of secret key to authenticated users that is infrequently rotated (perhaps just on server startup), with each one stored in the database.
- [ ] when generating QR code links, the app gets the latest secret key and encodes it as a token in the generated QR code link, along with the refcode and user in question.
- [ ] When visiting the link, the API gives access based on whether the token was once generated by datalab for that entry.

This is a bit insecure; a logged-in user as the key will be visible in network requests and could craft links for other entries, but perhaps this is functionality we want anyway wrt magic links! The only difference will be that the QRCode will be speicifcally encoded as a JWT with no baked-in expiry date.

Another issue is that JWTs might be too large to stuff into QR codes

cc @PeterKraus

ml-evs:
This PR initiates the work on allowing QR codes to be generated that are
accessible by anyone who has printed one out, requested by @PeterKraus in #1053.

Closes #1053.

It does this by adding a `/items/<refcode>/issue-access-token` endpoint, which stores a token in the
database that can be used to access that refcode without login. A corresponding
admin endpoint for invalidating the token is added, in case of nefarious use.

Outstanding questions:

- [x] ~For the UI, this should likely be a conscious second step -- e.g., QR code -> generate public QR code. We may want to limit how many there are per item too, perhaps to 1. Admins should probably also have access to another UI that shows when tokens were generated, with the option to invalidate them, though I think this could be a future PR.~ Completed, including admin UI
- [x] ~Right now, a user with existing access to a sample could generate access tokens to all samples they have access to without using the UI. This is a bit of a harder problem to solve, and involves some trust on existing verified users. For now, the token is stored with the user account in place, so tokens can be invalidated programmatically that correspond to that user. We may also want to investigate a way of sharing a secret between the app and API so that the API can be sure that the generation is being done from a known UI, though this would probably be a relatively complicated deployment detail to implement.~ Shelved for now, given that admins can revoke if need be
- [x] ~As it stands in this PR, the user still has to be logged in to get access to the item (though they don't necessarily need to have permissions for it). I will relax this constraint once the rest is properly tested.~ Unauthenticated users can now access, and the federation has been updated accordingly
- [x] ~I am currently using `uuid.uuid1()` for the token generation (and storing a sha-512 hash of it in the database), but we might need to reduce the entropy a bit so that it fits into a printable QR.~ QRs seem eminently printable

These questions were answered by @BenjaminCharmes's follow up work. To summarise:

- unlimited public QR tokens per item
- hashed token stored alongside who created it and when
- admins can revoke any token through a new UI
- the token gives access to metadata but ignores blocks and files
- has to be regenerated every time "new public token" is made

Still to-do:

- [ ] potentially send an email to the admin every token that gets created, to avoid misuse (as described above)

### DIFF
--- a/pydatalab/src/pydatalab/routes/v0_1/items.py
+++ b/pydatalab/src/pydatalab/routes/v0_1/items.py
@@ -23,6 +23,7 @@ from pydatalab.models.utils import generate_unique_refcode
 from pydatalab.mongo import ITEMS_FTS_FIELDS, flask_mongo
 from pydatalab.permissions import (
     PUBLIC_USER_ID,
+    access_token_or_active_users,
     active_users_or_get_only,
     check_access_token,
     get_default_permissions,
@@ -793,11 +794,11 @@ def update_item_permissions(refcode: str):
     return jsonify({"status": "success"}), 200
 
 
-@ITEMS.route("/items/<refcode>/issue-access-token", methods=["GET"])
+@ITEMS.route("/items/<refcode>/issue-access-token", methods=["POST"])
 def issue_physical_token(refcode: str):
     """Issue a token that will give semi-permanent access to an
     item with this refcode. This should be used when generating
-    physicsl labels to attach to a container.
+    physical labels to attach to a container.
 
     """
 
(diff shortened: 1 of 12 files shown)


BenjaminCharmes:
> Looking really great @BenjaminCharmes, just a couple of API comments to add to the other work. Can you also double-check the failing test?

@ml-evs Everything should work now! Let me know if something isn't right.

pydatalab/src/pydatalab/routes/v0_1/items.py
```diff
@@ -786,16 +794,66 @@ def update_item_permissions(refcode: str):
...
+        ), 404
+
+    token = str(uuid.uuid1())
+    access_document = {
+        "token": sha512(token.encode("utf-8")).hexdigest(),
+        "refcode": refcode,
+        "user": ObjectId(current_user.id),
+        "active": True,
+        "created_at": datetime.datetime.now(tz=datetime.timezone.utc),
+        "type": "access_token",
```
ml-evs:
```suggestion
        "version": 1,
        "type": "access_token",
```

Final tweak I want, set a simple integer version of the token scheme in case we change this in future but want to use old tokens.

We could probably also wrap this up into a pydantic model at some point, but not necessary now.

PR #1222, opened 2025-05-30
Add spectrum selection controls to NMR insitu datablock

BenjaminCharmes:
**Dependencies:**
* Required for: [datalab-app-plugin-insitu PR#38](https://github.com/datalab-org/datalab-app-plugin-insitu/pull/38)

**Features:**
* Added user controls to select which NMR experiments to process and visualize:
  - Start/end experiment range selection  
  - Step parameter for sampling
  - Input validation for parameter combinations

### DIFF
--- a/webapp/src/components/datablocks/NMRInsituBlock.vue
+++ b/webapp/src/components/datablocks/NMRInsituBlock.vue
@@ -159,7 +159,7 @@ export default {
     hasParameterChanges() {
       const currentStartExp = this.start_exp || 1;
       const currentEndExp = this.end_exp || this.maxExperiments;
-      const currentStepExp = this.step_exp || 1;
+      const currentStepExp = this.step_exp || null;
       const currentExcludeExp = this.exclude_exp || "";
 
       const hasChanges =


be-smith:
One initial comment - currently when I updated one value for selecting the spectra it starts recalculating straight away. This could take a large amount of time. Worth discussing but maybe having an apply button once you have filled in all three values for selecting exp might be better/more intuitive as in #1246.

I am also getting an error when trying to save the changes to the sample (I increased the starting spectrum range to 300). Saving worked fine when I left it at 1 and changed the step value instead.

"{"message":"(\"'update' command document too large\",)","title":"DocumentTooLarge"}"

Other things I noticed - time starts at 0 as does the scan number even when I select a scan many hours into the experiment - I guess this depends on what we want this functionality to do, if it's just for getting rid of a few scans at the begining or end of the experiment that's fine - but I would expect the echem trace to still show the correct time.

Similarly when changing the step parameter it doesn't show the "actual" experiment number in relation to the files - just the number of the spectra that is plotted (i.e I would expect it to count up in 5s if the step value is 5).

Here is a picture of what I mean by the time going back to zero, and the scans not being labelled consistently with the files.

Screenshot 2025-07-08 at 17 37 12


saveItem Called! [server_fetch_utils.js:554:1](webpack://datalab-vue/src/server_fetch_utils.js?417a)
Item data before saving: 
Proxy { <target>: {…}, <handler>: {…} }
[server_fetch_utils.js:556:1](webpack://datalab-vue/src/server_fetch_utils.js?417a)
XHRPOST
http://localhost:5001/save-item/
[HTTP/1.1 500 INTERNAL SERVER ERROR 1768ms]

fetch was successful [server_fetch_utils.js:84:1](webpack://datalab-vue/src/server_fetch_utils.js?417a)
Response { type: "cors", url: "http://localhost:5001/save-item/", redirected: false, status: 500, ok: false, statusText: "INTERNAL SERVER ERROR", headers: Headers(2), body: ReadableStream, bodyUsed: true }
[server_fetch_utils.js:85:1](webpack://datalab-vue/src/server_fetch_utils.js?417a)
{"message":"(\"'update' command document too large\",)","title":"DocumentTooLarge"}
[server_fetch_utils.js:86:1](webpack://datalab-vue/src/server_fetch_utils.js?417a)

webapp/src/components/datablocks/NMRInsituBlock.vue
```diff
@@ -98,10 +156,82 @@ export default {
     availableFolders() {
       return this.currentBlock.available_folders || [];
     },
+    hasParameterChanges() {
+      const currentStartExp = this.start_exp || 1;
+      const currentEndExp = this.end_exp || this.maxExperiments;
+      const currentStepExp = this.step_exp || 1;
```
ml-evs, line 162:
```suggestion
      const currentStepExp = this.step_exp || null;
```

ml-evs (approved):
Thanks @BenjaminCharmes!

PR #1225, opened 2025-06-03
Update the collection page table with the latest dynamic data table options

### LINKED ISSUE #1192: Update the collection page table with the latest dynamic data table options
The table of collection entries does not currently use any of the nice new column stuff, nor does it have searching or item selection. Hopefully these are straightforward to enable; but perhaps we could have one global column list for all collections rather than storing table settings per collection. The selected items dialogue could also do more things than just delete, e.g., "Remove from collection" rather than "add to collection", but for now this could just be omitted.

Related to #1190; the collection page could do with a bit of a facelift generally.

BenjaminCharmes:
Closes #1192

### DIFF
--- a/pydatalab/src/pydatalab/routes/v0_1/collections.py
+++ b/pydatalab/src/pydatalab/routes/v0_1/collections.py
@@ -419,3 +419,36 @@ def add_items_to_collection(collection_id):
         )
 
     return (jsonify({"status": "success"}), 200)
+
+
+@COLLECTIONS.route("/collections/<collection_id>/items", methods=["DELETE"])
+def remove_items_from_collection(collection_id):
+    data = request.get_json()
+    item_ids = data.get("item_ids", [])
+
+    collection = flask_mongo.db.collections.find_one(
+        {"collection_id": collection_id, **get_default_permissions()}
+    )
+
+    if not collection:
+        return jsonify({"error": "Collection not found"}), 404
+
+    if not item_ids:
+        return jsonify({"error": "No item IDs provided"}), 400
+
+    update_result = flask_mongo.db.items.update_many(
+        {"item_id": {"$in": item_ids}, **get_default_permissions()},
+        {
+            "$pull": {
+                "relationships": {
+                    "immutable_id": ObjectId(collection["_id"]),
+                    "type": "collections",
+                }
+            }
+        },
+    )
+
+    if update_result.matched_count == 0:
+        return jsonify({"status": "error", "message": "No matching items found."}), 404
+
+    return jsonify({"status": "success", "removed_count": update_result.modified_count}), 200
--- a/webapp/src/components/CollectionInformation.vue
+++ b/webapp/src/components/CollectionInformation.vue
@@ -36,18 +36,14 @@
       </div>
     </div>
     <DynamicDataTable
+      :key="tableKey"
       :data="children"
       :columns="collectionTableColumns"
-      :data-type="'samples'"
-      :global-filter-fields="[
-        'item_id',
-        'name',
-        'refcode',
-        'chemform',
-        'blocks',
-        'characteristic_chemical_formula',
-      ]"
-      :show-buttons="false"
+      :data-type="'collectionItems'"
+      :global-filter-fields="['item_id', 'name']"
+      :show-buttons="true"
+      :collection-id="collection_id"
+      @remove-selected-items-from-collection="handleItemsRemovedFromCollection"
     />
   </div>
 </template>
@@ -75,6 +71,7 @@ export default {
   },
   data() {
     return {
+      tableKey: 0,
       collectionTableColumns: [
         {
           field: "item_id",
@@ -114,6 +111,13 @@ export default {
         .catch(() => {
           this.fetchError = true;
         });
+    },
+    handleItemsRemovedFromCollection() {
+      this.getCollectionChildren();
+
+      this.$nextTick(() => {
+        this.tableKey += 1;
+      });
     },
   },
 };
--- a/webapp/src/components/DynamicDataTable.vue
+++ b/webapp/src/components/DynamicDataTable.vue
@@ -45,6 +45,7 @@
           :show-buttons="showButtons"
           :available-columns="availableColumns"
           :selected-columns="selectedColumns"
+          :collection-id="collectionId"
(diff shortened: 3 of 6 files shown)


pydatalab/src/pydatalab/routes/v0_1/collections.py
```diff
@@ -419,3 +419,36 @@ def add_items_to_collection(collection_id):
         )
 
     return (jsonify({"status": "success"}), 200)
+
+
+@COLLECTIONS.route("/collections/<collection_id>/items", methods=["DELETE"])
+def remove_items_from_collection(collection_id):
```
ml-evs, line 425: It would be good to have a quick test of this endpoint! I think until now, I've been using `/save-item` to remove an item from a collection, but this is not very elegant and handling it this way would be much better. Could we also use this endpoint on the item page when removing a collection from the list of collections that a sample is in?

ml-evs (commented):
Thanks @BenjaminCharmes, apologies for the delay... I'd left this as a pending review comment.

pydatalab/src/pydatalab/routes/v0_1/collections.py
```diff
@@ -419,3 +419,36 @@ def add_items_to_collection(collection_id):
...
+                "relationships": {
+                    "immutable_id": ObjectId(collection["_id"]),
+                    "type": "collections",
+                }
+            }
+        },
+    )
+
+    if update_result.matched_count == 0:
+        return jsonify({"status": "error", "message": "No matching items found."}), 404
```
ml-evs, line 452: I think we probably need to also differentiate the case where some items were found with a 207 mixed response

```suggestion
    if update_result.matched_count == 0:
        return jsonify({"status": "error", "message": "No matching items found."}), 404
        
    elif update_result.matched_count != len(item_ids):
        return jsonify({"status": "partial-success", "message": "Only n items updated"}), 207
```

pydatalab/src/pydatalab/routes/v0_1/collections.py
```diff
@@ -419,3 +419,36 @@ def add_items_to_collection(collection_id):
...
+    )
+
+    if not collection:
+        return jsonify({"error": "Collection not found"}), 404
+
+    if not item_ids:
+        return jsonify({"error": "No item IDs provided"}), 400
+
+    update_result = flask_mongo.db.items.update_many(
+        {"item_id": {"$in": item_ids}, **get_default_permissions()},
```
ml-evs, line 440: Maybe should use refcodes rather than item ids to be a bit safer?

```suggestion
        {"refcode": {"$in": refcodes}, **get_default_permissions()},
```

pydatalab/src/pydatalab/routes/v0_1/collections.py
```diff
@@ -419,3 +419,36 @@ def add_items_to_collection(collection_id):
...
     return (jsonify({"status": "success"}), 200)
+
+
+@COLLECTIONS.route("/collections/<collection_id>/items", methods=["DELETE"])
+def remove_items_from_collection(collection_id):
+    data = request.get_json()
+    item_ids = data.get("item_ids", [])
+
+    collection = flask_mongo.db.collections.find_one(
+        {"collection_id": collection_id, **get_default_permissions()}
```
ml-evs, line 430:
```suggestion
        {"collection_id": collection_id, **get_default_permissions()}, project={"_id": 1}
```

Can just project the underlying ID to avoid returning a load of extra info from the db

pydatalab/tests/server/test_samples.py
```diff
@@ -750,3 +750,174 @@ def test_add_items_to_collection_success(client, default_collection, example_ite
     child_refcodes = [item["refcode"] for item in collection_data["child_items"]]
 
     assert all(refcode in child_refcodes for refcode in refcodes)
+
+
+@pytest.mark.dependency()
```
ml-evs, line 755:
```suggestion
```

I would double check, but if these tests really are isolated then you can remove all the dependency markers -- I would just make extra sure they don't have any side effects, i.e., stuff left in the database after the test (which is non-trivial), otherwise you can mark them as having a linear dependency chain by listing the function name above it in this decorator, should at least be deterministic in that case

webapp/src/components/CollectionInformation.vue
```diff
@@ -38,16 +38,11 @@
...
-        'item_id',
-        'name',
-        'refcode',
-        'chemform',
-        'blocks',
-        'characteristic_chemical_formula',
-      ]"
-      :show-buttons="false"
+      :data-type="'collectionItems'"
+      :global-filter-fields="['item_id', 'name']"
```
ml-evs, line 42:
```suggestion
      :global-filter-fields="['item_id', 'name', 'refcode', 'blocks', 'chemform', 'characteristic_chemical_formula']"
```

May as well keep these if they are being shown in the table anyway?

ml-evs (commented):
UI seems to work nicely, thanks @BenjaminCharmes! Just a couple of comments about the API endpoint and tests

pydatalab/tests/server/test_samples.py
BenjaminCharmes: So I've put it back for now. Maybe I'm doing something wrong ?

be-smith (dismissed):
UI changes look great - and all the functionality seems to be working for me!

pydatalab/tests/server/test_samples.py
```diff
@@ -750,3 +750,186 @@ def test_add_items_to_collection_success(client, default_collection, example_ite
...
+    collection_dict["collection_id"] = "test_collection_empty_items"
+    response = client.put("/collections", json={"data": collection_dict})
+    assert response.status_code == 201
+
+    collection_id = collection_dict["collection_id"]
+    response = client.delete(f"/collections/{collection_id}/items", json={"refcodes": []})
+
+    assert response.status_code == 400
+    data = response.get_json()
+    assert data["error"] == "No refcodes provided"
```
ml-evs, line 854: Just a comment that this made me realise the collections routes do not use standard errors (i.e., status = "error", then "message" = "whatever"). Not something to fix in this PR, but we should go through and unify all routes on this at some point.

ml-evs (approved):
Looks good and works well, thanks @BenjaminCharmes!

Issue #1184, opened 2025-05-15
Database migrations


ml-evs:
I just wanted to open an issue to track this. We have a few bugs right now that have been plastered over to avoid the need for a robust database migration system. At some point we will want to be able to do this. In this issue, I'll memory-dump the things I know are semi-broken, and briefly suggest how we might handle this in the future (to be expanded upon in a future issue). I will refer to "documents" as those stored in the database, and "entries" as the datalab usage.

Known issues dating back to first public release of datalab:

| Title/field | Severity | Description | Remedy | Related issues/PRs |
|-------------|----------|-------------|--------|--------------------|
| Item `files` | 🟢  | Some documents contain out-of-date `files` storing file metadata, in addition to that stored in the `files` collection. | Delete any top-level `files` fields on an entry | #1055 
| `relationships` with mutable `item_id` | 🔴  | Previously, we would happily refer to an item by its `item_id`, before deciding that this should be mutable. There are therefore several `relationships` (and `synthesis_constituents` fields) that point to `item_id`s that could in theory change. This has been preventing us from allowing `item_id` to be truly editable for some time. | A migration task that carefully matches `item_id` to the real `refcode` and deprecates the relationship if not found. | #1173 #550 #323 
| `blocks` in their own collection | 🟡 | Blocks are still currently stored within a document under the `blocks_obj` key, which has unpleasant custom block IDs and imposes limits on their size. | Migrate all existing blocks to a new collection and replace all custom IDs with real uuids. | #48 

Quick note about the best ways of handling this generically: either offline scripts or server startup checks, plus a collection that tracks which migrations have been applied. Will also need to track which migrations depend on one another, though perhaps sensible just to enforce a linear chain and ensure that the migrations themselves are idempotent.

davidwaroquiers:
Just a small comment here. We've done something similar for jobflow-remote (jfr) with an update procedure. If there is any backward incompatible change between one version of jfr and another, there is a list of updates that are applied one after another to make sure the database is consistent and compatible with the version of jfr used. Might want to take some ideas from there.

Link to the documentation: https://matgenix.github.io/jobflow-remote/user/install.html#upgrade
Relevant part of the code with the DatabaseUpgrader: https://github.com/Matgenix/jobflow-remote/blob/develop/src/jobflow_remote/jobs/upgrade.py#L92

PR #1227, opened 2025-06-04
Add relationships by `refcode` rather than `item_id` from the UI

### LINKED ISSUE #1173: Synthesis constituents should be added from UI by refcode rather than `item_id`
All of the models were adjusted to handle this case, but it seems like the UI does not save the refcode of an item when it is returned from the search (and thus the database only stores item_id -- bad as these are technically mutable). We need to make sure that the synthesis table prefers refcode over any other identifier.

BenjaminCharmes:
Closes #1173

Stock item_id and refcode in the db, but refcode has priority.


ml-evs (approved):
Works nicely, thanks @BenjaminCharmes! Will need to consider a migration for all existing relationships at some point (c.f. #1184), but perhaps a bit further down the line...

PR #1228, opened 2025-06-04
Display NotImplementedBlock when an unknown block type is encountered

### LINKED ISSUE #1159: Only emit warnings when an unknown block type is encountered

BenjaminCharmes:
Closes #1159 

Could also add an alert from #1212 when an unknow block type is ancountered

### DIFF
--- a/webapp/src/components/datablocks/BokehBlock.vue
+++ b/webapp/src/components/datablocks/BokehBlock.vue
@@ -6,7 +6,7 @@ DataBlockBase as a prop, and save from within DataBlockBase  -->
       v-model="file_id"
       :item_id="item_id"
       :block_id="block_id"
-      :extensions="blockInfo?.attributes.accepted_file_extensions"
+      :extensions="blockInfo?.attributes?.accepted_file_extensions"
       update-block-on-change
     />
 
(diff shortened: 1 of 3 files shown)


be-smith:
I'm not sure what the desired behaviour is, when loading a sample with a block not defined in resources.js it loads a default bokehblock and doesn't throw any errors.

Weirdly to test this when I removed the blocks from resources.js they still appeared in the dropdown menu as available blocks, even when they simply defaulted to a bokeh block after I created them.

ml-evs (approved):
> I'm not sure what the desired behaviour is, when loading a sample with a block not defined in resources.js it loads a default bokehblock and doesn't throw any errors.
> 
> Weirdly to test this when I removed the blocks from resources.js they still appeared in the dropdown menu as available blocks, even when they simply defaulted to a bokeh block after I created them.

`resources.js` is now only used as an override for custom components, the web app will try to show any block defined in `/info/blocks` so to test out this PR you have to make a block that is not defined there.

My previous problem seems to have gone away and this now works nicely for me, so I'm going to merge! Thanks @BenjaminCharmes!

PR #1242, opened 2025-06-17
Add exhaustive block API testing for all block types

### LINKED ISSUE #1191: Need much more exhaustive block API testing
In #1188, only one test failed when I completely broke how blocks were being imported:

https://github.com/datalab-org/datalab/blob/c98cda00dd70de9a79eac584f772ed6910b88da9/pydatalab/tests/server/test_starting_materials.py#L138-L180

We need to add exhaustive tests of the block API in a new module `tests/server/test_blocks.py` that, in the first instance, loops through all available block types and tries to create a sample with that block via the API. This will supplement the more scientific "apps" tests in `tests/apps` and the webapp e2e tests, which themselves are also non-exhaustive atm.

I'd suggest its worth doing this before #1188 is merged.

BenjaminCharmes:
Closes #1191 

Add tests for the block API in a new test module `tests/server/test_blocks.py`.

### DIFF
--- a/pydatalab/tests/server/test_blocks.py
+++ b/pydatalab/tests/server/test_blocks.py
@@ -12,7 +12,9 @@ def test_get_all_available_block_types():
     for block_class in BLOCKS:
         assert block_class.blocktype is not None
         assert block_class.name is not None
-        assert block_class.description is not None, f"{block_class.blocktype} is missing a description"
+        assert (
+            block_class.description is not None
+        ), f"{block_class.blocktype} is missing a description"
         assert block_class.blocktype in BLOCK_TYPES
         assert BLOCK_TYPES[block_class.blocktype] == block_class
 
@@ -119,16 +121,18 @@ def test_add_multiple_blocks_to_sample(admin_client, default_sample_dict):
         assert block_type in added_block_types
 
 
-def test_block_permissions(client, admin_client, default_sample_dict):
-    """Test that normal users can add blocks to samples they have access to."""
+def test_block_permissions(client, admin_client, unauthenticated_client, default_sample_dict):
+    """Test that normal users can add blocks to samples they have access to, but unauthenticated users cannot."""
     sample_id = "test_sample_user_permissions"
     sample_data = default_sample_dict.copy()
     sample_data["item_id"] = sample_id
 
-    response = admin_client.post("/new-sample/", json=sample_data)
+    # Create sample with normal user
+    response = client.post("/new-sample/", json=sample_data)
     assert response.status_code == 201
 
     block_type = list(BLOCK_TYPES.keys())[0]
+
     response = client.post(
         "/add-data-block/",
         json={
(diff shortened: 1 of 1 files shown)


pydatalab/tests/server/test_blocks.py
```diff
@@ -0,0 +1,180 @@
...
+@pytest.mark.dependency()
+def test_get_all_available_block_types():
+    """Test that we can enumerate all available block types."""
+    assert len(BLOCKS) > 0
+    assert len(BLOCK_TYPES) > 0
+
+    for block_class in BLOCKS:
+        assert hasattr(block_class, "blocktype")
+        assert hasattr(block_class, "name")
+        assert hasattr(block_class, "description")
```
ml-evs, line 15:
```suggestion
        assert block_class.blocktype is not None
        assert block_class.name is not None
        assert block_class.description is not None, f"{block_class.blocktype} is missing a description"
```

Marginal improvement; I think the hasattr will pass for anything that inherits from the base block by default, this checks that a description is properly provided.

pydatalab/tests/server/test_blocks.py
```diff
@@ -0,0 +1,180 @@
...
+            "block_type": block_type,
+            "item_id": sample_id,
+            "index": 0,
+        },
+    )
+
+    if response.status_code == 200:
+        assert response.json["status"] == "success"
+    elif response.status_code == 401:
+        pass
```
ml-evs, line 144: I think this test will always just pass through with 401 right? Could we instead test that the "normal" client can make a sample, add a block, then the admin can add another block, but an unauthenticated client cannot?

pydatalab/tests/server/test_blocks.py
```diff
@@ -0,0 +1,180 @@
...
+        assert BLOCK_TYPES[block_class.blocktype] == block_class
+
+
+@pytest.mark.dependency(depends=["test_get_all_available_block_types"])
+@pytest.mark.parametrize("block_type", list(BLOCK_TYPES.keys()))
+def test_create_sample_with_each_block_type(admin_client, block_type, default_sample_dict):
+    """Test creating a sample and adding each available block type via API."""
+    sample_id = f"test_sample_with_{block_type.replace('-', '_')}"
+    sample_data = default_sample_dict.copy()
+    sample_data["item_id"] = sample_id
```
ml-evs, line 26: I wonder if this test could be expanded to use actual files out of the box, we can make a "fat" default sample that has a load of example data uploaded to it, then try to make the blocks with real files. Could be pushed to another PR, as you prefer!

ml-evs (changes requested):
Looks good @BenjaminCharmes, couple of comments that can be addressed here or in a future PR.

ml-evs (approved):
Great work, thanks @BenjaminCharmes! I'll convert some of the final comments to sub-issues to address later on, but this is already super useful (and another 2.5% coverage bump!)

pydatalab/tests/server/test_blocks.py
ml-evs: #1247

PR #1243, opened 2025-06-20
Remove broken deployment.md link in INSTALL.md

yue-here:
deployment.md no longer exists, this removes the broken link in the first instance and changes the link to https://docs.datalab-org.io/en/stable/deployment/ in the second


ml-evs (commented):
This "internal" link actually works fine when rendered by the docs (https://docs.datalab-org.io/en/latest/INSTALL) -- I think I'm of the opinion that we should actually just move this file into the docs directly and link to it from the README, rather than maintaining it as a symlink. There's also lots of formatting that doesn't render properly on GitHub as its targeted at mkdocs, e.g., the tabbed `uv` vs `venv` instructions.

PR #1253, opened 2025-06-24
Remove browersify stream polyfill

ml-evs:
This PR removes the stream polyfill that is required by parts of uppy. As we are using XHR with Uppy, I don't think we need this (and its an annoying dependency to have).


BenjaminCharmes (dismissed):
Everything works perfectly locally for me 👍

ml-evs:
Also works for me, so will just remove it and see what happens in deployments (though pinging @jdbocarsly in case there was a specific reason to keep it)

PR #1259, opened 2025-06-27
Add test for sample with several example files attached to corresponding blocks

### LINKED ISSUE #1247: Create a test sample with several example characterisation files attached
Related to #1242 and #1191, it would be helpful to have a sample we can load in our tests that contains any simple test data for many block types; this would complement the e2e tests by checking that the Python block code can really be run e2e.

BenjaminCharmes:
Closes #1247


ml-evs (approved):
Thanks for this @BenjaminCharmes, +4% coverage :heart_eyes: 

I just added a commit that 

a) fixes the uv-vis test case, as the special selected file order was not being set
b) tests that the bokeh plot is actually created

Think this is now good to go once the tests pass!

Issue #1260, opened 2025-06-28
XRD block suggestions and feedback


ml-evs:
Some feedback from @MJCliffe about next steps for the XRD block. Some of these move more towards a dedicated XRD app (which wouldn't be a bad thing), which we could build out for certain techniques.

Plotting options:

- [ ] show point coordinates in cross-hair mode
- [ ] ability to hide points and only show line
- [ ] move legend outside plot by default (long file names ruin the layout)
- [ ] ability to stagger rather than overlap patterns
- [ ] make plot generally bigger
- [ ] ability to rebin data with fixed resolution (and compute background from that)

Many of these can be made as improvements across all blocks by enhancing the big general bokeh plot method

Longer term:

- [ ] automatic peak picking
- [ ] hkl on hover for theoretical patterns
- [ ] ability to mark a given CIF as the "real" one for a sample, rather than just comparison

PR #1263, opened 2025-07-01
Block plotting improvements: ability to hide points, larger plots, external legends

### LINKED ISSUE #1264: ability to hide points and only show line

### LINKED ISSUE #1265: make plot generally bigger

### LINKED ISSUE #1267: move legend outside plot by default (long file names ruin the layout)

BenjaminCharmes:
Related to #1260 

Closes #1264
Closes #1265 
Closes #1267

All these features (except "Ability to stagger rather than overlap patterns") are added to all blocks using BokehPlot. Should it be available only for XRD? Or should also consider adding the stagger option to other blocks as well?

About 'long file names ruin the layout' in #1267, I also set a 15-character limit for the legend.

Rebased + cherry picked a subset of the changes by @ml-evs (excludes #1262 and #1266 for now).

### DIFF
--- a/pydatalab/src/pydatalab/bokeh_plots.py
+++ b/pydatalab/src/pydatalab/bokeh_plots.py
@@ -25,6 +25,8 @@ from bokeh.plotting import ColumnDataSource, figure
 from bokeh.themes import Theme
 from scipy.signal import find_peaks
 
+from .utils import shrink_label
+
 FONTSIZE = "12pt"
 TYPEFACE = "Helvetica, sans-serif"
 COLORS = Dark2[8]
@@ -339,19 +341,14 @@ def selectable_axes_plot(
         if isinstance(df, dict):
             df_ = df[df_]
 
-        df_with_metadata = df_.copy()
-
         if labels:
             label = labels[ind]
         else:
             label = df_.index.name if len(df) > 1 else ""
 
-        if hasattr(df_, "attrs"):
-            for attr in ["item_id", "original_filename", "wavelength"]:
-                if attr in df_.attrs:
-                    df_with_metadata[attr] = df_.attrs[attr]
+        label = shrink_label(label)
 
-        source = ColumnDataSource(df_with_metadata)
+        source = ColumnDataSource(df_)
 
         if color_options:
             color = {"field": color_options[0], "transform": color_mapper}
@@ -392,7 +389,14 @@ def selectable_axes_plot(
         )
 
         lines = (
-            p.line(x=x_default, y=y_default, source=source, color=line_color, legend_label=label)
+            p.line(
+                x=x_default,
+                y=y_default,
+                source=source,
+                color=line_color,
+                legend_label=label,
+                line_width=2,
+            )
             if plot_line
             else None
         )
@@ -442,6 +446,21 @@ def selectable_axes_plot(
         p.legend.click_policy = "hide"
         if len(df) <= 1:
             p.legend.visible = False
+        else:
+            legend_items = p.legend.items
+            p.legend.visible = False
+
+            from bokeh.models import Legend
+
+            external_legend = Legend(
+                items=legend_items,
+                click_policy="hide",
+                background_fill_alpha=0.8,
+                label_text_font_size="9pt",
+                spacing=1,
+                margin=2,
+            )
+            p.add_layout(external_legend, "right")
 
     input_widgets = []
     if parameters:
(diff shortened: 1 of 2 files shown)


pydatalab/src/pydatalab/apps/xrd/blocks.py
```diff
@@ -23,7 +23,7 @@ class XRDBlock(DataBlock):
     description = "Visualize XRD patterns and perform simple baseline corrections."
     accepted_file_extensions = (".xrdml", ".xy", ".dat", ".xye", ".rasx", ".cif")
 
-    defaults = {"wavelength": 1.54060}
+    defaults = {"wavelength": 1.54060, "stagger_enabled": False, "stagger_offset": 1.0}
```
ml-evs, line 26:
```suggestion
    defaults = {"wavelength": 1.54060, "stagger_enabled": True, "stagger_offset": 1.0}
```

Let's go for it!

pydatalab/src/pydatalab/apps/xrd/blocks.py
```diff
@@ -261,6 +280,21 @@ def generate_xrd_plot(self) -> None:
...
+            if stagger_enabled and len(pattern_dfs) > 1:
+                for ind, df in enumerate(pattern_dfs):
+                    offset = ind * stagger_offset
+                    for col in df.columns:
+                        if "intensity" in col.lower() and df[col].dtype in [
+                            "float64",
+                            "float32",
+                            "int64",
+                            "int32",
+                        ]:
```
ml-evs, line 295: Can we make sure this column is always float64 rather than falling over if not?

pydatalab/src/pydatalab/bokeh_plots.py
```diff
@@ -274,6 +279,23 @@ def selectable_axes_plot(
...
+                if len(ext) < 6:
+                    available = 15 - len(ext) - 4
+                    if available > 3:
+                        label = f"{name[:available]}...{ext}"
+                    else:
+                        label = f"{label[:12]}..."
+                else:
+                    label = f"{label[:12]}..."
+            else:
+                label = f"{label[:12]}..."
```
ml-evs, line 297: Let's isolate this as a little function, either inlined or as a separate utility:

```suggestion
        if label is None:
            label = ""
            
        label = shrink_label(label)
            
        def shrink_label(label: str | None, max_length: int = 15):

            if label and len(label) > 15:
                if "." in label:
                    name, ext = label.rsplit(".", 1)
                    if len(ext) < 6:
                        available = 15 - len(ext) - 4
                        if available > 3:
                            label = f"{name[:available]}...{ext}"
                        else:
                            label = f"{label[:12]}..."
                    else:
                        label = f"{label[:12]}..."
                else:
                    label = f"{label[:12]}..."
                    
            return label
```

The logic is fine in principle but bear in mind that some file extensions are longer, e.g., xrdml. Perhaps we can just always append the file extension to a smaller filename limit (10?) There may also be a nice way to always generate a label that contains the unique part of a filename given 5 similar filenames...

e.g.,

ABCD-1234.xyz
ABCD-1235.xyz

-> ABC...4.xyz
-> ABC...5.xyz

pydatalab/src/pydatalab/bokeh_plots.py
```diff
@@ -394,7 +431,70 @@ def selectable_axes_plot(
...
-    layout = column(*plot_columns)
+    if plot_points and plot_line:
+        from bokeh.layouts import row
+
+        show_lines_btn = Button(
+            label="✓ Show lines", button_type="primary", width_policy="min", margin=(2, 5, 2, 5)
+        )
+        show_points_btn = Button(
+            label="✓ Show points", button_type="primary", width_policy="min", margin=(2, 5, 2, 5)
+        )
```
ml-evs, line 442: I think for the sake of ease, we can keep only the button for showing the points or not (i.e., always show the line in the background). Perhaps the line in this case could also be made a little bit thicker.

webapp/src/components/datablocks/CycleBlock.vue
```diff
@@ -278,7 +278,7 @@ export default {
 }
 
 .limited-width {
-  max-width: 650px;
+  max-width: 100%;
```
ml-evs, line 281: Can we also port this limited width to the normal bokeh block? I'd like to then play around with different widths a bit (and see if we can even have something resizable...)

ml-evs (commented):
As discussed earlier, this looks really great @BenjaminCharmes! Just a few comments I'd like to address (or at least discuss) before merging.

PR #1277, opened 2025-07-08
Add XRD export to .xy

### LINKED ISSUE #1272: XRD export to .xy
Some refinement packages don't read raw data files (.xrdml in particular). It would be useful to add a feature to export a file as a .xy (2th, I columns, tsv) which can be natively read by Topas.

BenjaminCharmes:
Closes #1272

### DIFF
--- a/pydatalab/src/pydatalab/bokeh_plots.py
+++ b/pydatalab/src/pydatalab/bokeh_plots.py
@@ -18,7 +18,7 @@ from bokeh.models import (
     LinearColorMapper,
     TableColumn,
 )
-from bokeh.models.widgets import Select
+from bokeh.models.widgets import Dropdown, Select
 from bokeh.palettes import Accent, Dark2
 from bokeh.plotting import ColumnDataSource, figure
 from bokeh.themes import Theme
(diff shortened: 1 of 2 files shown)


pydatalab/src/pydatalab/bokeh_plots.py
```diff
@@ -62,6 +62,36 @@
...
+  }
+
+  const encodedUri = encodeURI(xyContent);
+  const link = document.createElement("a");
+  link.setAttribute("href", encodedUri);
+  link.setAttribute("download", "data.xy");
+  document.body.appendChild(link);
+  link.click();
+  document.body.removeChild(link);
+"""
```
ml-evs, line 94: Can we figure out if we can add some of the sample info into this header, as per Matt's comment? https://github.com/datalab-org/datalab/issues/1272#issuecomment-3053848859

I'm not sure if this is currently possible, but I think the block at least knows the item ID, so can just say `"'Generated by datalab from file xyz.xrdml, wavelength lambda for sample <ID>"` as the first line and see if it works in Topas

BenjaminCharmes: Done 👍

pydatalab/src/pydatalab/bokeh_plots.py
```diff
@@ -62,6 +62,43 @@
...
+  xyContent += "' Generated by datalab from file " + filename + ", wavelength " + wavelength + " for sample " + item_id + "\\n";
+
+  for (var i = 0; i < x_values.length; i++) {
+    xyContent += x_values[i].toFixed(5) + "\\t" + y_values[i].toFixed(3) + "\\n";
+  }
+
+  const encodedUri = encodeURI(xyContent);
+  const link = document.createElement("a");
+  link.setAttribute("href", encodedUri);
+  link.setAttribute("download", "data.xy");
```
ml-evs, line 97:
```suggestion
  link.setAttribute("download", filename + ".xy");
```

Perhaps?

pydatalab/src/pydatalab/bokeh_plots.py
```diff
@@ -380,13 +424,38 @@ def selectable_axes_plot(
...
-        save_data_callback = CustomJS(
-            args=dict(source=source),
-            code=GENERATE_CSV_CALLBACK,
-        )
-        save_data.js_on_click(save_data_callback)
-        plot_columns = [save_data] + plot_columns
+        is_xrd_data = any(col in df[0].columns for col in ["2θ (°)", "intensity", "twotheta"])
+
+        if is_xrd_data:
+            dropdown_menu = [("Download .csv", "csv"), ("Download .xy", "xy")]
```
ml-evs, line 430:
```suggestion
            dropdown_menu = [("Download .csv", "csv"), ("Download .xy (TOPAS-compatible)", "xy_topas"")]
```

Just marking out that this may not work for other software

pydatalab/src/pydatalab/bokeh_plots.py
```diff
@@ -380,13 +424,38 @@ def selectable_axes_plot(
...
+            )
+
+            dropdown_callback = CustomJS(
+                args=dict(source=source),
+                code="""
+                if (cb_obj.item == "csv") {
+                    """
+                + GENERATE_CSV_CALLBACK
+                + """
+                } else if (cb_obj.item == "xy") {
```
ml-evs, line 442:
```suggestion
                } else if (cb_obj.item == "xy_topas") {
```

ml-evs (changes requested):
Looks good, some minor suggestions only! Perhaps also need to think about what this button does if there are multiple patterns

PR #1280, opened 2025-07-11
Fix item selection dropdown missing scrollbar in synthesis info

### LINKED ISSUE #1278: Component Selector UI Issue
The scroll bar / scroll functionality in the component selection dialog seems to be somewhat buggy, at least on Firefox:

Image

This doesn't happen with more items, but with exactly 3 matching items, it's impossible to see what the bottom one is.

Cheers!

BenjaminCharmes:
Closes #1278

The `vs__dropdown-menu` was getting cut off because the parent container had `overflow: hidden` for the collapse/expand animation

### DIFF
--- a/webapp/cypress/e2e/batchSampleFeature.cy.js
+++ b/webapp/cypress/e2e/batchSampleFeature.cy.js
@@ -131,7 +131,7 @@ describe("Batch sample creation", () => {
 
     cy.get('[data-testid="batch-modal-container"]').findByText("Submit").should("be.disabled");
     getBatchAddCell(1, 1).type("component1");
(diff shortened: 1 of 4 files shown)


ml-evs (commented):
Thanks for finding the route cause of this @BenjaminCharmes! I think this breaks the clipping of long names, which needs at least `overflow-x` to be `"hidden"`. Can we set this via CSS on the `ItemSelect` (which presumably should always show all the options with a scroll anyway?)

I also see the vue-select has some popper integration that we are not using, it seems to work nicely (though the styles go a bit wonky when I naively enable it via `append-to-body`).

PR #1285, opened 2025-07-25
Bump Pydantic to v2+

BenjaminCharmes:


pydatalab/src/pydatalab/login.py
```diff
@@ -104,6 +104,12 @@ def get_by_id(user_id: str) -> LoginUser | None:
     if not user:
         return None
 
+    if "_id" in user:
+        user["immutable_id"] = str(user.pop("_id"))
+
+    if "managers" not in user:
+        user["managers"] = None
+
```
ml-evs, line 112: Note to self that I want to come back and verify this wrt. model aliases

pydatalab/src/pydatalab/models/__init__.py
```diff
@@ -25,3 +25,9 @@
     "Equipment",
     "ITEM_MODELS",
 )
+
+Sample.model_rebuild()
+StartingMaterial.model_rebuild()
+Cell.model_rebuild()
+Equipment.model_rebuild()
+Collection.model_rebuild()
```
ml-evs, line 33:
```suggestion
{model.model_rebuild() for model in ITEM_MODELS.values()}
```

Might be neater, though not sure why we need this now and not before (will play around with it later)

ml-evs (commented):
Hi @BenjaminCharmes, here's some partial initial comments/suggestions. I'll try to deploy a system based on this PR from scratch (and also investigate some of the new incantations that you've had to use for pydantic 2 compat).

pydatalab/src/pydatalab/models/__init__.py
BenjaminCharmes: Doesn't work with pre-commit for me, need to find out why.

It seems that this is necessary because Pydantic v2 is stricter with forward reference and circular dependencies

BenjaminCharmes: Also need to keep Collection since it's not in ITEM_MODELS

pydatalab/src/pydatalab/models/traits/__init__.py
```diff
@@ -0,0 +1,180 @@
...
+    )
+
+    synthesis_description: str | None = Field(
+        None, description="Free-text details of the procedure applied to synthesise the sample"
+    )
+
+    @model_validator(mode="before")
+    @classmethod
+    def add_missing_synthesis_relationships(cls, values):
+        """Add any missing sample synthesis constituents to parent relationships"""
```
ml-evs, line 57: I've moved this whole module, but want to check this validator as it seems a bit excessive

pydatalab/src/pydatalab/routes/v0_1/users.py
```diff
@@ -41,13 +40,19 @@ def save_user(user_id):
...
+                    {"status": "error", "message": "Invalid display name provided."}
+                ), 400
+            update["display_name"] = display_name
 
         if contact_email or contact_email in (None, ""):
             if contact_email in ("", None):
                 update["contact_email"] = None
             else:
-                update["contact_email"] = EmailStr(contact_email)
+                if "@" not in contact_email or len(contact_email) > 1000:
```
ml-evs, line 53: Things like this really need to be handled at the validator level

PR #1287, opened 2025-07-30
Add in situ XRD block that can handle temperature and electrochemical data

be-smith:
Add XRD insitu block for PR https://github.com/datalab-org/datalab-app-plugin-insitu/pull/67

The new block should look as below with the option to toggle between modes

Screenshot 2025-09-10 at 17 11 54

### DIFF
--- a/pydatalab/pyproject.toml
+++ b/pydatalab/pyproject.toml
@@ -132,7 +132,7 @@ dev-dependencies = [
 ]
 
 [tool.uv.sources]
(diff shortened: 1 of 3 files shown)


BenjaminCharmes (dismissed):
LGTM 👍

ml-evs (approved):

ml-evs (changes requested):
Only one minor fix up, then I think this is good to go!

PR #1311, opened 2025-08-12
Validate block database save/export operations with `DataBlockResponse` model

ml-evs:
This PR initially just updated the "all blocks" test to separate out each block and make sure that they were all doing something sensible. However, I realised that without the app to pass block data back in `update-block` calls, some of the block results was not being saved in the database. This PR fixes that by explicitly calling `_save_to_db` after every block creation call, making the `save_to_db` option somewhat defunct.

Following the principle of least suprise, I think the previous undocumented behaviour (i.e., being able to essentially see a block in an unsaved state) was not expected, so the only side effect here should be improved persistence.

In order to properly implement this, block operations now go through the new `DataBlockResponse` model introduced in #1310, which we can refine over time. I've refactored the NMR, XRD and ChatBlock to extend and use this model directly. No other blocks should be affected, as of yet, as the model still allows extra keys to be passed through transparently. A future PR will have to try restrict this and refactor all blocks to store data in the correct places whilst maintaining backwards compat.

### DIFF
--- a/pydatalab/src/pydatalab/models/blocks.py
+++ b/pydatalab/src/pydatalab/models/blocks.py
@@ -6,6 +6,9 @@ from pydatalab.models.utils import JSON_ENCODERS, PyObjectId
 class DataBlockResponse(BaseModel):
     """A generic response model for a block, i.e., what is stored in `self.data`
     in the corresponding DataBlock class.
+
+    It is expected but not mandatory that this model will be extended by the specific block type
+    where possible.
     """
 
     blocktype: str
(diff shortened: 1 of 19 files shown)


pydatalab/src/pydatalab/models/blocks.py
```diff
@@ -32,12 +32,21 @@ class DataBlockResponse(BaseModel):
...
+
+    b64_encoded_image: dict[PyObjectId, str] | None
     """Any base64-encoded image data associated with the block, keyed by file_id, if any."""
 
-    bokeh_plot_data: str | None = None
+    bokeh_plot_data: dict | None
     """A JSON-encoded string containing the Bokeh plot data, if any."""
 
+    processed_data: dict | None = None
+    """Any processed data associated with the block, small enough to store."""
```
jdbocarsly, line 48: Should we estimate what "small enough to store" means? Technically its something like 16 MB since that is the maximum document size, but presumably the idea is to store smaller things here?

pydatalab/src/pydatalab/models/blocks.py
```diff
@@ -32,12 +32,21 @@ class DataBlockResponse(BaseModel):
...
+
+    warnings: list[str] | None = None
+    """Any warnings that occurred during block processing."""
+
+    b64_encoded_image: dict[PyObjectId, str] | None
     """Any base64-encoded image data associated with the block, keyed by file_id, if any."""
 
-    bokeh_plot_data: str | None = None
+    bokeh_plot_data: dict | None
     """A JSON-encoded string containing the Bokeh plot data, if any."""
```
jdbocarsly, line 46: Can we also have a metadata field here? We use it at least in the NMR block.

pydatalab/src/pydatalab/models/blocks.py
ml-evs: I've just expanded on these docstrings a bit as a first hash (and added `metadata`). We can iterate on these as we migrate all blocks to using and extending them. One thing I'm a bit worried about is that some metadata will necessarily need to be keyed by file ID, like `wavelength` in an XRD block with multiple measurements, so I haven't quite locked that part in yet. We also need the distinction between computed and provided properties, then need to prevent the app from sending back and computed properties (in fact, I only realised this evening that we call `save-item` with all unfiltered block data, which I've stripped back manually in [1fba85](https://github.com/datalab-org/datalab/pull/1311/commits/61fba85a58d1ae29fa5a0c621c7eb6dfdea2dcaa) -- I think ultimately a lot of this stuff will be nested so that we can ignore anything under the `computed` key -- this is fine for the app, but the API will then have to do the work of making old blocks backwards compatible.

ml-evs: </braindump>

ml-evs:
This is the first of several scary technical debt resolving PRs, with this one changing how we save block data -- I've tried to test the current behaviour extensively so that we know when we are breaking it, but will need several follow-ups to refine it. Next up, trying to get pydantic v2 in... this PR started as an off-shoot for that to just robustify some of the tests!

PR #1338, opened 2025-08-28
Improve regex item search: implict word boundaries, chaining and literal matches

### LINKED ISSUE #1336: Search function in synthesis information block not working correctly
There is an issue with the search function in the synthesis information block. When searching for compounds they seem to appear only when certain parts of the name are typed in. For example, when searching for ‘vanadium(II) chloride’ it will not appear in the list when either ‘v’ or ‘vanadium(II)’ is searched, but will appear when ‘vanadium’ is searched. I also noticed this when searching for samples created within datalab, not just those from the inventory. For some compounds this is ok, but some won’t appear seemingly regardless of what the search input is.

ml-evs:
Closes #1336 by:

- sanitizing user input for search and matching on literal chars
- inserting implicit word boundaries around whitespace
- chaining whitespace delimited search query into multiple queries
- for queries longer than 5 chars, do not add word boundaries

### DIFF
--- a/pydatalab/src/pydatalab/routes/v0_1/items.py
+++ b/pydatalab/src/pydatalab/routes/v0_1/items.py
@@ -317,7 +317,9 @@ def search_items():
         query = query.strip("'")
 
     if isinstance(query, str) and query.startswith("%"):
+        # Old FTS query style, using MongoDB text indexes
         query = query.lstrip("%")
+        query = query.strip("'")
         match_obj = {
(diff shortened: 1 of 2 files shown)


jdbocarsly (dismissed):
Seems to work very nicely! The one thing is that I don't understand exactly the rules, i.e. surrounding short vs. long. A slightly longer writeup of the behavior somewhere would be nice so we can reference for future debugging/ additional enhancements.

PR #1343, opened 2025-09-05
Migrate from TinyMCE to QuillJS editor

### LINKED ISSUE #1232: Migrate away from TinyMCE
We should probably move off TinyMCE now they have moved to a paid/freemium model for even v5.11 security updates.

BenjaminCharmes:
Closes #1232 

Migrates from TinyMCE  to QuillJS.
Add component for backward compatibility with existing TinyMCE


jdbocarsly:
If you haven't already, can you look a little into how custom tags would be handled in Quill? We would like to be able to type @ and select another item to link to. The editor would then insert in the colored badge according to item type. Ideally this would even be handled by just inserting the vue component, but inserting a vue component into the editor's environment may or may not be possible (I think this was very hard when I tried to do it in TinyMCE).

These tags don't have to be implemented in this PR, but it would be good to make sure we have a path towards getting the tags in QuillJS

PR #1345, opened 2025-09-12
[WIP] Add Mermaid support to Quill

BenjaminCharmes:
Additional work from issue #1232 and PR #1343 

Already works nicely, but we need to store `description` in Delta format inside the database to make Mermaid work in QuillJS. Could be improved further if needed.

Some issues between Quill 2 and quill-mermaid (which seems to have been made for Quill 1), but difficult to find another library that works with Quill 2, need to spend more time on this issue.


ml-evs:
Thanks @BenjaminCharmes, this superseded by #1349 where we chose TipTap!

PR #1349, opened 2025-09-17
Migrate from TinyMCE to TipTap

### LINKED ISSUE #1232: Migrate away from TinyMCE

BenjaminCharmes:
Closes #1232

Duplicate of the PR #1343 , but this one uses Tiptap and supports all the basic features of TinyMCE.
Perhaps it would be easier to merge the two to compare Quill and Tiptap? After trying both, TipTap seems more robust for the long term.


BenjaminCharmes:
Hey @jdbocarsly, as discussed last week in the PR using QuillJS, cross-referencing (should) work in this PR if you want to test it. I think I'm going to abandon QuillJS in favour of Tiptap to replace TinyMCE.

jdbocarsly:
This is really slick! Thanks for doing this. The @ tagging is really smooth, and I really like the mermaid editor! 

A few notes:
1. Is this on tiptap v2? Seems like v3 was released this summer...
1. When  a tag is in the text and you are not in edit mode, clicking on the tag opens edit mode-- is there a way we could get this click to go straight to the tagged link (potentially in a new tab), instead of opening edit mode?
1. Is there a way to directly input images into tiptap? TinyMCE had a feature for that, which I think broke at some point on our deployment. That's really helpful for adding in photos, etc. to describe a procedure. I think tinymce would b64 encode it and put it directly in the text, which could cause an issue with the database since MongoDB documents are limited to 16 gb. We could think about an alternate workflow where the image is automatically uploaded to the sample and then embedded by link...
1. How is the markdown support here? Is it possible to have a toggle between "markdown" and "preview", like on Github?

ml-evs (commented):
Looking really nice @BenjaminCharmes! As discussed, let's get this merged with some final tweaks and things to think about:

- [ ] Either write a formal test or just check what an old complex (e.g., colours, tables, images) TinyMCE description looks like when rendered here 
- [ ] Make sure the cross-ref item search can also search for equipment (may need future API change)
- [ ] Investigate image plugin for tiptap for uploading images
- [ ] Tweak button for equations to say f(x) (or whatever)
- [ ] Enable markdown preview/toggle either in this PR or future

ml-evs (changes requested):
Thanks again for this @BenjaminCharmes! Happy to merge this with one minor style tweak: can we try to replicate the old behaviour with:

- placeholder text for "Add a description" in all editors
- border hidden unless hovered/active
- slight extra padding (side and top) to avoid the editor touching the bottom of blocks

ml-evs (approved):

PR #1353, opened 2025-09-22
Added a comparison panel to the CycleBlock for displaying multiple files for comparison

be-smith:
Adds a panel for comparison - handles legends and different colourmaps for different files etc.

Future add-ons

1. Add a mass for each file (probably needs to be its own analysis block or page for this - could grab the mass from each sample or cell you select for analysis
2. Different cycle plotting per file - I haven't played around with selecting different cycles for analysis from each file yet.
3. dQ/dV analysis where you can sum two profiles together to see how they fit to a full cell (again a much more complicated block)

Screenshot 2025-09-22 at 15 16 06


jdbocarsly:
This is a cool step forward. I would like to maintain a differentiation in the UI between the one (or multiple stiched) data sets that the block is keeping track of, and the one or more samples that are being compared to it. So, instead of having it as a tabbed option, I think we should have an option on the main tab to add additional comparisons to the graph only. The block itself is in charge of keeping track of the main experiment along with any experimental details/metadata, and displaying additional comparisons is a helpful additional feature. 

This is the same thing we discussed with the XRD block, which currently has a similar issue.

be-smith:
I've split up the "main" block files and the comparison files into different file selectors. They are also stored as different fields:
`file_ids`
`comparison_file_ids`

Screenshot 2025-10-02 at 18 20 24

jdbocarsly:
yeah, that's exactly what I was imagining! And yes, agree that some way of grabbing the trace from another sample would be good (and that way you can normalize by the correct mass, ideally).

By the way, while you are at this-- would it be easy to add a second x- axis to the echem capacity vs voltage plot, which has capacity in units of Li (or Na) transferred per mol? We store the molar mass if the user has put in a formula, so I think in principle this should be pretty straightforward.

PR #1367, opened 2025-10-02
Item graph display too many items

### LINKED ISSUE #629: BUG: Local item graph contains everything sometimes
Somehow being a "manager" of a set of samples means that the item graph renders everything, making sample page loading very slow and unhelpful.

### LINKED ISSUE #1355: Item graph can slow down edit page significantly
Reported by @elbee99.

The item graph seems to choke up the loading of the edit page. First easy fix -- make this loading asynchronous.

However, I think this is related to another bug where the item graph for a single item seems to go way too deep, and returns a whole chunk of the overall item graph -- so we should investigate this too.

BenjaminCharmes:
Closes #1355 
Closes #629

Test from #1360 work with this fix (local at least)

### DIFF
--- a/pydatalab/tests/server/test_graph.py
+++ b/pydatalab/tests/server/test_graph.py
@@ -129,8 +129,8 @@ def test_simple_graph(admin_client):
     assert len(graph["edges"]) == 2
 
     graph = admin_client.get("/item-graph/parent").json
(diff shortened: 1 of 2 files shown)


ml-evs (approved):
Works great, thanks!

PR #1368, opened 2025-10-03
clean up relationships and constituents on item deletion

BenjaminCharmes:


ml-evs (commented):
Thanks @BenjaminCharmes, as discussed, I think we need to think about this a little bit more wrt. #831 (if it ever gets merged) -- for now I think the better approach is to sort this out at the get-query level and filter out any non-existent/deleted items.

Issue #1046, opened 2025-02-12
Datalab seems similar to eln


SteffenBrinckmann:
Hey, looking into datalab, it seems similar to an eln. Would you be interested in joining the eln condition and supporting the .eln for format?

ml-evs:
Hi @SteffenBrinckmann, in general, yes, though probably initially only export. I'm hoping this will be an outcome of the next MADICES for datalab.

SteffenBrinckmann:
- It might be easiest to start with an example version. From my point of view, the kadi4mat-collections example has all important feature included but no extraneous.  [link](https://github.com/TheELNConsortium/TheELNFileFormat/blob/master/examples/kadi4mat/collections-example.eln) 
- The biggest obstacle from my experience is getting the file RO-crate compliant. In the repo there is a pytest, include your .eln in it and remove the others. And your .eln will be focused on, locally - no github CI required.
- just create a pull-request and it will be included.
- if you have questions, ... I can help.

ml-evs:
Thanks @SteffenBrinckmann, we'll have some people looking at this over the next month before MADICES (tracked also at #1332).

PR #1371, opened 2025-10-06
Add ELN collection export

### LINKED ISSUE #1332: ELNFileFormat implementation roadmap
We should add a collection export endpoint with a UI button for saving a whole collection to a .eln file: https://github.com/TheELNConsortium/TheELNFileFormat

Initially, just folder per sample, zip up all attached files and metadata and let the user download

This might be too big to do in one process/request, so may need to think about endpoint that works like:

1. `/export/` -> API returns "ok i'll do it" with a link like `/elnexport/asdfdsafa`
2. API triggers background process to create .eln file (Celery, APScheduler, ...) -- probably just our mongo database with a new collection `tasks`
3. UI has to ping that link `/elnexport/asdfdsafa` and API will respond `ready` or `not-ready` with another link for download

BenjaminCharmes:
Closes #1332 & #1046 

Implements a new API endpoint and background process to export full collections as .eln files


pydatalab/src/pydatalab/routes/v0_1/export.py
```diff
@@ -0,0 +1,161 @@
...
+
+    collection_exists = flask_mongo.db.collections.find_one({"collection_id": collection_id})
+    if not collection_exists:
+        return jsonify({"status": "error", "message": "Collection not found"}), 404
+
+    collection_with_perms = flask_mongo.db.collections.find_one(
+        {"collection_id": collection_id, **get_default_permissions(user_only=True)}
+    )
+    if not collection_with_perms:
+        return jsonify({"status": "error", "message": "Access denied"}), 403
```
ml-evs:
```suggestion
    collection_with_perms = flask_mongo.db.collections.find_one(
        {"collection_id": collection_id, **get_default_permissions(user_only=True)}
    )
    if not collection_with_perms:
        return jsonify({"status": "error", "message": "Collection not found"}), 404
```

Better not to leak IDs that exist but the user does not have permission for, so just treat missing ID as 404

ml-evs: Can probably also set `user_only=False`

ml-evs (commented):
This looks fantastic, thanks @BenjaminCharmes! 

Just writing a couple of points here that we can discuss:

- configuring amount of resources APScheduler is allowed to use
- using pydantic models rather than direct database queries to create items etc.
- a few bits of permissions we need to be careful with

pydatalab/src/pydatalab/export_utils.py
```diff
@@ -0,0 +1,169 @@
...
+                json.dump(item_metadata, f, indent=2, ensure_ascii=False)
+
+            if item.get("file_ObjectIds"):
+                for file_id in item.get("file_ObjectIds", []):
+                    file_data = flask_mongo.db.files.find_one({"_id": ObjectId(file_id)})
+                    if file_data:
+                        source_path = Path(file_data["location"])
+                        if source_path.exists():
+                            dest_file = item_folder / file_data["name"]
+                            shutil.copy2(source_path, dest_file)
```
ml-evs: Somehow the actual raw data files are not making it into the `.eln` file when I try to export them

ml-evs (dismissed):

PR #1378, opened 2025-10-20
Filter out deleted items from graph relationships

BenjaminCharmes:
## Changes
* Modified `get_graph_cy_format` to filter out relationships pointing to non-existent/deleted items at the query level
* Prevents "nonexistent source" errors in the graph view without modifying the database

## Question
* Deleted items are still present in `synthesis_constituents` and remain clickable (even if leading to errors). Should we handle this in the UI (disable links and/or different badge)?


ml-evs:
I think this one has now gone stale and has been fixed in other ways.

PR #1379, opened 2025-10-22
Add Cypress e2e tests with authenticated user login via magic links

### LINKED ISSUE #1363: UI testing as fake user
Bugs like #1361 can't be caught unless we allow our e2e tests to fake user/admin login.

This should be possible to do via the magic-link mechanism, i.e., we insert a predefined magic link in the database and let the UI tests login with it. This might be a little bit trickier to handle for local tests.

Some simple UI tests can then be written for functionality that only works when the user is logged in, e.g., the admin dashboard.

BenjaminCharmes:
Closes #1363

Implements UI testing as authenticated users to catch bugs that only occur when logged in. Adds `/testing/create-magic-link` and `/logout` endpoints, plus Cypress commands `loginViaTestMagicLink()` and `logout()`. Includes tests for authenticated sample creation, admin dashboard access, and multi-user sample visibility.

### DIFF
--- a/pydatalab/src/pydatalab/routes/v0_1/auth.py
+++ b/pydatalab/src/pydatalab/routes/v0_1/auth.py
@@ -443,6 +443,19 @@ def _validate_magic_link_request(email: str, referrer: str) -> tuple[Response | 
 
 
 def _generate_and_store_token(email: str, is_test: bool = False) -> str:
+    """Generate a JWT for the user with a short expiration and store it in the session.
+
+    The session itself persists beyond the JWT expiration. The `exp` key is a standard
+    part of JWT that PyJWT treats as an expiration time and will correctly encode the datetime.
+
+    Args:
+        email: The user's email address to include in the token.
+        is_test: If True, generates a token for testing purposes that may have different
+                 expiration or validation rules. Defaults to False.
+
+    Returns:
+        The generated JWT token string.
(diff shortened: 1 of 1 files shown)


pydatalab/src/pydatalab/routes/v0_1/auth.py
```diff
@@ -448,25 +439,29 @@ def generate_and_share_magic_link():
...
         )
 
-    # Generate a JWT for the user with a short expiration; the session itself
-    # should persist
-    # The key `exp` is a standard part of JWT; pyjwt treats this as an expiration time
-    # and will correctly encode the datetime
+    return None, 200
+
+
+def _generate_and_store_token(email: str, is_test: bool = False) -> str:
```
ml-evs: Can we modify the info in the old comment as a docstring (cannot format it easily in the github interface) and add an explanation of `is_test`?

```suggestion
def _generate_and_store_token(email: str, is_test: bool = False) -> str:
"""
    # Generate a JWT for the user with a short expiration; the session itself
    # should persist
    # The key `exp` is a standard part of JWT; pyjwt treats this as an expiration time
    # and will correctly encode the datetime
"""
```

ml-evs (changes requested):
One more minor comment, otherwise good to go!

Issue #1392, opened 2025-10-30
Improve server logs for block errors


jdbocarsly:
Currently, block errors are nicely caught by the base datablock (`base.py`), allowing errors and warnings to be displayed in the frontend. However, this catching means that the log always reports all errors as coming from the file `base.py`, instead of the file/line where the error actually occurred. 

For example:
<img width="1341" height="85" alt="Image" src="https://github.com/user-attachments/assets/a78371a6-0711-4358-9ce1-4e745bd76495" />

It would improve DX to log more informative information about where the error originated.

ml-evs:
Good suggestion, I have a branch playing around with this already (can't remember if I pushed it). I think what needs to be done for logging is removing the "base" logger (which also pollutes all blocks with having to load the server config) and replace it with module-local logging that adopts the server settings when they are configured. We might also want to look at generally switching to something like https://www.structlog.org/en/stable/index.html for this.

jdbocarsly:
I put in enough to get through basic block development in #1393, just making sure the info is at least there.

ml-evs:
I think we can consider this closed after #1984

PR #1393, opened 2025-10-30
Log more informative block errors

jdbocarsly:
Edits the Warning log message printed by datablocks when a block experiences an error. Now, the errors look something like this:
Screenshot 2025-10-30 at 12 22 37 PM

(note, the first commit adds the line numbers and filenames of the actual log, the second commit pretty prints the data, which previously we were just printing. Pretty printing could lead to some long logs in the case of blocks with lots of data, so we can take this out if needed. We were already printing the data dictionary, so it could get long regardless.)

addresses issue #1392


jdbocarsly:
Actually, Probably the data dictionary part could be printed with a debug level after the warning, so that logs don't get too long (assuming we don't typically save debug data to server logs)

ml-evs (approved):

PR #1430, opened 2025-11-17
Enforce refcode requirement for synthesis constituent references

### LINKED ISSUE #1173: Synthesis constituents should be added from UI by refcode rather than `item_id`

BenjaminCharmes:
Closes #1173 

Prevents creating synthesis constituents via API with only `item_id`.

**Changes:**
- Modified `EntryReference` validator to require `refcode` or `immutable_id` when `item_id` is present
- Updated test fixtures to include refcodes in constituent references

### DIFF
--- a/pydatalab/src/pydatalab/models/utils.py
+++ b/pydatalab/src/pydatalab/models/utils.py
@@ -276,11 +276,21 @@ class EntryReference(BaseModel):
 
     @root_validator
     def check_id_fields(cls, values):
-        """Check that at least one of the possible identifier fields is provided."""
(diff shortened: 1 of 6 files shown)


ml-evs:
Superseded by #1727 and #1755 -- we now look up the refcode when only an item ID is provided (I will pinch the test cases though!)

PR #1444, opened 2025-11-20
Add minimum resizable column width based on header content

### LINKED ISSUE #1439: Minimum resizable column width in datatable
Right now, its quite easy to shrink columns to smaller than their header sizes, which hides the filter option. I mostly noticed this for `Date`, since my column widths are saved locally, the addition of the filter button was hidden.

@BenjaminCharmes is this something you could look at? I think now that we have the ability to hide columns, being able to resize columns to much smaller than the header is not that useful.

BenjaminCharmes:
Closes #1439 

Prevents columns from being resized smaller than their header content, ensuring filter buttons remain visible.

### DIFF
--- a/webapp/src/components/DynamicDataTable.vue
+++ b/webapp/src/components/DynamicDataTable.vue
@@ -74,7 +74,12 @@
       </template>
       <template #empty> No entries found. </template>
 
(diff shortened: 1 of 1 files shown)


ml-evs (approved):
Works well, thanks @BenjaminCharmes! I'm going to raise a separate minor issue about the fact that resizing seems to trigger the sort

PR #1454, opened 2025-11-24
Unify tooltip styling and add block version display

### LINKED ISSUE #1429: Unify styling of info boxes
There's now three sources of "info" about blocks:

1. hovering over the entry in the add a block menu
2. hovering over the (i) button in the block title
3. hovering over certain fields in the block
4. hovering over field names in the edit page (introduced in #1410)

They should probably all use the same component, but atm they do not. Some differences I'm aware of:

- (1) obeys line breaks but (2) does not (see e.g., echem block)
- (3) uses native html alt text so is not styled at all

BenjaminCharmes:
Closes #1429

Consolidates all tooltip/info box implementations to use a single styled component system.


ml-evs (commented):
Thanks for this! I don't think the app fonts are being set correctly though, at least for me it looks like this (i.e., small, and non-figtree):

2025-11-28-204717_747x310_scrot

PR #1455, opened 2025-11-25
Prevent double-click on modal submit button

### LINKED ISSUE #1453: Disable create button in modal after clicking it
I've noticed on bad internet that you can click on the create button twice as the modal only disappears after the API has responded/created. We should disable the button immediately before making the API request.

BenjaminCharmes:
Closes #1453

Prevents double-clicking the submit button in modals by immediately locking it before the API responds.

Changes:
- Changed submit button from `type="submit"` to `type="button"` with manual form submission
- Added `submitLocked` flag that locks on first click
- Use `requestSubmit()` with fallbacks to trigger form submission

### DIFF
--- a/webapp/src/components/AddToCollectionModal.vue
+++ b/webapp/src/components/AddToCollectionModal.vue
@@ -1,6 +1,10 @@
 <template>
   <form class="modal-enclosure" data-testid="add-to-collection-form" @submit.prevent="submitForm">
-    <Modal :model-value="modelValue" @update:model-value="$emit('update:modelValue', $event)">
+    <Modal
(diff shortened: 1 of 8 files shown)


ml-evs (commented):
Thanks for this @BenjaminCharmes!

Couple of points that we are discussing as I type:

- this approach looks good as it works for all modals, but the lockAndSubmit method needs some comments to help me understand it for the future
- also the original bug behaviour is a bit strange -- the duplicate items seem to disappear on refresh, indicating that the API requests are never actually made, so this may need to be fixed elsewhere too

PR #1461, opened 2025-11-27
Add contribution graph visualization for user and community activity

### LINKED ISSUE #1418: GH contribution style graph for user accounts
Really fun idea suggested by @elbee99 !

BenjaminCharmes:
Closes #1418

Adds GitHub-style contribution heatmap graphs to visualize user activity:

- **About page**: Combined activity graph showing all users' contributions over the past year
- **Account Settings modal**: Individual user activity graph showing personal contributions over recent months

### DIFF
--- a/webapp/src/components/UserActivityGraph.vue
+++ b/webapp/src/components/UserActivityGraph.vue
@@ -1,13 +1,13 @@
 <template>
   <div class="activity-graph-container" :class="{ compact: compact }">
-    <h5 v-if="title">{{ title }}</h5>
     <div v-if="isLoading" class="text-center p-4">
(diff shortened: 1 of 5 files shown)


webapp/src/components/UserActivityGraph.vue
ml-evs: Really nice, without adding any extra deps! As per my message in Slack, I wonder if we need to make it look a *teency* bit less like GitHub's, so any suggestions would be appreciated! (even if its just changing the colours around a bit)

BenjaminCharmes: Screenshot 2025-12-01 at 14 06 23

PR #1468, opened 2025-12-01
Allow normal click to open items in same tab from table

### LINKED ISSUE #1465: Allow samples to be opened from table in same tab
Feedback from Annalena's group, which I think I agree with. Right now, modified click does nothing on the dynamic data table, and normal click opens a sample in a new tab. Probably we should allow modified click => new tab and normal click => same tab.

BenjaminCharmes:
Closes #1465

Changes click behavior in DataTable for both items and collections:
- Normal click: opens in same tab
- Ctrl/Cmd + click: opens in new tab


jdbocarsly:
Can you make the modified click open the page in a new tab in the background (i.e., don't change focus to the new tab)?That's standard for browsers.

PR #1472, opened 2025-12-04
Add error handling for file upload size limits

### LINKED ISSUE #1384: Add error handling to uppy when file uploads are too large
e.g.:

- capture uppy errors and don't let it restart repeatedly
- API can report max file size, uppy should then enforce max file size

BenjaminCharmes:
Closes #1384

- API now reports `max_upload_bytes` via `/info` endpoint
- Uppy enforces max file size from API config (instead of hardcoded limit)
- Captures and displays errors for:
  - Files exceeding size limit (before upload starts)
  - Server 413 responses (if client check is bypassed)
- Removes failed files from Uppy to prevent retry loops
- Added backend test for `/info` endpoint
- Added Cypress tests for upload error scenarios


ml-evs (approved):
Works nicely, thanks!

PR #1483, opened 2025-12-09
Redirect non-authenticated users to About page

### LINKED ISSUE #492: Make the default non-logged in page "About" rather than "Samples"
Currently the default view is samples. As a new user arriving at the platform, you will just see an empty table. We should reroute non-logged in users to About, whilst still allowing logged in users to directly start working on samples.

### LINKED ISSUE #1557: Broken logos on about page
I guess my browser was caching them before, but the Leverhulme site does not have proper CORS headers to be served from our site.

BenjaminCharmes:
Closes #492
Closes #1557 

Changes default landing page behavior based on authentication status.
- Non-authenticated users: redirected to About page
- Authenticated users: redirected to Samples page


ml-evs (dismissed):
Works well, thanks!

ml-evs (commented):
Hmmm, actually, just testing it a bit and realising that every load page now hits `/get-current-user` 3 times, with this PR adding the extra one... could you take a look at this and make better use of the store/cache and the `setCurrentUserInfoLoading` state in the store?

webapp/src/store/index.js
```diff
@@ -423,7 +429,48 @@ export default createStore({
...
+        return state.currentUserInfoPromise;
+      }
+
+      if (!fullInfo && state.currentUserInfoLoaded) {
+        return state.currentUserID !== null;
+      }
+
+      commit("setCurrentUserInfoLoading", true);
+
+      const promise = fetch(`${API_URL}/get-current-user/`, {
```
ml-evs: Sorry to be a pain but I'd rather all the direct API fetches are done in `server_fetch_utils`, we likely need to use the `fetch_get` wrapper there anywhere.

Could you put this back there and manage the store state upon request?

ml-evs (approved):
Thanks @BenjaminCharmes! Also took this opportunity to fixup the about page and close #1557.

PR #1484, opened 2025-12-09
Fix NotificationDot positioning in user dropdown menu

BenjaminCharmes:
Fixes the NotificationDot component placement in UserDropdown.vue. The dot now displays correctly inline next to "Administration" instead of appearing on a separate line at the top of the dropdown.


ml-evs (approved):
Works well, thanks @BenjaminCharmes!

PR #1491, opened 2025-12-12
Add prototype asynchronous data block processing backend

### LINKED ISSUE #1469: Asynchronous data block processing
Placeholder issue to track this process. Current idea: add a `prefers_async` annotation to data blocks that need to be processed "offline". Develop an API spec for how a block should respond as its being processed, and let the UI poll and reload this. Then use `APScheduler` as our initial backend for offline tasks, following its inclusion in #1371.

Future work: use more robust/decentralised backend like RMQ/Arq + celery, but for now APScheduler will give us 90% of the benefit with limited extra work.

BenjaminCharmes:
Closes #1469

Linked to [Add prefers_async flag to insitu blocks #76](https://github.com/datalab-org/datalab-app-plugin-insitu/pull/76)

## Summary
Implements asynchronous data block processing using APScheduler for computationally intensive blocks, with a unified task management system for all async operations.

### Backend

**Task Management Refactoring:**
* **Unified Task model** replacing separate `BlockTask` and `ExportTask` models
* **Single `tasks` MongoDB collection** with `type` field to distinguish export vs block processing tasks
* **Consolidated status enums** - `TaskStatus` replaces `ExportStatus` and `BlockProcessingStatus`

**Async Block Processing:**
* Added Task model to track async processing status (PENDING/PROCESSING/READY/ERROR)
* Added TaskStage model to log timestamped processing stages with severity levels (info/warning/error)
* Modified `update_block()` to support async processing via `prefers_async` attribute
* Added `/blocks/<task_id>/status` endpoint for status polling, including stage history
* Reused existing export_scheduler infrastructure for background job execution
* Added AsyncCanaryBlock for testing async processing workflow

### Frontend
* Added `pollBlockStatus()` to poll task status every 2 seconds
* Modified `updateBlockFromServer()` to handle async responses
* Added real-time progress display showing current processing stage in UI
* Updated insitu block components (NMR & UV-Vis) to trigger async processing only when all required data is present

## Implementation Details
* Blocks annotate with `prefers_async = True` to enable async processing
* UI sends `trigger_async: true` in `event_data` when ready to process
* Processing stages logged at each step (starting, loading, processing, generating viz, saving, completion)
* `bokeh_plot_data` is regenerated on-demand via `block.to_web()` in status endpoint
* Only creates async tasks when all folders/parameters are selected
* Stage information displayed as blue info alert during processing, auto-cleared on completion
* Task type filtering ensures export and block processing tasks remain isolated despite sharing infrastructure

EDIT by @ml-evs:

Took over this PR and tweaked it so that block processing does not go via APScheduler, but instead uses a separate task processor per flask/gunicorn worker  that is in charge of updating the mongodb tasks collection. The results are also pushed to gridfs to use as a transfer buffer.


ml-evs (commented):
Excited to try this @BenjaminCharmes! Let me know when you think its ready to play around with! I would suggest also extending the `CanaryBlock` to make an async version we can use in testing.

pydatalab/src/pydatalab/models/block_task.py
```diff
@@ -0,0 +1,43 @@
+from datetime import datetime, timezone
```
ml-evs: Could we combine the block tasks and export task models and MongoDB collection?

PR #1514, opened 2026-01-08
Add API search functionality to data tables

### LINKED ISSUE #1434: Enable API search within the data table
Longer-term goal, hopefully enabled by #886

BenjaminCharmes:
Closes #1434

Adds a toggle switch to data tables allowing users to search via the API endpoint (`/search-items/`) instead of client-side filtering.

**Changes:**
- Added UI/API toggle switch in table header
- API search uses existing `/search-items/` endpoint
- Debounced search input to reduce API calls (400ms)
- Search results replace table data when API mode is active


ml-evs (commented):
Looks like a great prototype, thanks @BenjaminCharmes 

Some quick thoughts:

- Can we see how this works with pagination from the API too? i.e., we should make the default samples listng route only return 50 entries for now, and see how both kinds of searches could deal with this (might mean we switch to full API search for everything)
- I don't know how the client-side filtering (as opposed to searching) will handle this pagination, as it won't know what the possible values are
- Then, as discussed, it would be nice to replace the UI/API toggle into an "advanced search" modal that for now just accepts a raw query string (and in future could be a query builder). For now, this query string will just be the free-text search string the user wants.

PR #1520, opened 2026-01-09
Unify single item and related item export UI

### LINKED ISSUE #1518: Unify single item and related item export UI
We should maybe have an option in the related items export UI that exports the single item, rather than a dropdown.

BenjaminCharmes:
Closes #1518

Removes redundant dropdown menu with single export option. Users now click a single "Export" button which:
- For collections: exports directly
- For items: opens modal with option to export with or without related items


ml-evs (approved):
Picked some of the style changes out of #1546 and now this lgtm, thanks @BenjaminCharmes

PR #1525, opened 2026-01-12
Add batch sharing menu option in data table and API

### LINKED ISSUE #1523: Batch sharing menu
The check-box data table options should have an option for batch sharing items with people/groups, in the same way that you can add a collection right now.

BenjaminCharmes:
Closes #1523

Adds a "Share with people/groups" option to the checkbox dropdown menu in data tables. Users can select multiple items and batch-add creators (read/write) and groups (read-only) while preserving existing permissions.

## Implementation:

**Backend:**
- Refactored permissions logic into `_process_item_permissions` helper function
- `PATCH /items/<refcode>/permissions` - Replace mode (existing behavior)
- `PUT /items/<refcode>/permissions` - Append mode (new)
- Server-side concatenation eliminates need to fetch full items
- Returns `"No changes needed"` when permissions already exist

**Frontend:**
- New `BatchShareModal` component with user/group selectors
- `appendItemPermissions()` utility function for PUT requests
- Handles batch operations with informative logging

**Tests:**
- Added tests for append mode with creators and groups
- Verified no duplicate permissions
- Tested PATCH vs PUT behavior
- Validated base owner preservation
- Error handling for invalid IDs

**Current limitation:** This implementation only supports adding collaborators. Batch removal could be added in a future PR if needed.

### DIFF
--- a/pydatalab/src/pydatalab/routes/v0_1/items.py
+++ b/pydatalab/src/pydatalab/routes/v0_1/items.py
@@ -746,11 +746,9 @@ def create_samples():
     )  # 207: multi-status
 
 
-@ITEMS.route("/items/<refcode>/permissions", methods=["PATCH"])
-def update_item_permissions(refcode: str):
-    """Update the permissions of an item with the given refcode."""
-
-    request_json = request.get_json()
+def _process_item_permissions(
+    refcode: str, request_json: dict, append_mode: bool = False
+) -> tuple[dict, int]:
     creator_ids: list[ObjectId] = []
     group_ids: list[ObjectId] = []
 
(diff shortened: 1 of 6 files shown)


ml-evs (commented):
Thanks for this @BenjaminCharmes! After looking at it, I don't really like they we have to pull each full item just to get the permissions before constructing the query. Could you instead extend the item permissions API routes to have a "PUT" only mode that appends permissions? It will need a bit of refactoring to mesh nicely with the existing PATCH route, but it essentially would be just doing the concatenation on the server-side instead.

We should also just double check if you select an item for batch sharing what happens if it already has those permissions? i.e., do we need a "no permissions to change" state or something in the UI?

ml-evs (approved):
Works really nicely now, thanks @BenjaminCharmes!

PR #1526, opened 2026-01-13
Add group icon display in creators column

### LINKED ISSUE #1521: Adding table icon for group sharing
When an item is shared with a group, I think the data table column for creators should also have an icon for that group.

BenjaminCharmes:
Closes #1521

Displays groups alongside creators in data tables with filtering support for both.

**Changes:**
- Backend: Added groups lookup in item summary queries  
- Created GroupsIconCounter component (shows count badge for multiple groups)
- Created FormattedGroupName component (circular icon + name, matching creator style)
- Added creatorsAndGroups virtual field for unified creator/group filtering
- Updated Creators component to display groups and hide creator names when groups present
- Updated ToggleableGroupsFormGroup to use DialogService for permission confirmations


ml-evs (commented):
Also looks good, can you rebase this after I merge #1533? 

I think the idea of making the column component be a variant of the other one (e.g., display the number if multiple groups are passed).

I would also suggest adding the logic to the table (or encapsulating in a component one level higher up) that the user's names are hidden when there are groups present too.

BenjaminCharmes:
> Oh and while I think of it (I didn't test myself yet) -- does the creator filtering work with groups now too?

I haven't thought about adding it -- or testing it, but I don't think it works at the moment without modification -- But yes, of course, I'll take care of it.

ml-evs (dismissed):
Looks awesome, thanks @BenjaminCharmes!

I have a tiny niggle that I'm happy to merge and then fix, but perhaps you will have a chance to fix it first... in the dynamic data table, the user bubble and group bubble are not quite aligned:

Image

pydatalab/src/pydatalab/routes/v0_1/items.py
```diff
@@ -1354,7 +1364,8 @@ def restore_version(refcode):
         "version": next_version_number,
         "timestamp": datetime.datetime.now(tz=datetime.timezone.utc),
         "action": VersionAction.RESTORED,  # Audit trail: this is a restored version
-        "restored_from_version": version_object_id,  # Track which version was restored from (ObjectId)
+        # Track which version was restored from (ObjectId)
+        "restored_from_version": version_object_id,
```
ml-evs:
```suggestion
        "restored_from_version": version_object_id,  # Track which version was restored from (ObjectId)
```

webapp/src/components/DynamicDataTable.vue
```diff
@@ -126,8 +126,19 @@
...
-                <span class="ml-1">{{ slotProps.option.display_name }}</span>
+                <UserBubble
+                  v-if="slotProps.option.type === 'creator'"
+                  :creator="slotProps.option"
+                  :size="24"
+                />
+                <FormattedGroupName
+                  v-if="slotProps.option.type === 'group'"
+                  :group="slotProps.option"
+                  :size="20"
```
ml-evs:
```suggestion
                  :size="24"
```

webapp/src/components/DynamicDataTable.vue
```diff
@@ -138,7 +149,11 @@
                     :key="index"
                     class="inline-flex items-center mr-2"
                   >
-                    <UserBubble :creator="option" :size="20" />
+                    <UserBubble v-if="option.type === 'creator'" :creator="option" :size="20" />
+                    <FormattedGroupName v-if="option.type === 'group'" :group="option" :size="16" />
```
ml-evs:
```suggestion
                    <FormattedGroupName v-if="option.type === 'group'" :group="option" :size="20" />
```

ml-evs (approved):
Looks great, thanks @BenjaminCharmes! I just went through and changed the group bubble size to be the same as the users in the filter UI too as it looked a bit weird to me.

PR #1532, opened 2026-01-14
Add NMR compatibility with JEOL .jdx files

### LINKED ISSUE #1204: Enable other NMR formats within nmrglue
nmrglue already supports data from jeol, jcamp-dx etc, but we only expose the bruker zip support. It should be straightforward enough to enable this.

isabelcooley:
Edits to the NMR block and NMRBlock.vue which allow JEOL .jdx files to be read, spectra shown and metadata parsed and displayed. Missing metadata for probe and cnst31, added a metadata entry for jeol for spectral window, and jeol files show relaxation delay instead of recycle delay.

Closes #1204


pydatalab/src/pydatalab/apps/nmr/utils.py
```diff
@@ -124,9 +123,56 @@ def read_jcamp_dx_1d(filename: str | Path) -> tuple[pd.DataFrame, dict, str, tup
...
+    uc = ng.fileiobase.uc_from_udic(udic)
+
+    if udic[0]['time'] == True:
+        data = ng.process.proc_base.fft(data)
+        udic[0]['time'] = False
+        udic[0]['freq'] = True
+
+    data = ng.process.proc_base.rev(data)
+    data = ng.process.proc_autophase.autops(data,'acme')
+    data = ng.process.proc_base.di(data)
```
ml-evs: We don't currently support processing the raw FID in the other parsers (though have been considering it), does this work well enough for your purposes?

We may just want to emit a little warning to the user that the processing has done its best automatically and that the raw data should probably be processed in software from the spectrometer first instead.

ml-evs (commented):
Thanks for this @isabelcooley, think we can get this merged soon! Just a couple of comments below, the main one being about whether the FFT/phasing procedure is necessary from your end, or if we should exclude it as we do for other formats.

ml-evs (approved):

PR #1535, opened 2026-01-15
Fix `/save-item` to preserve private collections

### LINKED ISSUE #842: Unable to save item when creator if not creator of subcollection
Reported by @MJCliffe, there's an issue where users are unable to edit items if that item is part of a (potentially) private collection.

We should go back to scrubbing collection data in the save-item endpoint and develop something bespoke (a la #828).

BenjaminCharmes:
Closes #842

Users were unable to save items in private collections because save-item validated all collections and failed on inaccessible ones.

Solution: Implement a merge strategy that preserves inaccessible collections while allowing users to modify accessible ones.

Changes:
- Add `_scrub_collections_for_save()` to filter collections by permissions
- Merge accessible (user modifications) + inaccessible (preserved) collections
- Update tests to verify merge behavior

Behavior:
- Users can add/remove accessible collections
- Private collections are automatically preserved
- No permission errors or data loss

### DIFF
--- a/pydatalab/src/pydatalab/export.py
+++ b/pydatalab/src/pydatalab/export.py
@@ -16,7 +16,6 @@ from pydatalab.logger import LOGGER
 from pydatalab.models import ITEM_MODELS
 from pydatalab.mongo import flask_mongo
 from pydatalab.routes.v0_1.items import (
-    collections_lookup,
     creators_lookup,
     files_lookup,
     groups_lookup,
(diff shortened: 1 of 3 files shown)


ml-evs (approved):
Thanks @BenjaminCharmes! I pushed an extra change that does collection permissions lookup to avoid over-returning collections in get requests.

This is still quite cumbersome, so I'm going to open an issue that we should eventually move all collections management to purely the collections endpoints, which can then be used in the app directly.

PR #1536, opened 2026-01-16
Adding "substance" information to samples

ml-evs:
This PR adds the `HasSubstanceInfo` trait for use in items that have chemical information to attach to them. The model follows roughly what we already had for starting materials, with a few expansions, but all of this should be discussed before merging.

Previously, it also had some UI prototyping, which now lives in a different PR: #1728 

The component is currently vibe-coded to give a custom collapsible view that shows, when provided, the "best" guess of the substance:
- If SMILES/Inchi/Inchikey are provided, it renders a molecular structure (currently using pubchem)
- If a chemical formula is provided, show that
- If hazard codes are provided, show pictograms
etc.

When not editing, these are nicely rendered in a compact component. The "edit" mode is then just the normal input string boxes:

Edit mode (kinda gross):
2026-01-15-143154_915x949_scrot

View mode (not too bad)
2026-01-15-143147_549x252_scrot

It would be really nice to have a "small" mode of the view component, which we can maybe use in the item table to do a hover row -> show chemical info lookup.

### DIFF
--- a/pydatalab/schemas/sample.json
+++ b/pydatalab/schemas/sample.json
@@ -3,6 +3,49 @@
   "description": "A model for representing an experimental sample.",
   "type": "object",
   "properties": {
+    "chemform": {
+      "title": "Chemform",
+      "example": [
+        "Na3P",
+        "Na<sub>3</sub>P",
+        "LiNiO2@C",
+        "Na3+xP",
+        "LiNi1/3Co0.1Mn0.1O2"
+      ],
+      "type": "string"
+    },
+    "smiles": {
+      "title": "Smiles",
+      "aliases": [
+        "SMILES",
+        "smiles_representation"
+      ],
+      "type": "string"
+    },
+    "inchi": {
+      "title": "Inchi",
+      "type": "string"
+    },
+    "inchi_key": {
+      "title": "Inchi Key",
+      "type": "string"
+    },
+    "GHS_codes": {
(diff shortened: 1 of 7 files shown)


ml-evs:
Discussed with @BenjaminCharmes about taking over the webapp part of this PR, I may still decide to merge the Python models bit first but otherwise I think this is a decent starting point to work from.

BenjaminCharmes:
Hey @ml-evs ! Everything should be good now, can't request your review since it's your own PR. Let me know if you want me to change anything

ml-evs:
Looks really nice, thanks @BenjaminCharmes! Here's my plan of attack with this:

- I will pick the UI commits into a separate branch and let you create the PR
- I will finalise the pydantic side and API (I think I just want to add CAS), and make sure it also gets returned for starting materials

Then perhaps you could (can add this as a review to the follow up PR):

- propagate the substance component to the starting materials page and table
- remove the now-duplicated box for chemical formula from samples
- perhaps replace the hover-only column with an icon in the formula column? i.e., `formula = [icon] <chemical_formula>`?
- maybe do a few tweaks to the spacing of the component 
	- slightly bigger when not in edit mode
	- line up the molar mass input box with the rest

ml-evs:
I just pushed bc/substance-info-ui if you want to make your PR from there -- okay this PR is now just the Python migration side.

PR #1539, opened 2026-01-19
Improve error handling for block updates

### LINKED ISSUE #1488: Better handling of unhandled server errors in UI from update block
The `updateBlockFromServer` function in `server_fetch_utils.js` should be updated to handle unknown errors more gracefully, e.g., if the request times out without returning any data. I think basically we should passthrough most errors rather than silently failing.

BenjaminCharmes:
Closes #1488

### Changes
- Modified `updateBlockFromServer` to propagate errors instead of silently failing
- Added `.catch()` handlers in all components calling `updateBlockFromServer` to prevent uncaught errors
- Errors are now properly displayed to users while being logged to console


ml-evs (commented):
Could you give an example of the kind of errors that get caught here? The console logs are stripped in production, so its hard to say if we should do this (or use dialog service for unhandled errors).

ml-evs (approved):

PR #1549, opened 2026-01-22
Prevent repeated clicks on collection export button

### LINKED ISSUE #1547: Prevent repeated click on the collection export button
Probably one for @BenjaminCharmes . Right now its possible to spam click the export button for collections, wihtout getting a huge amount of feedback, until it suddenly gives you loads of zip files and dialogs.

Probably the button should be disabled whilst an export is happening, and there should be a dialog telling the user that its in progress before giving them the download.

BenjaminCharmes:
Closes #1547

**Changes:**
- Add "Export in Progress" dialog immediately when export starts
- Change completion dialog from info to success type
- Automatically close progress dialog before showing success dialog
- Only close progress dialog if one is currently displayed (prevents issues with fast exports)

### DIFF
--- a/webapp/src/components/ExportButton.vue
+++ b/webapp/src/components/ExportButton.vue
@@ -68,6 +68,13 @@ export default {
         }
 
         const taskId = response.task_id;
+
+        DialogService.alert({
+          title: "Export in Progress",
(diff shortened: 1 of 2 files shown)


ml-evs (approved):

Just a note that I'm seeing some weird behaviour on my dev server where collections have unsaved changes on first opening which conflicts a bit with this, but this seems unrelated to the changes here.

Issue #1551, opened 2026-01-22
Separate inventory and equipment by user groups


jdbocarsly:
In order to better support multiple research groups using a single Datalab deployment, we need a clean way to separate the inventories of different groups. 

From discussion with @ml-evs , the cleanest way to do this is probably to allow inventory to be assigned to one or more user groups. Then, users in those groups should be preferentially shown the inventory / equipment from groups they are a part of, especially in the ItemSelect. 

Equipment or Starting materials with no user groups can probably be shown to everyone.

Note: There are likely also cases when groups may want to be able to see the inventory of other groups, so this should still be an option if desired by the deployment. For example, cheminventory allows searching for chemicals across departments so that users can ask to borrow chemicals from other labs.

PR #1566, opened 2026-01-28
Add group-based inventory separation for equipment and starting materials

BenjaminCharmes:
Implements #1551

### Changes
- **Backend**: Apply group permissions to equipment/starting_materials endpoints
- **Frontend**: Add group selection to CreateEquipmentModal
- **Frontend**: Add ToggleableGroupsFormGroup to inventory edit pages

### Behavior
- Items with no groups → accessible to all users
- Items with groups → accessible only to group members

### Questions

- Is it the expected behavior that a user can share items with groups they are not a member of?
- “Maintainers” column in the equipment datatable is currently empty because the /equipment/ endpoints do not perform the MongoDB $lookup to populate creators and groups objects from their IDs. Is this intentional, or should these endpoints include the lookup?

### DIFF
--- a/pydatalab/src/pydatalab/permissions.py
+++ b/pydatalab/src/pydatalab/permissions.py
@@ -183,6 +183,15 @@ def _get_base_permissions(
             {"creator_ids": {"$exists": False}},
         ]
     }
+
+    # Define permission for inventory items with no groups (accessible to all)
+    null_group_perm = {
+        "$or": [
+            {"group_ids": {"$size": 0}},
+            {"group_ids": {"$exists": False}},
+        ]
+    }
+
     if current_user.is_authenticated and current_user.person is not None:
         # find managed users under the given user (can later be expanded to groups)
         managed_users = list(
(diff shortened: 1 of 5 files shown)


davidwaroquiers:
Hi,

Very nice that this is ongoing. We were planning to look into that and make a similar change. One thing here. We think that the default behavior should be the reverse (i.e. should be the same as for Samples and Cells):
Items with no groups => only the creator can access
Users part of an added group can access.

We had such requests.

One idea could be that there is a configurable option to switch this. (now this seems like a very tiny option for a small feature, and at some point we'll need to think about how to "organize" the options if pile up more and more customizable options, but that's probably for another discussion).

If the above is interesting (having a configurable option to switch on or off the "shared with all when no group"), we can make a small change of this PR to add it.

PeterKraus:
I would certainly prefer to have Samples private by default (creators only) and Starting Materials public by default, but I also completely understand @davidwaroquiers point about consistency above.

I don't know about the back end, but on the front end, it currently looks like this:
image

Perhaps these two "optional" widgets could be behind a "Make Starting Material Public" checkbox with a configurable default state?

davidwaroquiers:
> I would certainly prefer to have Samples private by default (creators only) and Starting Materials public by default, but I also completely understand @davidwaroquiers point about consistency above.
> 
> I don't know about the back end, but on the front end, it currently looks like this: image
> 
> Perhaps these two "optional" widgets could be behind a "Make Starting Material Public" checkbox with a configurable default state?

I think something like that would be reasonable indeed.

Is there any blocker for this PR ?

ml-evs:
Note to self: check that the group select works with only the groups the user is in

ml-evs:
> Note to self: check that the group select works with only the groups the user is in

Believe this is now ready for review

PR #1641, opened 2026-03-04
Selectable axes bokeh plot unit conversion features

jdbocarsly:
This branch is being co-developed along with the seppya plugin, and adds features to the basic bokeh plot to allow different units on the axes. Currently keeping this in draft, but some of this is probably useful to commit to core datalab as it will allow other blocks to use unit conversions.

### DIFF
--- a/pydatalab/pyproject.toml
+++ b/pydatalab/pyproject.toml
@@ -28,6 +28,7 @@ dependencies = [
     "pandas[excel] ~= 2.2",
     "pymongo ~= 4.7",
     "deepdiff ~= 8.1",
+    "seppya",
 ]
 
 [project.urls]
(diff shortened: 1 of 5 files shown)


pydatalab/pyproject.toml
```diff
@@ -137,6 +138,7 @@ dev = [
 [tool.uv.sources]
 datalab-app-plugin-insitu = { git = "https://github.com/datalab-org/datalab-app-plugin-insitu.git", rev = "v0.3.3" }
 nmrglue = { git = "https://github.com/jjhelmus/nmrglue.git", rev = "b6408751bf18801a0a4ce084acf4d31335e2b02a" }  # TODO: Update when new release is made
+seppya = { path = "./plugins/sepPya", editable = true }
```
ml-evs, line 141: May now be useful to you... https://docs.datalab-org.io/en/latest/plugins/#installing-plugins

ml-evs (commented):
Just a quick comment letting you know about the new plugin installation method -- how do we think these other changes interact with bokeh 3? We should likely try to make the jump soon (and use the opportunity to make a proper plotting module that abstracts away a lot of bokeh stuff, and splits up our giant selectable_axis_plot...)

pydatalab/pyproject.toml
jdbocarsly: > how do we think these other changes interact with bokeh 3?

No idea... but I think, in general this amount of customjs is going to pose a maintenance challenge. Its especially a concern that  I was thinking that maybe we should instead push the unit conversions to python, and use the event system to efficiently update bokeh plots. Ideally we could make this fairly easy/automatic on for the block developers. thoughts?

PR #1659, opened 2026-03-23
Add blocks as nodes to item graph and optimise rendering

### LINKED ISSUE #1083: Show blocks in the item graph

ml-evs:
Closes #1083 

Adds blocks to the item graph, toggleable

Also changes the way animations work for the graph -- instead the layout is computed and only then is it displayed, rather than animating each frame.

This PR got missed a few weeks ago before the release -- as well as introducing the blocks to the graph (now disabled by default) it also adds significant performance improvements by "giving up" at computing complex layouts after a timeout.

### DIFF
--- a/webapp/src/components/ItemRelationshipVisualization.vue
+++ b/webapp/src/components/ItemRelationshipVisualization.vue
@@ -23,7 +23,7 @@
       style="height: 200px; width: 100%; border: 1px solid transparent; border-radius: 5px"
       :default-graph-style="'elk-stress'"
       :show-options="false"
-      :default-show-blocks="true"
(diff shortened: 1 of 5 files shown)


webapp/src/components/ItemRelationshipVisualization.vue
```diff
@@ -23,6 +23,7 @@
       style="height: 200px; width: 100%; border: 1px solid transparent; border-radius: 5px"
       :default-graph-style="'elk-stress'"
       :show-options="false"
+      :default-show-blocks="true"
```
ml-evs:
```suggestion
      :default-show-blocks="false"
```

PR #1663, opened 2026-03-27
Use BDF parquet files when caching parsed echem

be-smith:
Changes the echem cacheing file from a .pkl to a .bdf.parquet file.

Fixes a bug where the cache wasn't being loaded for the single file mode, instead reloading from the raw file each time.

Also bumps navani to v0.1.17 so neware excel files are supported by the echem block

### DIFF
--- a/pydatalab/pyproject.toml
+++ b/pydatalab/pyproject.toml
@@ -68,6 +68,7 @@ apps = [
     "nmrglue ~= 0.12",
     # Electrochemistry
     "navani ~=0.1.16",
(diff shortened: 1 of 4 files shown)


ml-evs (dismissed):
Thanks @be-smith, pushed some minor fixes in [b692d14](https://github.com/datalab-org/datalab/pull/1663/commits/b692d14babf0aa2566cc89fa5691f7915114e7eb) and will bump this temporarily to the development version of navani to test the faster parsing, then we're good to go!

ml-evs (approved):

PR #1694, opened 2026-04-24
Bes/echem block parameter extraction

be-smith:


pydatalab/src/pydatalab/apps/echem/blocks.py
```diff
@@ -435,6 +443,21 @@ def plot_cycle(self):
                         cycle_summary_df["discharge capacity (mAh)"] / characteristic_mass_g
                     )
 
+            if electrode_area_cm2:
+                raw_df["capacity (mAh/cm2)"] = raw_df["capacity (mAh)"] / electrode_area_cm2
+                raw_df["current (mA/cm2)"] = raw_df["current (mA)"] / electrode_area_cm2
```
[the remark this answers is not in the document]
be-smith: Pydantic enforces it's a float > 0

webapp/src/components/CellInformation.vue
```diff
@@ -78,7 +78,17 @@
...
             />
           </div>
-          <div class="form-group col-lg-4 col-md-4 pr-3">
+          <div class="form-group col-lg-3 col-md-4 pr-3">
+            <label for="cell-electrode-area">Electrode area (cm²)</label>
+            <input
+              id="cell-electrode-area"
+              v-model="ElectrodeArea"
+              class="form-control"
+              type="text"
```
ml-evs, line 87:
```suggestion
              type="number"
```

I think

pydatalab/src/pydatalab/apps/echem/models.py
```diff
@@ -0,0 +1,34 @@
...
+
+    final_capacity_mAh_g: float | None = None
+    """Discharge capacity of the last cycle in mAh/g (only set when characteristic mass is available)."""
+
+    final_capacity_mAh_cm2: float | None = None
+    """Discharge capacity of the last cycle in mAh/cm² (only set when electrode area is available)."""
+
+    class Config:
+        extra = "forbid"
+
```
ml-evs, line 26: Some other suggestions we could include:

- initial capacity (or alternatively max capacity)
- number of cycles to reach 90% capacity (after max or initial cap) (also 95%, 80%, etc. and other magic values)
- energy (Wh) and specific energy (Wh/g)

pydatalab/src/pydatalab/apps/echem/blocks.py
```diff
@@ -510,6 +532,49 @@ def plot_cycle(self):
             )
         return
 
+    def _extract_cycle_parameters(
+        self,
+        cycle_summary_df: pd.DataFrame,
+        characteristic_mass_g: float | None,
+        electrode_area_cm2: float | None,
+    ) -> None:
+        """Extract summary parameters from the cycle summary DataFrame and store in self.data["computed"].
```
ml-evs, line 541: Do we need some guard for "is this cycling data" before try to compute these things? We might want to separate each metric (or sensible group of metrics) out into its own method so we try to compute each one and set it to `None` if not possible. Probably for this we'd need to analyse the cycling protocol and decide if it fits what we expect roughly, or if its something weird (so that e.g., we dont apply these metrics to things like rate tests)

ml-evs (commented):
Couple of extra comments from thinking after our chat @be-smith

PR #1696, opened 2026-04-28
Add cyclic voltammetry (CV) block

be-smith:
Adds a CV block to datalab that can handle .mpr CV files.

Makes changes to selectable_axes_plot to handle colouring by a numeric index, e.g cycle number rather than discretely colouring based on filename - this should be reviewed as can have impacts on other blocks that use this function. From testing has no impact on other blocks.

Also adds the functionality to hide traces by clicking on the legend by default (selectable_axes_plot). This effects:

- UV-Vis (positive change)
- XRD (doesn't really play nicely with plotting the lines and the markers, but the line plot disappears)
- Raman (no option to plot multiple files)
- TGA/Mass spec (no option to plot multiple files)
- NMR (Seems to only plot one file - may have to check further)

Resolves #1654


pydatalab/src/pydatalab/apps/cv/utils.py
```diff
@@ -0,0 +1,122 @@
...
+    for cycle_num, start in enumerate(reversals):
+        half_cycle[start:] = cycle_num + 1
+
+    return half_cycle
+
+
+def parse_cv_mpr(filename: Path) -> pd.DataFrame:
+    """Parses .mpr CV data generated by a biologic potentiostat.
+
+    Returns a dataframe containing Potential (V), Current (mA), and Cycle number columns.
```
ml-evs, line 91: Only issue I can see here is that we're not looking in the mpr file to see if it is actually CV data, we're just trying to open any CV file as if it were. Might be worth looking for the specific tags for this?

ml-evs (approved):

pydatalab/src/pydatalab/apps/cv/utils.py
ml-evs: Unresolving this just so you see it

PR #1727, opened 2026-05-14
Fixing duplication of parthood relationships for cells and samples 

wuppersaver:
Fix duplicated parthood relationships from refcode/item_id dedup mismatch

The `add_missing_electrode_relationships` and `add_missing_synthesis_relationships` validators deduplicated relationships against constituents using a single `refcode or item_id` token, evaluated independently on each side. Because `entry_reference_lookup` enriches constituents with their refcode at read time but leaves the stored `relationships` rows untouched, the stored side keys on `item_id` while the enriched side keys on `refcode` — the two never compare equal, so a duplicate parthood relationship is appended on every read and persisted on every write.

Both validators now index existing relationships by refcode *and* item_id (a dict `identifier -> relationship`), match a constituent if *either* identifier is found, and back-fill the missing identifier in place rather than appending a duplicate. This also fixes the latent reverse-asymmetry case and replaces the synthesis validator's create-then-delete back-fill side-effect with an explicit one.

Adds `test_cell_relationship_deduplication` covering the forward/reverse asymmetry and idempotency.

### DIFF
--- a/pydatalab/src/pydatalab/models/cells.py
+++ b/pydatalab/src/pydatalab/models/cells.py
@@ -116,4 +116,11 @@ class Cell(Item):
                     relationship.refcode = relationship.refcode or refcode
                     relationship.item_id = relationship.item_id or item_id
 
+                # Register the relationship's identifiers in the index so a later
+                # constituent referencing the same entry matches it rather than
+                # appending a duplicate within this same validation pass.
+                for identifier in (relationship.refcode, relationship.item_id):
+                    if identifier:
+                        existing_parthood_relationships[identifier] = relationship
+
         return values
--- a/pydatalab/src/pydatalab/models/cells.py
+++ b/pydatalab/src/pydatalab/models/cells.py
@@ -77,30 +77,43 @@ class Cell(Item):
         """Add any missing sample synthesis constituents to parent relationships"""
         from pydatalab.models.relationships import RelationshipType, TypedRelationship
 
-        existing_parthood_relationship_ids = set()
+        existing_parthood_relationships = {}
         if values.get("relationships") is not None:
-            existing_parthood_relationship_ids = {
-                relationship.refcode or relationship.item_id
+            # Index by refcode *and* item_id so a stored relationship carrying an
+            # item_id still matches a refcode-enriched constituent (and vice-versa).
+            existing_parthood_relationships = {
+                identifier: relationship
                 for relationship in values["relationships"]
                 if relationship.relation == RelationshipType.PARTHOOD
+                for identifier in (relationship.refcode, relationship.item_id)
+                if identifier
             }
         else:
             values["relationships"] = []
 
         for component in ("positive_electrode", "negative_electrode", "electrolyte"):
             for constituent in values.get(component, []):
-                if (
-                    isinstance(constituent.item, EntryReference)
-                    and (constituent.item.refcode or constituent.item.item_id)
-                    not in existing_parthood_relationship_ids
-                ):
+                if not isinstance(constituent.item, EntryReference):
+                    continue
+
+                refcode = constituent.item.refcode
+                item_id = constituent.item.item_id
+
+                relationship = existing_parthood_relationships.get(
+                    refcode
+                ) or existing_parthood_relationships.get(item_id)
+                if relationship is None:
                     relationship = TypedRelationship(
                         relation=RelationshipType.PARTHOOD,
-                        refcode=constituent.item.refcode,
-                        item_id=constituent.item.item_id,
+                        refcode=refcode,
+                        item_id=item_id,
                         type=constituent.item.type,
                         description="Is a constituent of",
                     )
                     values["relationships"].append(relationship)
+                else:
+                    # Back-fill any identifier missing from the stored relationship
+                    relationship.refcode = relationship.refcode or refcode
+                    relationship.item_id = relationship.item_id or item_id
 
         return values
--- a/pydatalab/src/pydatalab/models/cells.py
+++ b/pydatalab/src/pydatalab/models/cells.py
@@ -74,7 +74,7 @@ class Cell(Item):
 
     @root_validator
     def add_missing_electrode_relationships(cls, values):
-        """Add any missing sample synthesis constituents to parent relationships"""
+        """Add any missing cell component constituents to parent relationships"""
         from pydatalab.models.relationships import RelationshipType, TypedRelationship
 
         existing_parthood_relationships = {}
--- a/pydatalab/src/pydatalab/models/traits.py
+++ b/pydatalab/src/pydatalab/models/traits.py
@@ -106,12 +106,16 @@ class HasSynthesisInfo(BaseModel):
 
         constituents_set = set()
         if values.get("synthesis_constituents") is not None:
-            existing_parent_relationship_ids = set()
+            existing_parent_relationships = {}
             if values.get("relationships") is not None:
-                existing_parent_relationship_ids = {
-                    relationship.refcode or relationship.item_id
+                # Index by refcode *and* item_id so a stored relationship carrying an
+                # item_id still matches a refcode-enriched constituent (and vice-versa).
+                existing_parent_relationships = {
+                    identifier: relationship
                     for relationship in values["relationships"]
                     if relationship.relation == RelationshipType.PARENT
+                    for identifier in (relationship.refcode, relationship.item_id)
+                    if identifier
                 }
             else:
                 values["relationships"] = []
@@ -121,20 +125,28 @@ class HasSynthesisInfo(BaseModel):
                 if isinstance(constituent.item, InlineSubstance):
                     continue
 
-                constituent_id = constituent.item.refcode or constituent.item.item_id
+                refcode = constituent.item.refcode
+                item_id = constituent.item.item_id
 
-                if constituent_id not in existing_parent_relationship_ids:
+                relationship = existing_parent_relationships.get(
+                    refcode
+                ) or existing_parent_relationships.get(item_id)
+                if relationship is None:
                     relationship = TypedRelationship(
                         relation=RelationshipType.PARENT,
-                        refcode=constituent.item.refcode,
-                        item_id=constituent.item.item_id,
+                        refcode=refcode,
+                        item_id=item_id,
                         type=constituent.item.type,
                         description="Is a constituent of",
                     )
                     values["relationships"].append(relationship)
+                else:
+                    # Back-fill any identifier missing from the stored relationship
+                    relationship.refcode = relationship.refcode or refcode
+                    relationship.item_id = relationship.item_id or item_id
 
                 # Accumulate all constituent IDs in a set to filter those that have been deleted
-                constituents_set.add(constituent_id)
+                constituents_set.update(i for i in (refcode, item_id) if i)
 
         # Finally, filter out any parent relationships with item that were removed
         # from the synthesis constituents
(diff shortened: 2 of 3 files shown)


wuppersaver:
More detailed description and what was done to mitigate below.

# Fixing the duplicated parthood relationships

A detailed walk-through of the relationship-duplication bug in `pydatalab` and the fix applied in the fork. Companion to
[`relationships-duplication-explainer.md`](./relationships-duplication-explainer.md), which diagnosed the original cause.

The fix touches two pydantic validators:

- `Cell.add_missing_electrode_relationships` — `pydatalab/src/pydatalab/models/cells.py`
- `HasSynthesisInfo.add_missing_synthesis_relationships` — `pydatalab/src/pydatalab/models/traits.py`

---

## 1. The setup: two representations of the same constituent

A cell stored in MongoDB references its electrode constituents in two places that are meant to stay in sync — the `positive_electrode` / `negative_electrode`/ `electrolyte` arrays, and the `relationships` array.

When the cell was first saved (before refcodes were widely populated, or before the UI resolved them on entry), Mongo holds something like:

```jsonc
{
  "item_id": "2200",
  "type": "cells",
  "positive_electrode": [
    { "item": { "item_id": "bcaf-241co-ss", "type": "starting_materials" }, "quantity": 1 }
  ],
  "relationships": [
    {
      "relation": "is_part_of",
      "type": "starting_materials",
      "item_id": "bcaf-241co-ss",
      "refcode": null,
      "description": "Is a constituent of"
    }
  ]
}
```

Both sides carry **`item_id` only, `refcode: null`**. Internally consistent.

---

## 2. The trigger: `entry_reference_lookup` enriches one side but not the other

On every GET, `entry_reference_lookup` (in `routes/v0_1/items.py`) walks the electrode arrays and replaces each `item` subdocument with a fresh DB lookup that **includes the refcode**. It does **not** touch the `relationships` array.

So by the time the pydantic `Cell` model is reconstructed, the in-memory document looks like this:

```jsonc
{
  "positive_electrode": [
    {
      "item": {
        "item_id": "bcaf-241co-ss",
        "refcode": "datalab:JJSDGK",   // <-- refcode now present
        "type": "starting_materials",
        "name": "LCO"
      },
      "quantity": 1
    }
  ],
  "relationships": [
    {
      "relation": "is_part_of",
      "type": "starting_materials",
      "item_id": "bcaf-241co-ss",
      "refcode": null                  // <-- still null
    }
  ]
}
```

The constituent has been **refcode-enriched**; the stored relationship has **not**. That asymmetry is the whole bug.

---

## 3. Before: why it duplicated

The old `add_missing_electrode_relationships` deduplicated using a single token,
`refcode or item_id`, computed **independently on each side**.

### Step 1 — build the index of existing relationships

```python
existing_parthood_relationship_ids = {
    relationship.refcode or relationship.item_id   # <-- single token per relationship
    for relationship in values["relationships"]
    if relationship.relation == RelationshipType.PARTHOOD
}
```

The one stored relationship has `refcode=null, item_id="bcaf-241co-ss"`, so `null or "bcaf-241co-ss"` evaluates to `"bcaf-241co-ss"`:

```jsonc
existing_parthood_relationship_ids = ["bcaf-241co-ss"]   // keyed by item_id
```

### Step 2 — check the constituent against that index

```python
if (constituent.item.refcode or constituent.item.item_id)   # <-- single token, the OTHER side
        not in existing_parthood_relationship_ids:
    # ...append a new relationship...
```

The constituent now has a refcode, so `"datalab:JJSDGK" or "bcaf-241co-ss"` evaluates to `"datalab:JJSDGK"`:

```jsonc
"datalab:JJSDGK" not in ["bcaf-241co-ss"]   ->   true   ->   APPEND
```

The check asks *"is `datalab:JJSDGK` known?"* against an index that only knows `bcaf-241co-ss`. It isn't — so the validator concludes the constituent has no relationship yet and appends a second one:

```jsonc
"relationships": [
  { "relation": "is_part_of", "item_id": "bcaf-241co-ss", "refcode": null },            // original
  { "relation": "is_part_of", "item_id": "bcaf-241co-ss", "refcode": "datalab:JJSDGK" }  // appended duplicate
]
```

### The core failure in one sentence

> The index was keyed by the **stored** side's `refcode or item_id`  (→ `item_id`, because refcode was null), but the lookup used the **enriched**
> side's `refcode or item_id` (→ `refcode`). The two tokens describe the same  item but can never be equal, so dedup always misses and a duplicate is  appended.

`@root_validator` runs on every model construction — **read and write** — so the duplicate appeared in every API response, and each save through the API grew the stored list by one more row per electrode.

---

## 4. After: how it's mitigated

Two changes, applied identically to `cells.py` and `traits.py`.

### Change 1 — index by *both* identifiers, into a dict

```python
existing_parthood_relationships = {
    identifier: relationship
    for relationship in values["relationships"]
    if relationship.relation == RelationshipType.PARTHOOD
    for identifier in (relationship.refcode, relationship.item_id)   # <-- flatten BOTH
    if identifier                                                   # <-- drop nulls
}
```

The stored relationship contributes an entry for **every** identifier it
carries:

```jsonc
existing_parthood_relationships = {
  "bcaf-241co-ss": "<relationship object>"   // only item_id, since refcode was null
}
```

If a relationship had both identifiers, it would appear under both keys — all
pointing at the *same* object.

### Change 2 — look up by *either* identifier, then back-fill

```python
refcode = constituent.item.refcode      # "datalab:JJSDGK"
item_id = constituent.item.item_id      # "bcaf-241co-ss"

relationship = existing_parthood_relationships.get(refcode) \
    or existing_parthood_relationships.get(item_id)
```

- `.get("datalab:JJSDGK")` → `None` (refcode isn't an index key — the stored row had none)
- `.get("bcaf-241co-ss")` → **the existing relationship object**

The `or` means *"match on refcode if you can, otherwise fall back to item_id."*
Either identifier finds the row.

```python
if relationship is None:
    # genuinely new constituent — append a fresh relationship
    ...
else:
    # already represented — back-fill whatever identifier is missing
    relationship.refcode = relationship.refcode or refcode    # null  -> "datalab:JJSDGK"
    relationship.item_id = relationship.item_id or item_id    # stays "bcaf-241co-ss"
```

No duplicate is appended. Instead the **existing** row is upgraded in place:

```jsonc
"relationships": [
  { "relation": "is_part_of", "item_id": "bcaf-241co-ss", "refcode": "datalab:JJSDGK" }  // one row, both ids
]
```

### Why this is symmetric

Because both the index *and* the lookup now consider **both** identifiers, there is no longer a "stored side keys on A, enriched side keys on B" mismatch — any shared identifier is enough to recognise the same item.

| Scenario | stored row | constituent | old result | new result |
|---|---|---|---|---|
| **Forward** (the observed bug) | `{item_id: X, refcode: null}` | `{item_id: X, refcode: R}` | 2 rows | 1 row `{X, R}` — refcode back-filled |
| **Reverse** (latent) | `{item_id: null, refcode: R}` | `{item_id: X, refcode: R}` | refcode silently dropped | 1 row `{X, R}` — item_id back-filled |
| **Already clean** | `{item_id: X, refcode: R}` | `{item_id: X, refcode: R}` | matched | matched; back-fill is a no-op — idempotent |
| **Genuinely new** | *(no relationship for this item)* | `{item_id: Y, refcode: S}` | appended | appended (unchanged) |


---

## 5. The `traits.py` extra wrinkle: the cleanup filter

`add_missing_synthesis_relationships` also has a trailing filter that removes parent relationships whose constituent was *deleted* from `synthesis_constituents`. The old version had the **same single-token bug** there:

```python
# OLD — single token on both sides again
constituents_set.add(constituent.item.refcode or constituent.item.item_id)
...
if (rel.refcode or rel.item_id) not in constituents_set ...   # filter
```

This is actually what *masked* the duplication for samples: the buggy add-check created a refcode-bearing duplicate, and the buggy filter happened to delete the refcode-null original — a create-then-delete dance that accidentally looked like
a clean back-fill. It still carried the latent reverse-asymmetry data-loss bug, so it got the same treatment:

```python
# NEW — both identifiers tracked, both tested
constituents_set.update(i for i in (refcode, item_id) if i)
...
if rel.refcode not in constituents_set
   and rel.item_id not in constituents_set ...   # stale only if NEITHER matches
```

Now the back-fill is done **explicitly** (Change 2 above) rather than as a
side-effect of the filter, and the filter does only its real job: dropping
relationships for constituents that were actually removed.

---

## 6. The regression test

A new test, `test_cell_relationship_deduplication`, was added to
`pydatalab/tests/test_models.py`. Pure-model tests don't run
`entry_reference_lookup`, so the test **simulates the enrichment asymmetry by
hand**: it constructs a `Cell` whose `relationships` array and
`positive_electrode` constituent deliberately disagree on which identifiers they
carry, then asserts the validator reconciles them into a single, fully-populated
relationship.

It covers three cases:

### Case 1 — forward asymmetry (the observed bug)

Stored relationship carries `item_id` only; the constituent has been
refcode-enriched.

```jsonc
"relationships":      [ { "relation": "is_part_of", "item_id": "test_cathode" } ]
"positive_electrode": [ { "item": { "item_id": "test_cathode", "refcode": "grey:ABCDEF" } } ]
```

```python
assert len(parthood) == 1
assert parthood[0].refcode == "grey:ABCDEF"   # refcode back-filled from the constituent
assert parthood[0].item_id == "test_cathode"
```

### Case 2 — reverse asymmetry (the latent bug)

Stored relationship carries the `refcode`; the constituent has `item_id` only.

```jsonc
"relationships":      [ { "relation": "is_part_of", "item_id": "test_cathode", "refcode": "grey:ABCDEF" } ]
"positive_electrode": [ { "item": { "item_id": "test_cathode" } } ]
```

```python
assert len(parthood) == 1
assert parthood[0].refcode == "grey:ABCDEF"
assert parthood[0].item_id == "test_cathode"   # item_id back-filled from the constituent
```

### Case 3 — idempotency

Re-validating an already-clean cell (via a `json` round-trip) must not grow the
list or drop identifiers.

```python
cell = Cell(**json.loads(cell.json()))
assert len(parthood) == 1
assert parthood[0].refcode == "grey:ABCDEF"
assert parthood[0].item_id == "test_cathode"
```

The pre-existing `test_cell_with_inlined_reference` and the server-side
`test_new_sample_with_relationships` continue to pass unchanged — the latter
is what pins the back-fill behaviour for the synthesis validator.

---

## 7. Scope and limitations

- **Fix-only.** The validators stop *new* duplicates from being created and
  back-fill missing identifiers. Documents that *already* carry a duplicate pair
  in Mongo are not collapsed — the validator recognises the constituent as
  present and adds nothing; for samples it back-fills the first match but leaves
  the second. Healing pre-existing duplicates would need a separate one-shot
  migration.
- **Verified** against the full `pydatalab` test suite plus the new
  `test_cell_relationship_deduplication` regression test.

ml-evs:
Hi @wuppersaver, this is a long-standing bug that we have to be quite careful to iron out, with a few existing PRs. #1430 is the main fix that prevents this from happening when updating constituents via the API, but I've been working on a simplified version that will also do the relevant lookup when saving. This is one of two blockers for the next 0.7.0 release that I'm hoping to get out this week.

Whilst I'm sure these changes are fine, there's quite a lot to unpick here, so I'd prefer to just fix this myself (and maybe steal your test case for validation).

wuppersaver:
My bad - I should have checked the issue tracker better. At least it was not futile if the new test helps to check for the correct behaviour. 😄

ml-evs (dismissed):
We merged one of the underlying fixes to prevent this happening in #1755, but I think its also worth merging this fix and the test, plus some associated tests I've transplanted from #1430 and added myself. Thanks for spotting this @wuppersaver!

pydatalab/src/pydatalab/models/cells.py
```diff
@@ -77,30 +77,50 @@ def add_missing_electrode_relationships(cls, values):
         """Add any missing sample synthesis constituents to parent relationships"""
```
ml-evs:
```suggestion
        """Add any missing cell component constituents to parent relationships"""
```

ml-evs (approved):

PR #1760, opened 2026-05-21
Update echem block to better handle bdf exports and caching

be-smith:
Updated navani to 0.1.20
  
  - Bumped navani to 0.1.20
  - BDF export is now split into `build_bdf_df` (DataFrame construction and downcasting) and `save_bdf` (file writing). Previously export_to_bdf was called twice to write both parquet and CSV, doing the full
   DataFrame processing twice — now build_bdf_df is called once and both files are written in two separate save_bdf calls (to isolate an error to either call without preventing saving the other file)
  - CSV regeneration from a parquet cache hit no longer re-processes the raw DataFrame — it reads the parquet directly and calls save_bdf
  - Parquet and CSV save failures are now handled independently, so a parquet write failure no longer suppresses the CSV download link
  - Updated tests to cover each failure mode separately and test build_bdf_df/save_bdf directly

Currently cache is only regenerated for live streaming files, or if there is no cache.

### DIFF
--- a/pydatalab/pyproject.toml
+++ b/pydatalab/pyproject.toml
@@ -66,7 +66,7 @@ apps = [
     # NMR
     "nmrglue ~= 0.12",
     # Electrochemistry
-    "navani >= 0.1.19",
+    "navani >= 0.1.20",
     "pyarrow ~= 23.0.1",
     # Raman
     "pybaselines ~= 1.1",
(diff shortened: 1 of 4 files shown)


ml-evs (approved):
Looks good, thanks @be-smith!

PR #1775, opened 2026-05-28
Only write a new version when item data actually changes

### LINKED ISSUE #1680: Spurious item versions created when item is updated with no-op
e.g., when syncing cheminventory, `save-item` is called but the information remains the same; the item versioning process is currently minting a new version for this, but really it should skip it.

be-smith:
Before this change, every call to `/save-item/` or `/save-version/` unconditionally minted a new version snapshot, even when the item content was identical. This caused scripts like the cheminventory sync plugin to generate thousands of redundant versions on each run.

Changes:
- `save_version_snapshot` now compares the current item state against the most recent snapshot (excluding mechanical fields: `last_modified`, `version`, `_id`) and returns early if content is unchanged
- `save_item` no longer increments `item.version` unconditionally; it now only writes the counter-derived version number back to the item document when a snapshot is actually minted
- `_create_sample` likewise writes the counter-derived version number back to the item after creation, so `item.version` is set by the counter rather than the model default

Resolves #1680

### DIFF
--- a/pydatalab/src/pydatalab/routes/v0_1/items.py
+++ b/pydatalab/src/pydatalab/routes/v0_1/items.py
@@ -1622,9 +1622,6 @@ def save_item():
             400,
         )
 
-    # Increment version number on the item itself
-    updated_data["version"] = item.get("version", 0) + 1
-
     user_only = item["type"] not in ("starting_materials", "equipment")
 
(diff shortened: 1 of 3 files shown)


ml-evs (approved):
Tested with the cheminventory sync plugin, works well! Thanks @be-smith!

PR #1780, opened 2026-05-31
Enable echem block cache invalidation based on file upload timestamps

### LINKED ISSUE #1745: Async block reload  is not triggered when file is replaced
We need to improve the file hashing system (#1744) then make use of it to invalidate the block cache. As echem is the main consumer of this for now, I will fix it there first.

ml-evs:
This PR updates the single file mode of the echem block to allow the parquet cache to be invalidated if the underlying source file has changed. Ideally, we would store the hash at time of block creation and directly compare to the new hash, but for now we can just check the `mtime` of the cache vs the database last_modified time of the file.

Also does a bit of associated tidying to the echem block.

Closes #1745 for the one async block we have, for now...

### DIFF
--- a/pydatalab/src/pydatalab/apps/echem/blocks.py
+++ b/pydatalab/src/pydatalab/apps/echem/blocks.py
@@ -296,7 +296,7 @@ class CycleBlock(DataBlock):
         parquet_path = cache_location.with_name(cache_location.name + "_cached.bdf.parquet")
         csv_path = cache_location.with_name(cache_location.name + ".bdf.csv")
         return self._load_and_cache_echem(
-            cache_location, parquet_path, csv_path, reload, locations=locations
+            cache_location, parquet_path, csv_path, reload=True, locations=locations
         )
 
     @staticmethod
--- a/pydatalab/src/pydatalab/apps/echem/blocks.py
+++ b/pydatalab/src/pydatalab/apps/echem/blocks.py
@@ -237,7 +237,7 @@ class CycleBlock(DataBlock):
         return raw_df, csv_path
 
     def _load_single(self, file_id: ObjectId, reload: bool) -> tuple[pd.DataFrame, Path | None]:
-        """Parse a single echem file using navani and caches to disk.
+        """Parse a single echem file using navani and cache to disk.
 
         Returns the raw DataFrame and the BDF export path (or None if the source is already BDF
         or export failed).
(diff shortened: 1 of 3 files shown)


pydatalab/src/pydatalab/apps/echem/blocks.py
```diff
@@ -253,18 +252,26 @@ def _load_single(self, file_id: ObjectId, reload: bool) -> tuple[pd.DataFrame, P
...
             csv_path = location.with_name(f"{bare_stem}.bdf.csv")
             return self._load_and_cache_echem(location, parquet_path, csv_path, reload)
         if ext.startswith(".bdf"):
             # Other BDF formats: cache to parquet but don't write a redundant .bdf.csv.
             # bdf_url will fall back to linking the source file directly.
-            parquet_path = location.with_name(f"{bare_stem}_cached.bdf.parquet")
             return self._load_and_cache_echem(location, parquet_path, None, reload)
-        parquet_path = location.with_name(f"{location.stem}_cached.bdf.parquet")
+
         csv_path = location.with_name(f"{location.stem}.bdf.csv")
```
be-smith: I guess this should also use bare_stem rather than location.stem, to be consistent with the other csv paths
```suggestion
        csv_path = location.with_name(f"{bare_stem}.bdf.csv")
```

pydatalab/src/pydatalab/apps/echem/blocks.py
```diff
@@ -288,8 +295,21 @@ def _load_multi(
         cache_location = locations[0].parent / f"merged_{cache_key}"
         parquet_path = cache_location.with_name(cache_location.name + "_cached.bdf.parquet")
         csv_path = cache_location.with_name(cache_location.name + ".bdf.csv")
+
+        # Check cache age and invalidate based on the most recent source file modification time
+        reload = False
+        if parquet_path.exists():
```
be-smith: I don't think we should hard set it to false here. Will be confusing if we ever do pass reload=True to the _load and _load_multi functions.


```suggestion
        if not reload and parquet_path.exists():
```

be-smith (commented):
Just two minor changes that won't change functionality with the current use case.

pydatalab/src/pydatalab/apps/echem/blocks.py
```diff
@@ -238,7 +237,7 @@ def _load_and_cache_echem(
         return raw_df, csv_path
 
     def _load_single(self, file_id: ObjectId, reload: bool) -> tuple[pd.DataFrame, Path | None]:
-        """Parse a single echem file using navani, with pickle caching.
+        """Parse a single echem file using navani and caches to disk.
```
ml-evs:
```suggestion
        """Parse a single echem file using navani and cache to disk.
```

pydatalab/src/pydatalab/apps/echem/blocks.py
ml-evs: Good catch, thanks -- see #1781

PR #1797, opened 2026-06-01
Add an admin task for cleaning up old dereferenced files

ml-evs:
New task: `invoke admin.cleanup-files`.

Prints the file directories that can be removed, and an estimate of how much space would be freed. Up to the admin to then clear the folders and remove as they wish.

Also removes the deprecated import cheminventory task.


ml-evs:
pre-commit.ci autofix

ml-evs:
pre-commit.ci autofix

PR #1798, opened 2026-06-01
Handle block cleanup when gridfs file has been dereferenced

ml-evs:
Have been seeing some log entries where the gridfs bucket is not found (i.e., it has already been removed). This PR simply continues and cleans up the task.

We may need an extra clean up task (manual) to clear grid fs infrequently too.


ml-evs:
pre-commit.ci autofix

PR #1785, opened 2026-06-01
Add user agent and creator retrieval for item versions

ml-evs:
This pull request enhances item versioning by:

- using a series of different user-agents to decide whether a save was manual or "agent" scripted (e.g., python-api, beholder, cheminventory-sync)
- lookup user IDs on version retrieval so they can be rendered in the UI alongside the associated changes

Closes #1783 and partially #1784 (still needs to the UI)

PR #1806, opened 2026-06-02
Expose KNOWN_USER_AGENTS

be-smith:
Exposes KNOWN_USER_AGENTS at the info level so clients (e.g datalab-api) can validate themselves against it.

Builds on #1785, and ensures future external use cases can have their agent stored in the item_version info


ml-evs (commented):
Don't think we need this atm, but can document the feature better in the future at least

PR #1813, opened 2026-06-05
Da/auth corecting bugs 

DianaAliabieva:
Hello! There were some problems about auth. 

example while connecting to git: 
{
  "message": "KeyError (\u003CIdentityType.GOOGLE: 'google'\u003E,)",
  "status": "error",
  "title": "Internal Server Error"
}

the main problem was in main.py, because of the button. So the solution was to add a value while trying to connect to any type of auth. 

I am not at all sure that everything is fully solved. I have done a lot of tests, looks like everything is working with any tipe of auth, but maybe there can be something else that I skipped. 

Let me know if there is a job to be done

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -1,0 +1,62 @@
+name: datalab
+services:
+  app:
+    profiles: ["prod"]
+    build:
+      context: .
+      dockerfile: .docker/app/Dockerfile
+    volumes:
+      - ./logs:/logs
+    restart: unless-stopped
+    environment:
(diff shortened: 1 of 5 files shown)


ml-evs (commented):
Thanks for this @DianaAliabieva, as I messaged this bug was fixed in #1811 but thanks for doing the same. I don't think I want to add tests for this landing page, as really it should just be removed (and I'll do this at some point myself). 

I assume the other changes in this PR are accidental/unnecessary at the moment?

PR #1823, opened 2026-06-09
Da/graph blocks

DianaAliabieva:
Hello!

This Pull request is about:
There is a new block "Sample block" (represents a graph of blocks that were done with any Sample)
Sorted by default from the Latest one. 
Each graph block is clickable (moving to a block that was created)
There is a system of filtration, navigation system and pagination (10/20/50 graphs with a scrolling bar)
Each graph has a short type name of block, time when it was created and date and it has a "hover" that shows a full name of block, and some more information. In the future the plan is to upgrade it and to add more details. 

In backend The new field "created at" was added. 

 
photo_1_2026-06-09_15-55-22
photo_2_2026-06-09_15-55-22
photo_3_2026-06-09_15-55-22
photo_4_2026-06-09_15-55-22
photo_5_2026-06-09_15-55-22
photo_6_2026-06-09_15-55-22

Don't hesitate to add ideas, to modify or give propositions of what else could be done!

### DIFF
--- a/webapp/src/components/SampleInformation.vue
+++ b/webapp/src/components/SampleInformation.vue
@@ -63,6 +63,10 @@
 
     <TableOfContents :item_id="item_id" :information-sections="tableOfContentsSections" />
     <SynthesisInformation class="mt-3" :item_id="item_id" />
+
+    <div id="sample-stages" class="mt-3">
+      <SampleStagesTimeline :stages="sampleStages" @stage-click="scrollToBlock" />
+    </div>
   </div>
 </template>
 
@@ -78,6 +82,7 @@ import SynthesisInformation from "@/components/SynthesisInformation";
 import SubstanceInformation from "@/components/SubstanceInformation";
 import TableOfContents from "@/components/TableOfContents";
 import ItemRelationshipVisualization from "@/components/ItemRelationshipVisualization";
+import SampleStagesTimeline from "@/components/SampleStagesTimeline";
 
 export default {
   components: {
@@ -86,6 +91,7 @@ export default {
     SubstanceInformation,
     TableOfContents,
     ItemRelationshipVisualization,
+    SampleStagesTimeline,
     FormattedRefcode,
     ToggleableCollectionFormGroup,
     ToggleableCreatorsFormGroup,
@@ -106,6 +112,7 @@ export default {
         { title: "Sample Information", targetID: "sample-information" },
         { title: "Substance Information", targetID: "substance-information" },
         { title: "Synthesis Information", targetID: "synthesis-information" },
+        { title: "Sample Stages", targetID: "sample-stages" },
       ],
     };
   },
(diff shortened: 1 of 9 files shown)


webapp/src/components/SampleInformation.vue
```diff
@@ -128,6 +135,65 @@ export default {
     possibleItemStatuses() {
       return this.schema?.attributes?.schema?.definitions?.ItemStatus?.enum;
     },
+    sampleStages() {
```
jdbocarsly, line 138: It would be good if this logic could live inside the SampleStagesTimeline.vue component (or in its own utility file). The `SampleInformation` component is for fields and logic that is specific to Samples (compared to Cells, StartingMaterials, etc.). This could be useful for all item types, so I don't think it should be specific to Samples.

webapp/src/components/SampleInformation.vue
```diff
@@ -106,6 +112,7 @@ export default {
         { title: "Sample Information", targetID: "sample-information" },
         { title: "Substance Information", targetID: "substance-information" },
         { title: "Synthesis Information", targetID: "synthesis-information" },
+        { title: "Sample Stages", targetID: "sample-stages" },
```
jdbocarsly, line 115: Again, I don't think this component would be specific to `Samples`, so I would call it something like `Block history` or `Item History`.

webapp/src/components/SampleStagesTimeline.vue
```diff
@@ -0,0 +1,1329 @@
...
+import AngleDoubleLeftIcon from "@primevue/icons/angledoubleleft";
+import AngleDoubleRightIcon from "@primevue/icons/angledoubleright";
+import ArrowDownIcon from "@primevue/icons/arrowdown";
+import ArrowRightIcon from "@primevue/icons/angleright";
+import DatePicker from "primevue/datepicker";
+import ChevronLeftIcon from "@primevue/icons/chevronleft";
+import ChevronRightIcon from "@primevue/icons/chevronright";
+import FilterIcon from "@primevue/icons/filter";
+import Select from "primevue/select";
+import "primeicons/primeicons.css";
```
jdbocarsly, line 268: The styling of this component is nice, but somewhat inconsistent with the style choices datalab overall. For better or for worse, datalab has been built with consistent styling using:

-  native HTML inputs wherever possible
- styled with Bootstrap for CSS
-  fontawesome for icons
- vue-select for multiselects
-  a variety of lightweight custom components we've written over the years

`Primevue` was added as dependency because we use their datatable, but for consistency with the design of the rest of the app, we should avoid using primevue components for things like icons, datepickers, buttons, inputs, etc. to avoid clashing with the existing datalab design. (even though there are some places where inconsistent styling and use of primevue components has definitely already crept in!)

Overall, I would recommend trying to match the style of the EditPage by using bootstrap classes as much as possible. This generally means less custom CSS is needed, and buttons, selects, colors, etc. are consistent across the app.

jdbocarsly (commented):
Neat feature! I will leave some general design comments while Matt is away. When he's back, I'm sure he'll be able to advise on implementation details and how to best integrate this with the overall app.

I have left some comments specific to the Vue code and design below. Additionally, I think the `SampleStagesTimeline.vue` file is a bit long for a single component. It would benefit from being broken up into a few modular components and/or js files that make it easier to read and maintain. A lot of the length is also from the CSS, which I believe could be reduced by using standard bootstrap classes where possible (see comment below)

PR #1829, opened 2026-06-11
Fix bug where only admins could be set as managers

### LINKED ISSUE #1816: Admin panel: Only admins can be selected as managers
Previously, any user could be selected in the manager Multiselect for each user. Now, it seems that only admins can be selected. Previously assigned managers are still working (I believe), but new ones can't be assigned. 

For context, we use this a lot for lab members who are supervising interns, etc.

jdbocarsly:
At some point, the admin panel started only letting admins be set as user managers. The issue seems to be some code that checks if the user is an admin OR already a manager before letting them be set as a manager. This PR removes that, and also fixed a spacing issue in the vSelect.

closes #1816


ml-evs (approved):
LGTM!

PR #1830, opened 2026-06-12
remove 'only' from SampleTable e2e tests, ensuring all tests run

jdbocarsly:
The SampleTablePage tests had one of the tests marked with cypress's `.only`, so only that test ran. 

This PR turns on all the tests, causing the test to fail due to numerous parts of the test that were quite out of date. Minimal fixes are made.


ml-evs (approved):
Thanks for spotting and fixing this!

PR #1831, opened 2026-06-12
Styling for readonly pages

### LINKED ISSUE #1826: UX with read-only permissions is poor
Currently, when one is given a read-only permission to a sample (e.g. via group directly or via collection), the UX is quite poor as the following widgets are editable:

- `creators` / `shared with groups` fields
- main `description` field
- `description` field on any existing block

Of course, when one tries to save the changes, this fails. But it might be better if the widgets were not editable in the first place.

(Note that trying to add a new block with read-only permissions fails immediately.)

jdbocarsly:
This in-progress PR creates consistent readonly styling across item views. 

Todo: 

- [x] Give each ItemInformation component an `isEditable` prop
- [x] Convert simple inputs to StyledInputs that can be set to `readonly` if the item is not editable
- [x] Give chemforminputs a readonly mode
- [x] give SubstanceInformation a readonly mode
- [x] give tiptap a readonly mode
- [ ] give the toggleable fields (collections, creators etc.) readonly modes
- [x] give the synthesis/construction information components readonly modes
- [ ] Figure out how to make blocks readonly neatly
- [ ] Use user permissions to decide whether to render the editable or readonly page for each time

Closes #1826

### DIFF
--- a/webapp/src/components/CellInformation.vue
+++ b/webapp/src/components/CellInformation.vue
@@ -84,7 +84,7 @@
           </div>
           <div class="form-group col-lg-4 col-md-4 pr-3">
             <label for="cell-chemform">Active formula</label>
-            <ChemFormulaInput id="cell-chemform" v-model="ChemForm" />
+            <ChemFormulaInput id="cell-chemform" v-model="ChemForm" :readonly="!isEditable" />
           </div>
           <div class="form-group col-lg-3 col-md-4">
             <label for="cell-characteristic-molar-mass">Molar mass</label>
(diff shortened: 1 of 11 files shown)


davidwaroquiers:
Hi @jdbocarsly 


Thanks for this PR, it would definitely be much better in terms of UI as currently it "feels" like you can edit your read only views (at least part of it). Is there anything blocking at this stage for this ? I've asked @DianaAliabieva to review this to have some feedback when she has time (she's busy on a few other things right now but she will probably have finished by end of this week).


Best,

DianaAliabieva:
Hello @jdbocarsly ! I have sent you a message in slack to clarify some things! Have a nice day!

PR #1820, opened 2026-06-07
Update to Pydantic v2

ml-evs:
Supersedes #1415... after many attempts, we are finally there. This PR updates pydantic to v2, starting from cherry-picking various commits from #1415, then adding fixes as I went. 

I still want to tidy up the commit narrative a little, but perhaps @be-smith you could review this heuristically (i.e., install the app, try break it) as likely there are some extra snagging points. e.g., not sure if we still need to apply [28d35ea](https://github.com/datalab-org/datalab/pull/1415/commits/28d35eac3147925493f7b4d66c9a8bfe5e5d7f58).

Main awkward changes that remain after doing `uvx bump-pydantic`:

- reworking all model validators to new format
- `x: int | None` used to imply a default value of `None` in pydantic 1, but no more...
- pydantic 2 is stricter around model loading and circular imports
- aliases work differently, and we were being quite haphazard in our application of them

Likely we'll want some follow up PRs that refine some of this, but I'm happy that I didn't have to update the tests very much at all beyond the simple `.dict() -> .model_dump_json()` and related changes. We could even ignore these for now by ignoring the warning and using the backward-compat version, which might make this PR clearer.

Copilot's full description:

# Generated summary

This pull request modernizes the codebase to fully support Pydantic v2 and its new companion library `pydantic-settings`. It updates dependencies, refactors model usage to leverage new Pydantic v2 APIs, and adapts schema and config handling accordingly. Additionally, it improves plugin compatibility by gracefully handling deprecation warnings from legacy plugins.

**Pydantic v2 Migration and Model Refactoring:**

* Updated dependencies to require `pydantic >=2.11,<3.0` and added `pydantic-settings` and `python-dotenv` for configuration management. Removed the block on `pydantic` v2 in Dependabot config. (`pydatalab/pyproject.toml`, `.github/dependabot.yml`, `.pre-commit-config.yaml`) [[1]](diffhunk://#diff-61345f47f81a84864b64b56b273c3c38951eec67b57ff56ee7c3634efd8b23dfL24-R26) [[2]](diffhunk://#diff-dd4fbda47e51f1e35defb9275a9cd9c212ecde0b870cba89ddaaae65c5f3cd28L28-L29) [[3]](diffhunk://#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9L70-R70)
[...]

PR #1832, opened 2026-06-12
Fix status dropdown broken by Pydantic v2 schema migration

jbouquiaux:
Pydantic v2's model_json_schema() uses "$defs" instead of "definitions" (Pydantic v1). Update the four item-information components to use the correct key so possibleItemStatuses is no longer undefined.


ml-evs:
merged into #1820

Issue #48, opened 2021-10-10
Move blocks to their own collection instead of being embedded documents within items


jdbocarsly:
For more flexibility with the data model, we should move blocks into their own collection instead of having them live as embedded documents in the `blocks_obj` field of an item. 

Need to decide:
- The ids for the blocks. Currently, they are randomly generated HTML-friendly strings, but we could choose to use the mongo ObjectIDs 
- Blocks could potentially be associated with more than one sample (or other item). We should probably change the `item_id` field to something like `item_ids` and then also store the block id in each item it is associated for so that there is two-way referencing of the items and blocks. 

  - If we do this, perhaps a block should still have (the option of) having a "primary" `item_id`

davidwaroquiers:
Bring this issue up as we might be interested in going forward with it.

@jdbocarsly what is/are the potential use case(s) that you see for allowing blocks to be associated with more than one sample ?

One thing we'd be interested to have (somewhat related to the moving of the blocks to their own collection): ability to share a block with a given group or user (without sharing e.g. the details of the item it is associated with except maybe its name and/or item_id).

jdbocarsly:
Hi @davidwaroquiers — from what I recall, I was imagining experiments that involved using 2 (or more) different samples being measured at the same time. Thinking back, this hasn't actually come up a lot in the 5 (!) years since this issue was written. But, I could envision some experiments when it would be useful to have the same block show up on two different Sample pages. 

In general, we have wanted to move blocks to their own MongoDB Collection to make various operations cleaner. For example, it is currently awkward to write a query for just a block (or several blocks), since they live as a nested object inside their Item documents. In some cases, the current approach can make the Item Documents too large given Mongo's max doc size of 16 MiB. 

Making blocks live in their own collection should make it more general to share or reference individual blocks too, as you say!

davidwaroquiers:
Hi @jdbocarsly 


Thanks for your message. We'd be happy to try to go forward with this. What would you say ? Should we consider the use case that a given block can belong to more than one item (Sample, Cell, ...) ? Or should we keep (maybe for now ?) the same behavior that basically, one block is tied to one and only one item, even though they are in different collections. I don't think it would be much more difficult to allow a given block to be associated to two (or more) samples.


There is obviously a backward incompatible change in terms of database structure of course and a migration as mentioned in #1184 is most likely essential here (even without allowing "many samples to one block").


[...]

PR #1833, opened 2026-06-12
Tools for backward incompatible DB upgrades

gpetretto:
As mentioned in #48 we would like to introduce a backward incompatible change to split the blocks data from the items. We are thus proposing a general procedure for handling backward incompatible upgrades to the DB (see  #1184).

The whole implementation is heavily based on the one we have developed for [jobflow-remote](https://github.com/Matgenix/jobflow-remote/blob/develop/src/jobflow_remote/jobs/upgrade.py) and should be used only when truly needed, as we recongnize the pain and the potential issues associated with DB migrations.

## Overview

The main change is the addition of the `upgrade` module that contains a `DatabaseUpgrader`. A new migration requires writing a function that performs the required updates on the DB and decorating it with`@DatabaseUpgrader.register_upgrade("X.Y.Z")`, where `"X.Y.Z"` is the new version that will be released.

In order to determine the datalab version associated with the DB a new collection (`database_metadata`) is created, where a document stores the current version number of datalab. 

When the code is updated to a newer version, the upgrade procedure checks the datalab version stored in the DB and compares it with the version of the code being executed, applying sequentially all the intermediate upgrades to the DB, and finally updating the document in `database_metadata` with the current datalab version. This allows to build incremental upgrades between different versions. 

This upgrade can be executed with an invoke task: `migration.upgrade`.

## Initialization

A downside of the current proposal is that storing the datalab version in the DB requires an initialization. This is done through the invoke task `admin.initialise-schema-version` that needs to be executed during the initial deployment. We have considered doing this automatically in the `create_app`, but the main issue is to tell apart a DB associated to an already existing deplyoment from a pristine one, since in both cases the `database_metadata` will be empty. In principle one could imagine defining a procedure to guess the status of the DB (e.g., no items present or all the collections should be empty), but it would be easy to imagine cases where these checks might lead to the wrong conclusion. For example, if checking that all the collections are empty, this may fail to recognize a new deployment if in the future someone adds some other metadata document or if a user is added to the DB before starting the server. So, we ended up proposing this manual initialization procedure. It would be easy to switch to an automated one if preferred or if a reliable criteria can be determined.

## Example
 
As an example of how to create a new upgrade, here is a draft to modify `relationships` to be based on `refcode` instead of `item_id` (see #1184). I don't think this covers properly all the cases and I did not attempt to make a complete one here, as it is beyond the scope of this PR. Just an idea of how this would work. Some more example can be found in jobflow remote, whose implementation is very similar.

<details>

<summary>relationships upgrade example</summary>

```python
@DatabaseUpgrader.register_upgrade("0.8.0")
def pin_item_id_references_to_refcodes(
    db: pymongo.database.Database,
    session: ClientSession | None = None,
    dry_run: bool = False,
) -> list[UpgradeAction]:
    # Build a lookup from item_id to refcode
    item_id_to_refcode = {
        doc["item_id"]: doc["refcode"]
        for doc in db.items.find(
            {"item_id": {"$ne": None}, "refcode": {"$ne": None}},
            projection={"item_id": 1, "refcode": 1},
            session=session,
        )
    }

    pinned = 0
    deprecated = 0

    for item in db.items.find({}, projection={"_id": 1, "relationships": 1}, session=session):
        relationships = item.get("relationships") or []

        changed = False
        for rel in relationships:
            # Skip relationships that already use refcode or don't use an item_id.
            if rel.get("refcode") or not rel.get("item_id"):
                continue

            refcode = item_id_to_refcode.get(rel["item_id"])
            if refcode is not None:
                rel["refcode"] = refcode
                pinned += 1
            else:
                # Deprecate when no item matches
                rel["deprecated"] = True
                deprecated += 1
            changed = True

        if changed and not dry_run:
            db.items.update_one(
                {"_id": item["_id"]},
                {"$set": {"relationships": relationships}},
                session=session,
            )

    return [
        UpgradeAction(
            description=f"Pin {pinned} item_id reference(s) to their refcode",
            collection="items",
            action_type="update",
            details={"pinned": pinned},
        ),
        UpgradeAction(
            description=f"Deprecate {deprecated} unresolvable item_id reference(s)",
            collection="items",
            action_type="update",
            details={"deprecated": deprecated},
        ),
    ]
```

</details>

## TODO

Document the upgrade and initialize procedure if approved.

Let us know what you think or if you need more details.

### DIFF
--- a/pydatalab/src/pydatalab/main.py
+++ b/pydatalab/src/pydatalab/main.py
@@ -23,6 +23,23 @@ from pydatalab.utils import BSONProvider
 COMPRESS = Compress()
 
 
+def _check_database_version() -> None:
+    """Run the non-fatal schema-version check at startup.
+
+    The server always starts; depending on the recorded version it may stamp the
+    current version (when provably safe) or log guidance to run
+    ``invoke admin.initialise-schema-version`` / ``invoke migration.upgrade``. See
+    :mod:`pydatalab.upgrade`.
+    """
+    try:
+        from pydatalab.mongo import get_database
+        from pydatalab.upgrade import check_database_version_on_startup
+
+        check_database_version_on_startup(get_database())
+    except Exception as exc:
+        LOGGER.warning("Could not check the database schema version: %s", exc)
+
+
 def create_app(
     config_override: dict[str, Any] | None = None, env_file: pathlib.Path | None = None
 ) -> Flask:
@@ -107,6 +124,7 @@ def create_app(
 
     pydatalab.mongo.create_default_indices()
     pydatalab.mongo.run_startup_migrations()
+    _check_database_version()
 
     if CONFIG.FILE_DIRECTORY is not None:
         pathlib.Path(CONFIG.FILE_DIRECTORY).mkdir(parents=False, exist_ok=True)
--- a/pydatalab/src/pydatalab/upgrade.py
+++ b/pydatalab/src/pydatalab/upgrade.py
@@ -1,0 +1,552 @@
+"""Administrator-driven database upgrades for backward-incompatible changes.
+
+This module provides the machinery for applying *backward-incompatible* changes
+to the database when a datalab deployment is upgraded to a new version. The
+intended workflow is to stop the server, run ``invoke migration.upgrade`` on a
+host with access to the database, and then start the new version.
+
+To register an upgrade, decorate a function with
+:meth:`DatabaseUpgrader.register_upgrade`. The function receives the database
+handle, an optional session (for transactions), and a ``dry_run`` flag, and must
+return the list of :class:`UpgradeAction` it performed (or would have performed,
+when ``dry_run`` is set).
+"""
+
+from __future__ import annotations
+
+import contextlib
+import enum
+import functools
+import time
+from collections.abc import Callable
+from dataclasses import dataclass
+from datetime import datetime, timezone
+from typing import TYPE_CHECKING, ClassVar
+
+from packaging.version import InvalidVersion, Version
+from packaging.version import parse as parse_version
+
+from pydatalab import __version__
+from pydatalab.logger import LOGGER
+from pydatalab.mongo import METADATA_COLLECTION
+
+if TYPE_CHECKING:
+    import pymongo.database
+    from pymongo.client_session import ClientSession
+
+# ``_id`` of the document (in :data:`~pydatalab.mongo.METADATA_COLLECTION`) that
+# holds the database schema version.
+SCHEMA_VERSION_ID = "datalab_schema_version"
+# Sentinel "start of history" used as the lower bound when collecting upgrades
+# for a database whose starting version is not otherwise known.
+BASELINE_VERSION = Version("0.0.0")
+
+
+class UpgradeConditionsFailedError(Exception):
+    """Raised when an upgrade is blocked because its preconditions are not satisfied."""
+
+
+class UninitialisedDatabaseError(RuntimeError):
+    """Raised when an operation needs a recorded schema version but none exists.
+
+    An unstamped database is ambiguous: it could be a pristine new installation
+    (which must be stamped at the current version with no upgrades applied) or an
+    existing database from before schema versioning (which may need upgrades).
+    The system never guesses between these; the operator must disambiguate with
+    ``admin.initialise-schema-version`` (new install) or
+    ``migration.upgrade --from-version`` (existing database).
+    """
+
+
+class DatabaseVersionStatus(enum.Enum):
+    """Outcome of comparing the recorded schema version with the code version."""
+
+    #: Development install (code version unknown); no check performed.
+    DEV = "dev"
+    #: Recorded version matches the code version.
+    CURRENT = "current"
+    #: Version was missing or behind but nothing needed applying, so it was
+    #: stamped to the current version automatically.
+    AUTO_STAMPED = "auto_stamped"
+    #: Recorded version is behind and registered upgrades need to be applied.
+    UPGRADE_PENDING = "upgrade_pending"
+    #: No version recorded and upgrades exist: the operator must initialise or
+    #: upgrade explicitly (see :class:`UninitialisedDatabaseError`).
+    UNINITIALISED = "uninitialised"
+    #: Recorded version is newer than the code (e.g. an attempted downgrade).
+    AHEAD = "ahead"
+
+
+def code_version() -> Version | None:
+    """The version of the installed ``datalab-server`` package.
+
+    Returns ``None`` for development builds. A build is considered "development"
+    when the version cannot be parsed or is_devrelease.
+    """
+    try:
+        version = parse_version(__version__)
+    except InvalidVersion:
+        return None
+    if version.local is not None or version.is_devrelease:
+        return None
+    return version
+
+
+@dataclass
+class UpgradeAction:
+    """Details of a single upgrade action to be performed."""
+
+    description: str
+    collection: str
+    action_type: str
+    details: dict
+    required: bool = False
+
+
+# Declared ``kw_only`` so that subclasses can introduce required fields.
+@dataclass(kw_only=True)
+class UpgradeCondition:
+    """A generic precondition that must hold before an upgrade can be applied."""
+
+    description: str
+    check_func: Callable[[DatabaseUpgrader, UpgradeCondition], dict | None] | None = None
+    skippable: bool = False
+
+    def check(self, upgrader: DatabaseUpgrader) -> dict | None:
+        if self.check_func is None:
+            raise NotImplementedError("check_func must be defined")
+        result = self.check_func(upgrader, self)
+        if result and self.skippable:
+            result["message"] += (
+                " This condition can be bypassed by running the upgrade with the `--force` option."
+            )
+        return result
+
+
+@dataclass(kw_only=True)
+class NoDocumentsIn(UpgradeCondition):
+    """Condition: a collection must contain no document matching a query."""
+
+    collection: str
+    query: dict | None = None
+    description: str = None  # type: ignore[assignment]
+
+    def __post_init__(self):
+        if not self.description:
+            q_str = f" matching {self.query}" if self.query else ""
+            self.description = (
+                f"There should be no document in the '{self.collection}' collection{q_str}"
+            )
+
+        def _check(upgrader: DatabaseUpgrader, _condition: UpgradeCondition) -> dict | None:
+            count = upgrader.db[self.collection].count_documents(self.query or {})
+            if count == 0:
+                return None
+            return {
+                "condition": self,
+                "message": f"Found {count} document(s)",
+                "count": count,
+            }
+
+        self.check_func = _check
+
+
+class DatabaseUpgrader:
+    """Handles upgrading the datalab database between versions.
+
+    Upgrade functions are registered against the version that introduces the
+    backward-incompatible change via :meth:`register_upgrade`. When run, the
+    upgrader collects every registered upgrade between the database's current
+    schema version and the target version and applies them in order, recording
+    the new version in the database after each step.
+    """
+
+    _upgrade_registry: ClassVar[dict[Version, Callable]] = {}
+    _upgrade_conditions_registry: ClassVar[dict[Version, list[UpgradeCondition]]] = {}
+
+    def __init__(self, db: pymongo.database.Database):
+        self.db = db
+        self.current_version = code_version()
+        self._supports_transactions: bool | None = None
+
+    @classmethod
+    def register_upgrade(
+        cls, version: str, upgrade_conditions: list[UpgradeCondition] | None = None
+    ):
+        """Decorator registering an upgrade function for the given version.
+
+        Parameters:
+            version: The version that introduces the change handled by ``func``.
+            upgrade_conditions: Preconditions that must hold before the upgrade
+                can be applied.
+        """
+
+        def decorator(func: Callable):
+            @functools.wraps(func)
+            def wrapper(*args, **kwargs):
+                LOGGER.info("Executing upgrade to version %s", version)
+                start_time = time.perf_counter()
+                result = func(*args, **kwargs)
+                duration = time.perf_counter() - start_time
+                LOGGER.info("Completed upgrade to %s in %.2fs", version, duration)
+                return result
+
+            vv = parse_version(version)
+            cls._upgrade_registry[vv] = wrapper
+            if upgrade_conditions:
+                cls._upgrade_conditions_registry[vv] = upgrade_conditions
+            return wrapper
+
+        return decorator
+
+    @property
+    def registered_upgrades(self) -> list[Version]:
+        return sorted(self._upgrade_registry.keys())
+
+    def get_db_version(self) -> Version | None:
+        """Return the schema version recorded in the database, or ``None``.
+
+        This is a pure read: it never infers a version from database contents and
+        never writes. ``None`` means the database has no recorded schema version
+        (see :class:`UninitialisedDatabaseError`).
+        """
+        doc = self.db[METADATA_COLLECTION].find_one({"_id": SCHEMA_VERSION_ID})
+        if doc and doc.get("version"):
+            return parse_version(doc["version"])
+        return None
+
+    def is_initialised(self) -> bool:
+        """Whether the database has a recorded schema version."""
+        return self.get_db_version() is not None
+
+    def update_db_version(
+        self,
+        version: Version,
+        session: ClientSession | None = None,
+        action: str = "upgrade",
+    ) -> None:
+        """Record ``version`` as the database schema version.
+
+        Every write also appends an entry to the ``history`` array of the metadata
+        document (version, action, timestamp and the running datalab version), as
+        an audit trail of how the recorded version evolved.
+
+        Parameters:
+            version: The version to record.
+            session: Optional session when running inside a transaction.
+            action: What caused the write: ``"upgrade"``, ``"initialise"`` or
+                ``"auto_stamp"``.
+        """
+        history_entry = {
+            "version": str(version),
+            "action": action,
+            "timestamp": datetime.now(timezone.utc),
+            "datalab_version": __version__,
+        }
+        self.db[METADATA_COLLECTION].update_one(
+            {"_id": SCHEMA_VERSION_ID},
+            {
+                "$set": {"version": str(version)},
+                "$push": {"history": history_entry},
+            },
+            upsert=True,
+            session=session,
+        )
+
+    def supports_transactions(self) -> bool:
+        """Whether the connected MongoDB supports multi-document transactions.
+
+        Transactions require a replica set (or sharded cluster); datalab can also
+        run against a standalone ``mongod``, where they are unavailable.
+        """
+        if self._supports_transactions is None:
+            try:
+                hello = self.db.client.admin.command("hello")
+                self._supports_transactions = bool(
+                    hello.get("setName") or hello.get("msg") == "isdbgrid"
+                )
+            except Exception:
+                self._supports_transactions = False
+        return self._supports_transactions
+
+    @contextlib.contextmanager
+    def open_transaction(self):
+        """Yield a transaction session if supported, otherwise yield ``None``."""
+        if self.supports_transactions():
+            with (
+                self.db.client.start_session() as session,
+                session.start_transaction(),
+            ):
+                yield session
+        else:
+            yield None
+
+    def collect_upgrades(self, from_version: Version, target_version: Version) -> list[Version]:
+        """Versions with a registered upgrade in ``(from_version, target_version]``."""
+        return [v for v in self.registered_upgrades if from_version < v <= target_version]
+
+    def check_upgrade_conditions(
+        self, versions: list[Version], force: bool = False
+    ) -> list[tuple[Version, dict]]:
+        """Return any failed preconditions for the given ``versions``."""
+        failed_conditions = []
+        for version in versions:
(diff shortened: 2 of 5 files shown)


pydatalab/src/pydatalab/upgrade.py
```diff
@@ -0,0 +1,552 @@
...
+from pydatalab.logger import LOGGER
+from pydatalab.mongo import METADATA_COLLECTION
+
+if TYPE_CHECKING:
+    import pymongo.database
+    from pymongo.client_session import ClientSession
+
+# ``_id`` of the document (in :data:`~pydatalab.mongo.METADATA_COLLECTION`) that
+# holds the database schema version.
+SCHEMA_VERSION_ID = "datalab_schema_version"
```
davidwaroquiers, line 39: Why is it named schema here ? It's just the database version no ? In more details, it is the version of the datalab-server that was installed before migrating. But I find it a bit confusing to talk about datalab schema version but maybe I'm missing something or it's standard name for this ? My feeling is that schema version is not always corresponding to package version (even though for datalab core I guess it should), also considering plugins for example, we may have times where we update a plugin in which the schema of the data is changed in the plugin but the (I don't know what we'd have to do in that case btw, but let's put that for another time)

davidwaroquiers: If this comment holds, I would rename all to "datalab version" or similar, e.g. below "admin.initialise-datalab-version" instead of "admin.initialise-schema-version".

pydatalab/src/pydatalab/upgrade.py
```diff
@@ -0,0 +1,552 @@
...
+                self._upgrade_registry[version](self.db, session=session)
+                self.update_db_version(version, session=session)
+
+        # Record the final target version (covers the case where the target is
+        # ahead of the last registered upgrade); skip it when the last applied
+        # upgrade already recorded exactly the target version.
+        if not versions_needing_upgrade or versions_needing_upgrade[-1] != target:
+            self.update_db_version(target)
+        LOGGER.info("Database upgrade completed successfully (now at %s).", target)
+        return True
```
davidwaroquiers, line 497: It's the same in jobflow-remote but I would maybe slightly change this and have the common parts together in another method so that if we change something there it is applied to both the dry-run and the upgrade. Something like below (codex generated - it wanted to make a new BuildPlan dataclass, I refused :p)


```suggestion
    def _build_upgrade_plan(
        self,
        from_version: str | None = None,
        target_version: str | None = None,
        force: bool = False,
        include_actions: bool = False,
    ) -> tuple[
        Version,
        Version,
        list[Version],
        list[tuple[Version, dict]],
        list[UpgradeAction],
    ]:
        """Resolve the common upgrade plan used by dry-runs and real upgrades."""
        db_version, target = self._resolve_versions(from_version, target_version)
        if db_version >= target:
            return db_version, target, [], [], []

        versions_needing_upgrade = self.collect_upgrades(db_version, target)
        failed_conditions = self.check_upgrade_conditions(versions_needing_upgrade, force=force)

        actions: list[UpgradeAction] = []
        if include_actions:
            for version in versions_needing_upgrade:
                actions.extend(self._upgrade_registry[version](self.db, dry_run=True))

            actions.append(
                UpgradeAction(
                    description=f"Update database schema version to {target}",
                    collection=METADATA_COLLECTION,
                    action_type="update",
                    details={
                        "filter": {"_id": SCHEMA_VERSION_ID},
                        "update": {"$set": {"version": str(target)}},
                        "upsert": True,
                    },
                )
            )

        return db_version, target, versions_needing_upgrade, failed_conditions, actions

    def dry_run(
        self,
        from_version: str | None = None,
        target_version: str | None = None,
        force: bool = False,
    ) -> tuple[list[UpgradeAction], list[tuple[Version, dict]]]:
        """Simulate the upgrade, returning the actions and any failed conditions.

        Makes no changes to the database.
        """
        _, _, _, failed_conditions, actions = self._build_upgrade_plan(
            from_version=from_version,
            target_version=target_version,
            force=force,
            include_actions=True,
        )
        return actions, failed_conditions

    def upgrade(
        self,
        from_version: str | None = None,
        target_version: str | None = None,
        force: bool = False,
    ) -> bool:
        """Apply the database upgrades from the current to the target version.

        Returns ``True`` if any upgrade was applied, ``False`` if the database
        was already at the target version. Raises
        :class:`UpgradeConditionsFailedError` if any precondition is not satisfied.
        """
        db_version, target, versions_needing_upgrade, failed_conditions, _ = (
            self._build_upgrade_plan(
                from_version=from_version,
                target_version=target_version,
                force=force,
            )
        )
        if db_version >= target:
            LOGGER.info("Database is already at version %s; nothing to upgrade.", db_version)
            return False

        if failed_conditions:
            err = ["Some upgrade conditions were not satisfied:"]
            for vv, failed in failed_conditions:
                err.append(
                    f" - {failed['condition'].description} (for version {vv}): {failed['message']}"
                )
            message = "\n".join(err)
            LOGGER.error(message)
            raise UpgradeConditionsFailedError(message)

        LOGGER.info("Starting upgrade from version %s to %s", db_version, target)
        for version in versions_needing_upgrade:
            with self.open_transaction() as session:
                LOGGER.info("Applying upgrade to version %s", version)
                self._upgrade_registry[version](self.db, session=session)
                self.update_db_version(version, session=session)

        # Record the final target version (covers the case where the target is
        # ahead of the last registered upgrade); skip it when the last applied
        # upgrade already recorded exactly the target version.
        if not versions_needing_upgrade or versions_needing_upgrade[-1] != target:
            self.update_db_version(target)
        LOGGER.info("Database upgrade completed successfully (now at %s).", target)
        return True
```

ml-evs (commented):
Thanks for this @gpetretto! Some initial thoughts. On the surface, this feels like taking on a lot of extra complexity that I need to really digest before deploying. I would really prefer if we can break this down a bit and build up from some simple cases first rather than a single commit/PR going to the final answer. 

Initially, I wasn't considering a system that would allow arbitrary migrations to be specified as database entries and applied in turn, but instead just a robust way of calling a series of migrations in turn based on database version numbers, i.e., developer driven migrations. I can see how this will be useful once custom items are in, but I really think we need to start simple.

Could you take a look at the existing single "migration" we have in mongo.py and build from there? I think the current approach would clash with the simple `run_startup_migrations` approach I added for that. We can definitely start using the database version stamp properly, but I think for the blocks collection migration (#48), I would also want to use this approach for it.

pydatalab/src/pydatalab/upgrade.py
```diff
@@ -0,0 +1,552 @@
...
+                start_time = time.perf_counter()
+                result = func(*args, **kwargs)
+                duration = time.perf_counter() - start_time
+                LOGGER.info("Completed upgrade to %s in %.2fs", version, duration)
+                return result
+
+            vv = parse_version(version)
+            cls._upgrade_registry[vv] = wrapper
+            if upgrade_conditions:
+                cls._upgrade_conditions_registry[vv] = upgrade_conditions
```
davidwaroquiers, line 197: It is the same as for jobflow remote but if we decorate two methods for the same version, then only the last one is registered (of course). I guess the developer should know or it should be clear, but maybe there could be a check here ?

pydatalab/src/pydatalab/upgrade.py
```diff
@@ -0,0 +1,552 @@
...
+    ) -> list[tuple[Version, dict]]:
+        """Return any failed preconditions for the given ``versions``."""
+        failed_conditions = []
+        for version in versions:
+            for condition in self._upgrade_conditions_registry.get(version, []):
+                if force and condition.skippable:
+                    continue
+                if (check := condition.check(self)) is not None:
+                    failed_conditions.append((version, check))
+        return failed_conditions
```
davidwaroquiers, line 299: I am wondering if we may have a flaw here. What if we do an upgrade from 0.1 to 0.3 (through 0.2) and that 0.3 requires e.g. that a given collection is empty, but the migration action from 0.1 to 0.2 solves that (it moves or cleans up that collection). We have no way of doing it right ? Or at least we would need to upgrade "manually" from 0.1 to 0.2, then from 0.2 to 0.3. What do you think ?

pydatalab/src/pydatalab/upgrade.py
```diff
@@ -0,0 +1,552 @@
...
+
+        Raises:
+            RuntimeError: if the database is already initialised and ``force`` is
+                not set, or if the version cannot be determined.
+        """
+        target = parse_version(version) if version else self.current_version
+        if target is None:
+            raise RuntimeError(
+                "Cannot determine the version to initialise to on a development "
+                "install; pass an explicit version."
```
davidwaroquiers, line 372: Here there are two cases no ? One is the explicitly passed version is "wrong" (version is passed as an argument), wrong being (version cannot be determined), or the current_version is None (i.e. it is a development install). I would maybe try to clarify the two cases and explicitly mention when it is a development install (case where version is None in the argument and current_version returns None => explicitly say something like "Cannot determine the version to initialise to on a development install; pass an explicit version.\nDevelop version : {__version__}") and when it is an explicitly passed value (case where version is not None => explicitly say something like "Cannot determine the version to initialise to from the version passed as an argument.\nVersion passed as an argument: {version}") What do you think ?

pydatalab/src/pydatalab/upgrade.py
davidwaroquiers: If we indeed have this problem, that means that it is also difficult to do a full dry-run when it is a "multiple version" upgrade

pydatalab/src/pydatalab/upgrade.py
```diff
@@ -0,0 +1,552 @@
...
+    def get_db_version(self) -> Version | None:
+        """Return the schema version recorded in the database, or ``None``.
+
+        This is a pure read: it never infers a version from database contents and
+        never writes. ``None`` means the database has no recorded schema version
+        (see :class:`UninitialisedDatabaseError`).
+        """
+        doc = self.db[METADATA_COLLECTION].find_one({"_id": SCHEMA_VERSION_ID})
+        if doc and doc.get("version"):
+            return parse_version(doc["version"])
```
davidwaroquiers, line 215: Related to my comment on datalab_schema_version, maybe we can rename the field "datalab_version" (in jobflow_remote it is named "jobflow_remote_version". I think it is more clear like that. What do you think ?

pydatalab/src/pydatalab/upgrade.py
```diff
@@ -0,0 +1,552 @@
...
+    CURRENT = "current"
+    #: Version was missing or behind but nothing needed applying, so it was
+    #: stamped to the current version automatically.
+    AUTO_STAMPED = "auto_stamped"
+    #: Recorded version is behind and registered upgrades need to be applied.
+    UPGRADE_PENDING = "upgrade_pending"
+    #: No version recorded and upgrades exist: the operator must initialise or
+    #: upgrade explicitly (see :class:`UninitialisedDatabaseError`).
+    UNINITIALISED = "uninitialised"
+    #: Recorded version is newer than the code (e.g. an attempted downgrade).
```
davidwaroquiers, line 76: Correct me if I'm wrong, but we are not trying to have this as a possibility right ? (I think it makes things much more complicated and probably impossible in some cases without loss of data). I would then explicitly mention this somewhere (maybe here, something like "Downgrades are and will never be covered by this tool"). The procedure is that the user should (I would say must) backup the database before trying to upgrade and if it fails he recovers from the backed up database. (somehow the "Simple operator flow" diagram that I shared shows this and it is explained in a few places in the code I think, it will (should) be explained in the docs when this is approved.

pydatalab/src/pydatalab/upgrade.py
```diff
@@ -0,0 +1,552 @@
...
+        """Apply the database upgrades from the current to the target version.
+
+        Returns ``True`` if any upgrade was applied, ``False`` if the database
+        was already at the target version. Raises
+        :class:`UpgradeConditionsFailedError` if any precondition is not satisfied.
+        """
+        db_version, target = self._resolve_versions(from_version, target_version)
+        if db_version >= target:
+            LOGGER.info("Database is already at version %s; nothing to upgrade.", db_version)
+            return False
```
davidwaroquiers, line 468: This is True for db_version == target, maybe distinguish when db_version > target, that then it is an attempted downgrade and that this is not possible.

pydatalab/src/pydatalab/upgrade.py
```diff
@@ -0,0 +1,552 @@
...
+        return target
+
+    def check_and_stamp_version(self) -> DatabaseVersionStatus:
+        """Compare the recorded version with the code version (startup check).
+
+        Updates the version in the DB to the current version only when it is safe,
+        i.e. when the database is unstamped or behind but no registered upgrade
+        falls in the range, so nothing could be skipped.
+        Never applies upgrades and never guesses an unstamped database is current
+        when upgrades exist.
```
davidwaroquiers, line 391: I am wondering if it is always fine to auto stamp (i.e., is it always "provably safe"). Should it be that the upgrade procedure always need to be performed "explicitly/manually" by the admin and if there is nothing to do, then it will just update the datalab_version. What do you think ?

pydatalab/src/pydatalab/main.py
```diff
@@ -23,6 +23,23 @@
 COMPRESS = Compress()
 
 
+def _check_database_version() -> None:
+    """Run the non-fatal schema-version check at startup.
```
davidwaroquiers, line 27: Naming (schema-version) => maybe better datalab-version (see also other comments related to this)

pydatalab/tasks.py
```diff
@@ -415,6 +415,123 @@ def add_missing_refcodes(_):
...
+        "force": "Overwrite an already-recorded schema version.",
+    }
+)
+def initialise_schema_version(_, version=None, force=False):
+    """Record the database schema version for a NEW installation.
+
+    Use this once on a fresh deployment. It stamps the current version without
+    applying any upgrades (there is no pre-existing data to migrate). For an
+    existing database that predates schema versioning, use `migration.upgrade`
+    with `--from-version` instead.
```
davidwaroquiers, line 430: Also see other comments on "naming"

pydatalab/tasks.py
```diff
@@ -415,6 +415,123 @@ def add_missing_refcodes(_):
...
+        print(f"  - [{action.collection}] {action.description}")
+
+    if failed_conditions:
+        print("\nThe following preconditions are NOT satisfied:")
+        for vv, failed in failed_conditions:
+            print(f"  - {failed['condition'].description} (for version {vv}): {failed['message']}")
+        if not force:
+            print(
+                "\nResolve the issues above, or re-run with `--force` to bypass skippable conditions."
+            )
```
davidwaroquiers, line 501: Here, it says to resolve the issues above if there are failed_conditions, but it does not stop. Is it normal ?

Also, (but I have another comment on this), I am wondering if there isn't a flaw in the logic of the conditions when the upgrade concerns multiple version upgrades (see the other comment)

davidwaroquiers (changes requested):
Good piece of work ! (we already had a nice base from jfr)
I have made quite some comments and I think we have a few things to discuss and possibly change/adapt. Maybe it would be good to discuss with @ml-evs about the logic before implementing any change from my comments so that we are aligned and that it covers both potential use cases such as the blocks (maybe "optional" migration of the blocks) and existing use cases that are already mentioned in e.g. #1184

davidwaroquiers:
> Thanks for this @gpetretto! Some initial thoughts. On the surface, this feels like taking on a lot of extra complexity that I need to really digest before deploying. I would really prefer if we can break this down a bit and build up from some simple cases first rather than a single commit/PR going to the final answer.
> 
> Initially, I wasn't considering a system that would allow arbitrary migrations to be specified as database entries and applied in turn, but instead just a robust way of calling a series of migrations in turn based on database version numbers, i.e., developer driven migrations. I can see how this will be useful once custom items are in, but I really think we need to start simple.
> 
> Could you take a look at the existing single "migration" we have in mongo.py and build from there? I think the current approach would clash with the simple `run_startup_migrations` approach I added for that. We can definitely start using the database version stamp properly, but I think for the blocks collection migration (#48), I would also want to use this approach for it.

Hi Matthew,


Thanks a lot for your initial thoughts and comments.


I had in the meantime done a thorough review of the PR and some my review comments are related to some of your comments.


I'm first attaching a rough schematics of the logic behind that piece of work. It might show that the complexity is not that high in the end but as a tool that is adapted but in the end very similar to what we have done in jobflow remote, it has been tested in production for quite some time now. Note that the current schematics is also still with a "naming" that is under discussion (see some explanations after the diagram).

image

On the left side is the "operator flow", basically what happens when you deploy. On the right side is the implementation logic of the upgrade tool.


Regarding the scope of this migration tool, the goal (initially at least) is not to cover schema versions as the naming could suggest (which, as you mention, would be very useful for custom item plugins) but only for core datalab migrations. So it is essentially something similar to your startup migrations. I have two main comments in my review that are related to this:
- one regarding the naming in the code (it mentions "schema version", but I'd rather make this very clear that it is the datalab version instead).
- one asking if we could extend this tool to deal with possible different schema versions (for the custom item types), I am btw indicating in this comment that usually (probably always ?) the two coincide for the datalab-core (i.e. if a data schema for a core item type is changed, the datalab version is changed too). This is not within the scope of this first PR


As such, the current PR for this tool is exactly what you were imagining: "a robust way of calling a series of migrations in turn based on database version numbers, i.e., developer driven migrations".


The @DatabaseUpgrader.register_upgrade decorator does not register an upgrade as a database entry or anything, it is just a helper to define the upgrade procedure for a given version. Now it does not deal with migrations that are or should be *always* applied at deployment time (which is what the startup migrations do in mongo.py, correct ?). We can probably merge the two approaches somehow so that there is one single tool for migrations.


I hope this clarifies some of your initial comments. I have to also mention that I have the feeling that it is almost completely orthogonal to the entire datalab codebase except for startup checks that only gives a warning about the "status" of the database wrt to the current code, suggesting what to do, e.g. (excerpt from the code):

```
"============================================================\n"
            "A DATABASE UPGRADE IS PENDING for this datalab version.\n"
            "The server will keep running, but some data may fail to load\n"
            "until you stop the server and run:\n"
            "    invoke migration.upgrade\n"
            "Take a backup first with `invoke admin.create-backup`.\n"
            "============================================================"
```

Regarding having the blocks moving to their own mongodb collection as a migration using this tool or as a startup migration that is always applied, I'm open to discussion. It of course depends whether the blocks moving is a backward incompatible change or if we try to keep the two coexisting (up to you, see discussion in the blocks issue).


Don't hesitate to reach out or reply here if you have any questions or further comments.

pydatalab/src/pydatalab/upgrade.py
davidwaroquiers: Ok following our discussion, we raise if it's already registered so there is only one upgrade method for one release. We don't want to have multiple migration/upgrade methods tied to the same release otherwise we may have problems of which one has to be executed first, difficulties to make the unit/integration tests etc ... At least with one single method, everything is gathered and the developer knows that he has to "deal" with possible other migrations.

pydatalab/src/pydatalab/upgrade.py
davidwaroquiers: Ok agreed, guido will revise

pydatalab/src/pydatalab/upgrade.py
davidwaroquiers: Ok agreed, guido will revise

pydatalab/src/pydatalab/upgrade.py
davidwaroquiers: Ok so we will remove this and make the check per version in the loop within the upgrade method as we have no way to know beforehand (as a dry-run) that the conditions needed for a given release migration will not change the conditions of the next release migration in a multi-release migration.

pydatalab/src/pydatalab/upgrade.py
davidwaroquiers: We don't need to know the Develop version in the error message but for the explicitly passed version, we do:

when it is a development install (case where version is None in the argument and current_version returns None => explicitly say something like "Cannot determine the version to initialise to on a development install; pass an explicit version.")

when it is an explicitly passed value (case where version is not None => explicitly say something like "Cannot determine the version to initialise to from the version passed as an argument.\nVersion passed as an argument: {version}")

pydatalab/src/pydatalab/upgrade.py
davidwaroquiers: Ok

pydatalab/src/pydatalab/upgrade.py
davidwaroquiers: I let you decide, verify and clarify

pydatalab/src/pydatalab/upgrade.py
davidwaroquiers: Split the two.

pydatalab/src/pydatalab/upgrade.py
davidwaroquiers: Maybe ask Matthew here about the safest "provably safe" new deployment and discuss with him. => to be discussed

pydatalab/src/pydatalab/main.py
davidwaroquiers: up to you for the name of the function, chnage in the comment anyway

pydatalab/tasks.py
davidwaroquiers: ok

pydatalab/tasks.py
davidwaroquiers: Ok, to be checked also with the change of the logic from all conditions checked at once to checking conditions per version.

PR #1836, opened 2026-06-12
Improvements to substance info display (rounding, toggle and styling)

jdbocarsly:
A few small tweaks to the very nice `SubstanceInfo` component:

1. Currently, it can display a large number of decimal places, especially due to floating point rounding errors:

Screenshot 2026-06-12 at 4 17 03 PM

This is changed so that the displayed value is rounded to 2 decimals (the actual value can be set to any desired precision; it just won't be displayed)

2. The toggle behavior is tweaked, allowing the user to close the edit window by clicking on the header for more consistency.

3. Switched over from PrimeVue inputs to native HTML inputs with bootstrap styling (`.form-control`), for consistency across the app.

### DIFF
--- a/webapp/src/components/SubstanceInfoSmall.vue
+++ b/webapp/src/components/SubstanceInfoSmall.vue
@@ -7,7 +7,12 @@
       </div>
       <div v-if="molarMass" class="substance-small-row">
         <span class="small-label">Mass</span>
-        <span class="small-value">{{ molarMass }} g/mol</span>
+        <span class="small-value"
+          >{{
+            Number.isFinite(Number(molarMass)) ? Number(molarMass).toFixed(2) : molarMass
+          }}
+          g/mol</span
+        >
       </div>
       <div v-if="ghsCodes && hasPictograms" class="substance-small-row">
         <span class="small-label">Hazards</span>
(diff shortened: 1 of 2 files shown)


jdbocarsly:
Also realized there is a "small" version of this component, so I formatted the molar mass on that as well. 

However, one thing I realize on the display of SubstanceInfoSmall on the data table is that it only pops up on hover, so CAS button will never be clickable (the display goes away if you move the mouse to go click on it). Not sure whether we want to:

-  leave that be
- change it so the CAS is a text display instead of a link
or
- change the behavior so that the substance info pops up when the user clicks the substance, instead of hovers. And goes away on clicking anywhere else.
Screenshot 2026-06-19 at 12 10 22 PM

ml-evs (approved):
Thanks for this @jdbocarsly!

> However, one thing I realize on the display of SubstanceInfoSmall on the data table is that it only pops up on hover, so CAS button will never be clickable (the display goes away if you move the mouse to go click on it). Not sure whether we want to:


Hmm, this is annoying (and the same for SMILES and other clickable components). Not sure how urgently we need to fix this, but we should open an issue. My preference would be that the hover still activates when the mouse is over the pop up.

PR #1838, opened 2026-06-13
add LocationInput component for hierarchical location editing

### LINKED ISSUE #1711: Add a specialised component for location field
I'm imagining some kind of accordion that shows each segment of a location string separately and enables you to autocomplete each segment when adding something. This is related to @be-smith's PR #1695 which added autocomplete and filtering (though I must admit I'm not sure how this location list is propagated atm).

Also related to #1447, as this field is one that should be tracked in a special way in the item versions component, e.g., if location changes between two versions, it should be explicitly listed in the version history.

Another one for @BenjaminCharmes to cogitate over!

BenjaminCharmes:
Closes #1711

- Adds `LocationInput.vue`, a dedicated component that splits the `" > "`-delimited location string into individually editable segments, each with context-aware autocomplete

- Uses `LocationInput` in read-only mode in `VersionHistoryModal.vue` so location values render consistently in version history

### DIFF
--- a/webapp/cypress/e2e/equipment.cy.js
+++ b/webapp/cypress/e2e/equipment.cy.js
@@ -141,7 +141,8 @@ describe("Equipment edit page", () => {
     cy.findByText("test_e3").click();
     cy.findByLabelText("Description").type("this is a description of testB.");
     cy.findByLabelText("Date").type("2000-01-01T00:00");
-    cy.findByLabelText("Location").type("room 101");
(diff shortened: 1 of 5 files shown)


ml-evs (approved):

Just need to think a bit about:

- a nicer style to show the collapsible/expandable location in view mode
- how to extend this approach to samples (I'm going to add the field now) and be able to pull all known locations more easily (will likely need a separate API)

Issue #1411, opened 2025-11-10
Make a separate repository of versioned database snapshots


ml-evs:
Useful for test data, unexpected migrations etc.

PR #1845, opened 2026-06-18
Add Docker-based development environment

jbouquiaux:
Hello,

We have recently added a "dev" profile in the docker-compose of Matgenix fork so that we start our developments from consistent setups. Since I think this is quite general and could potentially benefits everyone, I'm opening this PR to suggest this new "dev" profile. There is also a new dev.seed invoke command that allows to quickly populate the DB with items containing blocks created from the example files.

### Summary  

Add Docker-based development environment

Sets up a hot-reloading, fully containerised dev stack so contributors only need Docker to get started.

- `docker-compose.yml`: adds a dev profile (database-dev, api-dev, app-dev) with source bind-mounts, hot reload (Flask --reload + Vue HMR), and an isolated MongoDB on port 27018. DB name is configurable via DATALAB_DB_NAME.
- `.docker/app/Dockerfile`: adds a dev stage that stops before the production webpack bundle, so app-dev builds faster and runs vue-cli-service serve instead.
- `pydatalab/.env.example`,  `webapp/.env.example`: annotated config templates to copy for local dev.
- `DEVELOPMENT.md`: quick-start guide covering first-run, login (GitHub OAuth), handy commands, and a dev vs prod profile comparison table.
- `pydatalab/tasks.py`: dev.seed to cover every block type that has example data (XRD, Raman, echem, NMR, FTIR, MS, UV-vis, tabular, media, comment) plus a starting material, one item per file, idempotent re-runs.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,0 +1,71 @@
+# Development
+
+Hot-reloading datalab stack (API + web app + MongoDB) in Docker. Only Docker is required.
+
+## Quick start
+
+```bash
+cp pydatalab/.env.example pydatalab/.env     # backend config (edit it, see below)
+cp webapp/.env.example webapp/.env           # web app config
+
+docker compose --profile dev up --build      # first run builds; later just `up`
+docker compose exec api-dev /opt/.venv/bin/invoke dev.seed   # optional: spawn example data in the DB
+```
+
+- Web app: http://localhost:8081
+- API: http://localhost:5001
+- Database: `mongodb://localhost:27018`
+
+Edits to `pydatalab/` and `webapp/` hot-reload. Editing `.env` does not — restart the
+service: `docker compose --profile dev restart api-dev` (or `app-dev`).
+
+## Logging in
+
+The web app needs a logged-in session. Set up GitHub OAuth once:
+
+1. Create an OAuth App at https://github.com/settings/developers
+   - Homepage URL: `http://localhost:8081`
+   - Authorization callback URL: `http://localhost:5001/login/github/authorized`
+2. Put the credentials in `pydatalab/.env` (see the commented GitHub block in
+   `.env.example`), set `PYDATALAB_TESTING=false`, restart `api-dev`.
+3. Set `PYDATALAB_SECRET_KEY` to a real value (generate one with
+   `python3 -c 'import secrets; print(secrets.token_hex(64))'`).
+4. **Login via GitHub** in the web app.
+
+
+## Handy commands
+
+```bash
+# Make yourself admin (display name = the name shown in the web app)
+docker compose exec api-dev /opt/.venv/bin/invoke admin.change-user-role \
+  --display-name "Your Name" --role admin
+
+docker compose logs -f api-dev          # follow API logs
+docker compose --profile dev down       # stop (add -v to wipe the DB + files)
+
+# Use a custom database name (default: datalabvue). Set in your shell before
+# running `docker compose up` — it is not read from pydatalab/.env.
+export DATALAB_DB_NAME=my_project
+docker compose --profile dev up
+```
+
+Backend tests run natively (the Docker dev image ships without test deps); from
+`pydatalab/` use `uv sync --all-extras --dev` once, then `uv run pytest`.
+
+
+## `dev` vs `prod` profiles
+
+Both profiles are in `docker-compose.yml` and share the same Dockerfiles; the `dev`
+services target earlier build stages and override the run command.
+
+| | `prod` (`app`, `api`, `database`) | `dev` (`app-dev`, `api-dev`, `database-dev`) |
+|---|---|---|
+| Code | baked into the image at build time | bind-mounted from your checkout |
+| Reload | none (rebuild to change code) | hot reload (Flask `--reload`, Vue HMR) |
+| Server | gunicorn / static build | dev servers (`dev.serve`, `vue-cli-service serve`) |
+| Database | host path `/data/db`, port not exposed | Docker volume, port `27018`, isolated per dev |
+| Files/backups | mounted host paths under `/data` | a throwaway Docker volume |
+| Restart policy | `unless-stopped` | none |
+
+Run with `docker compose --profile prod up` or `--profile dev up`. Don't run both at
+once — they share ports 5001/8081.
--- a/pydatalab/.env.example
+++ b/pydatalab/.env.example
@@ -1,0 +1,32 @@
+# Example environment for the pydatalab API server during development.
+#
+# Copy this to `pydatalab/.env` and edit. It is loaded by `invoke dev.serve`
+# natively, and inside the Docker dev container (the ./pydatalab:/app bind-mount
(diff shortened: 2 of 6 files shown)


ml-evs (commented):
Hi @jbouquiaux, just building on my email response, could you take a look at the dev environment we used to have for this? The only differences I can see is that it also pulled all the dependencies needed to run the cypress tests, which might be useful. It was removed in 2f17743252e8513baf546754eba341872bda8dd1 

Regarding the seed database tasks, this looks super useful (see #1411) but I think I'd rather create snapshot dumps that are created from the UI to make sure they are "real". These can also be versioned so that it interacts well with any future migrations and can be used as test data for those (#1833).

Oh and the dev builds should likely be tested in the CI if we merge this, as removed in https://github.com/datalab-org/datalab/commit/721c322204289806c8a3b516204cc67c54ddbaac Perhaps it could be done in a lighter weight way (like your compose-only dev version of the API container) but we probably need to know when they break

OMWalmsley:
The database seed functionality appears to work well on my machine, alongside the new dev environment.

The seed data functionality looks useful. But  could be further refined with some of the following:
- For unique ids: by removing the while loop and adding random Id suffixes, generated with a predefined random function with known seed. This would allow the ids to be repeatable (and deterministic) but without the need for this `while True:` statement.
  - Even without this change the `_make_item_id` function could be simplified.
- Potentially directories could be listed instead of files in `_SEED_BLOCK_GROUPS` and then files could be extracted using `glob`. This would allow for more flexibility in case we decide to introduce more example files or indeed decide to go down a route of snapshot dumps.
- `_SEED_ITEMS` is used once after it is declared (though maybe this is intended for future use). It would be currently possible to move this inline.
- `def seed(_)` could be split into multiple functions. Especially within the `for spec in _SEED_ITEMS`. This is potentially a style difference. Although it would cut down on the use of continue statements and improve the readability of the code. Potentially one function for:
  -  posting the new sample
  - uploading the example files.
  - creating the data block.
  
In terms of the environment variables (`.env.example`), my preference would probably be to move them completely into the documentation (`DEVELOPMENT.md`). The user would then have to copy and paste these into their `.env` file, but would also reduce redundancy in the code base and be more inline with how the current set up seems to work for datalab.
  
I am happy to discuss my comments, some are more stylistic preferences rather than any concrete improvements.

jbouquiaux:
Hi @ml-evs, @OMWalmsley , thanks for your feedback.

@ml-evs, since the dev environment was previously removed, I want to make sure reintroducing it is still welcome? As this indeed add some maintenance burden. Let us know!

Assuming we do keep it, all the comments make sense to me. One clarification on the seed data approach before going further:

* Should I replace dev.seed with a snapshot-based approach, or keep it (with @OMWalmsley's suggested improvements) alongside a new dev.load-snapshot task?
*  Which instance would the snapshot be extracted from? The public demo, a dedicated seed instance, or a manually populated local dev?
* Should the dump live in the main repo (e.g., tests/fixtures/) or in the dedicated snapshot repo from [#1411](https://github.com/datalab-org/datalab/issues/1411) ?

In the meantime, I'll have a look at the other items (Cypress deps, CI).

pydatalab/.env.example
```diff
@@ -0,0 +1,32 @@
+# Example environment for the pydatalab API server during development.
+#
+# Copy this to `pydatalab/.env` and edit. It is loaded by `invoke dev.serve`
+# natively, and inside the Docker dev container (the ./pydatalab:/app bind-mount
+# maps it to /app/.env). `.env` is gitignored; this template is not.
+# Full reference: https://docs.datalab-org.io/en/latest/config/
+#
```
ml-evs, line 7: If this is to become the canonical source for these env vars, can we find a way to directly inline this file in the documentation at build time (to avoid having multiple places to update env var names if we change them)?

I think a lot of the "documentation" style comments here should also be removed in favour of reading the actual docs, but the template is useful

pydatalab/tasks.py
```diff
@@ -165,6 +165,337 @@ def serve(_, host: str = "127.0.0.1", port: int = 5001, reload: bool = True, tes
 dev.add_task(serve)
 
 
+# Example data for `dev.seed`, grouped by the block that renders it. Each group
```
ml-evs, line 168: Likely this script could be added next to a pre-made snapshot of the database in the separate repo discussed above, so let's not just delete it

webapp/.env.example
```diff
@@ -0,0 +1,22 @@
+# Example environment for the Vue web app during development.
```
ml-evs, line 1: Same comment as above for .env files, should be possible via pymdownx.snippets which we already use in the docs

DEVELOPMENT.md
```diff
@@ -0,0 +1,71 @@
+# Development
+
+Hot-reloading datalab stack (API + web app + MongoDB) in Docker. Only Docker is required.
+
+## Quick start
```
ml-evs, line 5: This should be put in the appropriate place in the installation docs, not in a separate file, and I think much of this content can be removed

ml-evs (commented):
> @ml-evs, since the dev environment was previously removed, I want to make sure reintroducing it is still welcome? 

If its useful to you guys, then we should definitely add it back in. Could you just make sure the dev build also get triggered in the CI?
 
> Assuming we do keep it, all the comments make sense to me. One clarification on the seed data approach before going further:
> 
>     * Should I replace dev.seed with a snapshot-based approach, or keep it (with @OMWalmsley's suggested improvements) alongside a new dev.load-snapshot task?
>     * Which instance would the snapshot be extracted from? The public demo, a dedicated seed instance, or a manually populated local dev?

I would prefer snapshots, though the first snapshot could be very similar to the one you have created using example data. We do already have `restore_snapshot` in `pydatalab.backups` that might be useful here.

>     * Should the dump live in the main repo (e.g., tests/fixtures/) or in the dedicated snapshot repo from [#1411](https://github.com/datalab-org/datalab/issues/1411) ?

I'd prefer a snapshot repo -- these things should be relatively small and change very rarely (so we can either use LFS or just not bother). I'm imagining we might want a series of snapshots that could be useful for different features, and the one we've discussed here can be a basic "example-data" snapshot. My only suggestion would be to make it "manually" in the datalab UI/API to ensure it truly follows what a real database will look like -- I'm happy to create this myself though if you can go through the machinery of loading it.

From your side, could you just set up the functionality to be able to restore a snapshot from a URL or a local file (tar archive or otherwise)?

I also added a couple of my own commenmts through the PR below.

jbouquiaux:
Hi @ml-evs, @OMWalmsley,

After discussion with @davidwaroquiers, we thought it would be best to better separate concerns since this PR actually covers quite different topics. So I'm going to split it into three smaller PRs (keeping in mind your above comments):
1) Docker development profile + CI test
2) Environment variable handling, with env.example as canonical source
3) Development seed snapshots. 

Apologies for the extra reworking here!

I'll address 1) first, as it is the most useful at the moment! I'll link the replacement PRs here as they are opened.

davidwaroquiers:
Superseded by #1928 for the docker dev profile.
Environment variable handling and development seed will follow in other PRs.

PR #1846, opened 2026-06-18
Add web UI support for custom/plugin item types

jbouquiaux:
Hi @ml-evs, thanks a lot the preliminary work to support custom item types. I have tested it locally with a few simples then more complex types, it seems to work well! 
On top of you branch, I tried to wire the front-end part with the help of Claude. I hope this overlap with what you had in mind for the UI rendering.

- If a custom item is "simple" enough, I thought that it would nice if we could avoid writing custom .vue file, and that the rendering would be infered from the pydantic model. This is implemented here by adding `json_schema_extra=...` specific keys annotations that are defined  in `CustomFieldsPanel.vue`.  For instance,  this could looks like (semantically weird that a protocol herits from Sample but this is for the sake of the example :D) 

<details>
<summary>Expand example</summary>

```python
from typing import Literal
from pydantic import ConfigDict, Field
from pydatalab.models.samples import Sample
from pydatalab.models.utils import EntryReference

class AnnealingProtocol(Sample):
    model_config = ConfigDict(
        title="Annealing Protocol",
        json_schema_extra={
            "datalab_ui_hidden_fields": ["synthesis_information", "substance_information"],
            "datalab_section_title": "Annealing Schedule",
            "datalab_ui_color": "#b5651d",
        },
    )
    type: Literal["annealing_protocols"] = "annealing_protocols"

    peak_temperature: float | None = Field(
        None, ge=0, title="Peak temperature",
        json_schema_extra={
            "units": ["°C", "K"], "default_unit": "°C",
            "datalab_unit_field": "peak_temperature_unit",
            "datalab_include_field_in_summary": True,
        },
    )
    peak_temperature_unit: Literal["°C", "K"] = Field(
        "°C", json_schema_extra={"datalab_hidden": True}
    )

    # A field grouped into its own "Equipment" card, linking to an `equipment` item:
    furnace: EntryReference | None = Field(
        None,
        json_schema_extra={"datalab_ref_types": ["equipment"], "datalab_section": "Equipment"},
    )
```
</details>

This renders like: [full model](https://github.com/Matgenix/datalab/blob/jbouquiaux/pr-custom-items/pydatalab/plugins/example_item_plugin/example_item_plugin/models.py#L60-L61)
Capture d'écran 2026-06-18 à 15 21 26

- If this annotation system is not enough to cover the use case, I still think it is usefull to have the possibility to have full control (values computed in the browser, pulling data from linked items, custom tables, ...). In this case, there is the option to provide a full .vue file that takes over the full rendering. [Example model](https://github.com/Matgenix/datalab/blob/jbouquiaux/pr-custom-items/pydatalab/plugins/example_item_plugin/example_item_plugin/models.py#L153) and [corresponding .vue file](https://github.com/Matgenix/datalab/blob/jbouquiaux/pr-custom-items/pydatalab/plugins/example_item_plugin/example_item_plugin/webapp/HeatTreatmentPanel.vue).
This looks like :
Capture d'écran 2026-06-18 à 15 23 13

## Claude generated summary : 

# Web UI support for custom / plugin item types

## Overview

This work wires the frontend end-to-end for custom item types registered by out-of-tree plugins — types that the backend already serves but that the app previously had no way to display or edit. It introduces two rendering paths (automatic from annotations, or via a custom Vue panel shipped by the plugin), a plugin panel discovery invoke task, and an in-repo example plugin that exercises both paths.

---

### `/info/types` enrichment — `routes/v0_1/info.py`

The `/info/types` and `/info/types/<type>` endpoints now return `title`, `base_type`, `hidden_fields`, and `ui_color` alongside the JSON schema. `base_type` is resolved by walking the MRO against the built-in models so it works regardless of the plugin's inheritance depth.

### `dev.collect-plugin-panels` task — `tasks.py`

New invoke task (`uv run invoke dev.collect-plugin-panels`) that scans installed `pydatalab.item_types` packages, copies any `webapp/*.vue` files into `webapp/src/plugins/<package>/`, and generates `webapp/src/plugins/panels.generated.js` — an auto-generated map of `type_name → () => import(...)` factories. Convention: `HeatTreatment` → `HeatTreatmentPanel.vue`. Run it after `dev.install` when installing or updating plugins (and before building the webapp). The generated file is gitignored; only the committed shim (`index.js`) is tracked.

---

### Dynamic type registry — `resources.js`

- `itemTypes` is now `reactive({...})` so Vue computeds re-evaluate when plugin types register after startup.
- `registerDynamicItemType(type, { title, base_type, hidden_fields, ui_color })` adds a custom type into the shared registry, deriving the base information component from `BASE_TYPE_COMPONENTS`, computing a pastel badge color with `lightTint()`, and storing the custom panel factory from `PLUGIN_PANELS`.
- `prettifyType()` and `lightTint()` added as shared helpers.

### Schema loading — `server_fetch_utils.js`

`loadItemSchemas()` (called once from `App.vue` on startup) now calls `registerDynamicItemType` for each type returned by `/info/types` before caching schemas. `createNewItem` prepends custom-type items to the samples list and returns `item_id`.

### `EditPage.vue` — `EditPage.vue`

- Falls back to the `itemType` string in the navbar if the registry entry isn't loaded yet.
- `informationComponent`: renders the built-in base component (e.g. `SampleInformation`) so standard fields (name, description, relationships…) always appear.
- `hasCustomPanel` / `resolvedCustomPanel`: if the type has a plugin-shipped panel factory, wraps it with `defineAsyncComponent`; otherwise falls back to `CustomFieldsPanel`.

### `CustomFieldsPanel.vue` — `CustomFieldsPanel.vue` — new

Annotation-driven renderer for custom type fields. Diffs the type schema against its base type schema and renders only the delta fields. Supports:

- Number + unit compound widget (`datalab_unit_field`)
- Inline item-reference selector / display (`datalab_ref_types`) via `ItemSelect` + `FormattedItemName`
- Enum dropdown, boolean checkbox, multiline textarea (`datalab_multiline`)
- Multi-card grouping by `datalab_section`; default card titled by `datalab_section_title`
- Fields marked `datalab_hidden` are excluded; non-scalar fields (arrays, objects) are left to custom panels.

### Plugin panel shim — `plugins/index.js` — new

Stable committed shim that loads `panels.generated.js` only when present, exporting `PLUGIN_PANELS`. The generated file itself stays gitignored.

### Existing components updated

`CellInformation`,
`SampleInformation`,
`CreateItemModal`,
`ItemSelect`,
`FormattedItemStatus`,
`CompactConstituentTable`,
`EquipmentInformation`,
`StartingMaterialInformation`

Updated to use `FormattedItemName` for all item-reference badges (type-colored, correct click/cmd-click behavior) and to surface `hiddenFields` from the registry.

---

## Example plugin (`pydatalab/plugins/example_item_plugin/`)

Three custom `Sample` subtypes registered via `pydatalab.item_types` entry points:

- **`ExampleSample`** — minimal; one number+unit field (`voltage`).
- **`AnnealingProtocol`** — annotation-only, no JavaScript. Two cards: an annealing schedule (peak temperature, ramp rate, dwell time, atmosphere) and an Equipment & Safety card (furnace reference, supervision flag, hazard notes). Fully rendered by `CustomFieldsPanel`.
- **`HeatTreatment`** — ships `HeatTreatmentPanel.vue`: references an `AnnealingProtocol` (a custom type), seeds its as-run schedule via "Populate from protocol" (unit normalization K→°C, h→min), draws a live SVG temperature profile, and shows a second characterization card with a mass-loss progress bar.

---> This might not go in the final version... just here for the example.
---

## Config / infra

- `.gitignore`: un-ignores `example_item_plugin` specifically; all other `pydatalab/plugins/` entries (deployment-specific plugins) and generated `webapp/src/plugins/` files remain local-only.
---> This one is rather specific to this PR... Note sure it will be like that in the final version. 

- `.eslintignore`: excludes `pydatalab/` so the pre-commit eslint hook doesn't try to lint plugin Vue files that have no eslint config.
- `pydatalab/docs/plugins.md`: full plugin-author guide covering both rendering paths, all `json_schema_extra` annotations (field-level and `model_config`-level), the collect command, and the custom panel conventions.
---> Surely too verbose at the moment.


ml-evs (commented):
Thanks @jbouquiaux. Some brief feedback so we can move ahead:

- Can you add a schema for the schema hints you've added?
- We looked at the protocol example and think maybe this is better suited as either a field on a custom item or a specific new type "Protocol" which I think we discussed previously. Could we make a different simpler example of e.g., just adding a few fields to the default sample? I don't think we'd want this example as it stands committed here.

jbouquiaux:
Hey @ml-evs, thanks for the feedback.

* I've added two pydantic models under `pydatalab/src/pydatalab/models/schema_hints.py` that defines the ``datalab_*`` ``json_schema_extra`` hints, `DatalabFieldExtra` and `DatalabModelExtra`. They are exported as json schemas via invoke dev.generate-schemas. 
I did also reworked a bit the vocabulary (for instance from units/default_unit to datalab_units/datalab_default_unit) 

* I agree that the protocol example is too far from a simple extension of the default sample! Especially if we standardize these "Protocols" later... I changed these two examples with two simpler ones. 1) A `Solution` that has a solute ref, a concentration with unit hint, and a solvent ref. Rendered without custom .vue. 2) A `MixedSolution` which is a list of `Solution`s + volumes, and that allows to showcase the use of a custom .vue that compute and display minimal informations client-side.

Solution
Capture d’écran 2026-06-30 à 15 13 49

MixedSolution
 
Capture d’écran 2026-06-30 à 15 04 00

ml-evs:
Same problem as #1936, these PRs keep getting autoclosed as the base branch is deleted. I'm trying to restore all branches and reopen the PRs but GitHu does not seem to enjoy it

PR #1853, opened 2026-06-22
Fixes events from bokeh route to fire

### LINKED ISSUE #1834: Events aren't firing on XRD block and various plugins
Various plugins (XPS, printability) and core blocks (XRD) events' aren't firing. E.g changing the wavelength in XRD doesn't persist and actually change the wavelength.

I believe the culprit is in  :
```

else:
        if event_data and not event_data.get("trigger_async", True):
```

In routes/v0_1/blocks.py line 450

Simple fix is to remove the async check:

```

else:
        if event_data:
```

I haven't dug too deep into the async stuff but this at least makes events fire again.

be-smith:
`block.process_events(event_data)` was gated on `not event_data.get("trigger_async", True)`. Because `trigger_async` defaults to `True` when the key is absent — and the client never explicitly sets it to false — this condition was always `False`, meaning `process_events` was never called in either the sync or async processing path.

The result was that `@event()`-decorated handlers on block classes (e.g XRD setting the wavelength) were silently skipped on every request.

The fix removes the condition in both places: `process_events` is now called whenever `event_data` is present.

What `trigger_async` is actually for

`trigger_async` exists to decide whether a block update should be processed asynchronously (dispatched to the task scheduler) or synchronously. It is checked correctly at line 402 for that routing decision and is not touched by this PR. The bug was using the same flag to also gate `process_events` inside each path, which was incorrect — once you are already in the sync or async handler, `process_events` should always run if there is event data.

It is currently used for the insitu-blocks, where some "events" were triggered, e.g `onFolderSelected()`, this was presumably seen as a heavy event and so should be done asychronously. Lighter calls like `updateBlock()` pass `event_data = null`, which also defaults to async via the `if event_data else True` branch. No caller currently passes `trigger_async: false`. The bug was reusing this flag to gate `process_events` inside each path — once you are already in the sync or async handler, `process_events` should always run if event data is present.

Closes #1834

### DIFF
--- a/pydatalab/src/pydatalab/routes/v0_1/blocks.py
+++ b/pydatalab/src/pydatalab/routes/v0_1/blocks.py
@@ -75,7 +75,7 @@ def _process_block_async(
 
             block = BLOCK_TYPES[block_type].from_web(block_data)
 
-            if event_data and not event_data.get("trigger_async", True):
+            if event_data:
                 add_stage("Processing block events")
                 try:
                     block.process_events(event_data)
--- a/pydatalab/src/pydatalab/routes/v0_1/blocks.py
+++ b/pydatalab/src/pydatalab/routes/v0_1/blocks.py
@@ -447,7 +447,7 @@ def update_block():
             202,
         )
     else:
-        if event_data and not event_data.get("trigger_async", True):
+        if event_data:
             try:
                 block.process_events(event_data)
             except NotImplementedError:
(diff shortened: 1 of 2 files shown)


ml-evs (approved):
Thanks @be-smith -- I added a little test case for the XRD end-to-end tests that mimics the way the UI uses this just so we can avoid breaking this in future.

PR #1859, opened 2026-06-25
Potential fix for pydantic version control bug

be-smith:
Ensures that the data key doesn't have None values removed from it by removing exclude_none=True everywhere and having custom logic for the data key. Alternative fix to #1863 

Solves #1861


ml-evs:
merged into #1820

PR #1860, opened 2026-06-25
save_item route doesn't correctly overwrite saved values with None if the user removes a value

be-smith:
Solves #1864 

Adds a test that fails before the fix


ml-evs:
merged into #1820

Issue #1861, opened 2026-06-25
Pydantic v2 PR doesn't succesfully restore None values when restoring to previous version


be-smith:
The Pydantic v2 PR #1820 has a bug where if you restore to a previous version with a field with value None the restored version doesn't actually overwrite the value from before.

exclude_none = true is the culprit

ml-evs:
Closed by #1820 now

PR #1863, opened 2026-06-25
Add to_mongo_doc method to ItemVersionModel to protect None values

be-smith:
Ensures that the data key doesn't have None values removed from it by adding a custom method. Alternative fix to #1859 

Solves #1861


ml-evs:
Believe we incorporated the alternative fix into #1820

Issue #1875, opened 2026-06-29
Add optional in-app notifications feature


davidwaroquiers:
We would like to add an optional in-app notification system for datalab. The goal is to allow messages to be created in the backend and shown to users in the UI, for example:

- messages from administrators
- notifications generated by datalab itself
- notifications of alarms (see #1874)
- potentially, this could be opened to users as well (maybe this could/should be an option) to e.g. notify another user that a sample is ready for ...

This would be an optional deployment feature, disabled unless explicitly enabled by configuration using e.g. VUE_APP_ENABLE_NOTIFICATIONS or similar.

Motivation:
datalab currently has no generic way to inform users about events that happen outside the immediate workflow, for example ingestion anomalies, task failures, admin messages, or future automated system events.

The goal is to add a small, optional in-app notification system that can be enabled per deployment, without changing the behavior of existing deployments where notifications are disabled.

Proposed scope:
A first version should support persistent notifications (one notification document per recipient user) stored in MongoDB, read/unread states, archive/unarchive, title/message, sender/date metadata, grouped notifications for multiple occurrences

Email notifications and notification configuration should be out of scope for the initial implementation.

Proposed implementation plan:

1. Schemas and API routes + feature flag
2. Notification creation by admins
3. Frontend UI to show notifications for a given user
4. Frontend UI for archiving and/or deleting notifications

Any comments/feedback welcome here @ml-evs @jdbocarsly others ?

ml-evs:
Sounds good, I've also thought about adding this in the past -- adding the skeleton UI and API routes for this should be straightforward, then just need to figure out what should be emitting notifications and to who! I can also think of e.g., "this datalab has updated to v0.7.2 -- see what's new" style notifs but wouldn't want it to be to overbearing.

davidwaroquiers:
Great, thanks for the feedback! I'll go forward with this then.

PR #1904, opened 2026-07-06
Notifications backend

davidwaroquiers:
# Backend notifications API

## What this PR does

This adds the backend pieces for in-app notifications (see https://github.com/datalab-org/datalab/issues/1875)

Notifications are stored per user/recipient: even when one API request targets multiple users, the backend creates one notification document per recipient. This is intentional, because it keeps read/archive/delete state user-specific and avoids shared notification state becoming awkward. We are also not foreseeing a lot of messages that are sent to a lot of users except for update/maintenance notifications.

The feature is behind `ENABLE_NOTIFICATIONS`, which is still disabled by default. When enabled, the backend exposes routes to create, list, read/unread, archive/unarchive, delete, count unread notifications, and mark everything as read.

The PR also adds grouped notifications, so repeated events from the same source can update one existing notification instead of creating a new notification every time.

## How to use it

Currently, only admins can create notifications with:

```
POST /notifications
{
  "recipient_ids": ["..."],
  "title": "Import finished",
  "summary": "2 rows need attention",
  "message": "Some imported rows could not be matched.",
  "level": "important"
}
```

To send to every active user:
```
{
  "send_all_users": true,
  "title": "Scheduled maintenance"
}
```

Users can then use:

- GET /notifications
- GET /notifications/unread-count
- PATCH /notifications/<id>
- POST /notifications/mark-all-read
- DELETE /notifications/<id>

Users can only access their own notifications.

## Grouping behavior

A notification can be grouped by passing a grouping object:

```
{
  "recipient_ids": ["..."],
  "title": "Equipment anomaly",
  "summary": "Drift detected",
  "message": "The latest ingestion run detected drift.",
  "level": "critical",
  "grouping": {
    "key": "ingestion:equipment:eq-1",
    "policy": "once",
    "max_occurrences": 100
  }
}
```

Grouping "identity" is based on:

- recipient
- title
- grouping key
- grouping policy
- window_seconds for window grouping

Archived notifications are not reused.
For grouped notifications:

- new occurrences are appended to occurrences
- summary and message are updated to the latest occurrence
- level keeps the highest priority level seen so far
- a read notification becomes unread again when a new occurrence is added
- each occurrence has is_new, meaning “new since this notification was last read.”
- max_occurrences is owned by the stored grouped notification. So if the first grouped notification is created with max_occurrences=2, later matching events with max_occurrences=100 still group only until the stored cap of 2 is reached. The incoming cap only matters when a new grouped notification document is created.

## Notification levels

Current notification levels are:
- low
- normal
- important
- urgent
- critical

They have explicit priorities in the model, so grouped notifications can keep the highest-priority level.

No frontend changes are included in this PR.

The helper API create_notification(...) remains available for direct backend callers.


pydatalab/src/pydatalab/notifications.py
```diff
@@ -0,0 +1,189 @@
...
+    Returns:
+        A tuple of ``(notification, created)`` when a notification is created or
+        grouped. ``created`` is ``True`` when a new notification document was
+        inserted and ``False`` when the notification was folded into an existing
+        grouped notification. Returns ``None`` when notifications are disabled.
+    """
+
+    if not CONFIG.ENABLE_NOTIFICATIONS:
+        LOGGER.debug("Notifications are disabled; not creating notification %r", title)
+        return None
```
gpetretto, line 95: Not sure if this is important for the backend, but does it make sense that the notification creation fails without any message if the functionality is disabled? From the usage of this function it also seems that this just goes through as a successful request. And the log message is only visible in debug. 
Probably nobody will use this just from the backend, but it may be better to fail if the feature is disabled?

I don't know if this will be a good general approach, but in the tags it is done in this way: https://github.com/Matgenix/datalab/blob/e325a01c5fef8ca1d38ed12e89d92faa1803d51f/pydatalab/src/pydatalab/routes/v0_1/tags.py#L24

pydatalab/src/pydatalab/routes/v0_1/notifications.py
```diff
@@ -0,0 +1,280 @@
...
+                recipient_id=recipient_id,
+                title=title,
+                summary=summary,
+                message=message,
+                level=request_json.get("level", "normal"),
+                created_by=current_user.person.immutable_id,
+                grouping=grouping,
+            )
+        except (ValidationError, ValueError) as exc:
+            raise BadRequest(str(exc)) from exc
```
gpetretto: If this fails it may be that some users still received the notification. It seems unlikely, since in the loop only the `recipient_id` changes but still mentioning it. Not sure if a transaction is worth here.

pydatalab/src/pydatalab/notifications.py
```diff
@@ -0,0 +1,189 @@
...
+            "$push": {"occurrences": occurrence.dict(exclude_none=True)},
+            "$unset": {"read_at": ""},
+        }
+        flask_mongo.db.notifications.update_one(
+            {"_id": grouped_notification["_id"]},
+            mongo_update,
+        )
+        updated_notification = flask_mongo.db.notifications.find_one(
+            {"_id": grouped_notification["_id"]}
+        )
```
gpetretto, line 148: Why not `find_one_and_update` with `ReturnDocument.AFTER`?

pydatalab/src/pydatalab/notifications.py
```diff
@@ -0,0 +1,189 @@
...
+        grouping=grouping,
+        occurrence_count=1,
+        last_occurred_at=now,
+        occurrences=[occurrence],
+    )
+
+    return _insert_notification(notification), True
+
+
+def create_notification(
```
gpetretto: Why two functions `create_notification_with_result` and `create_notification`? 
Also it seems that when called the bool generated by `create_notification_with_result` is never used.

pydatalab/src/pydatalab/routes/v0_1/notifications.py
```diff
@@ -0,0 +1,280 @@
...
+    title = str(request_json.get("title", "")).strip()
+    message = (
+        str(request_json["message"]).strip() if request_json.get("message") is not None else None
+    )
+    summary = (
+        str(request_json["summary"]).strip() if request_json.get("summary") is not None else None
+    )
+
+    if not title:
+        raise BadRequest("Notification title is required.")
```
gpetretto: Is this the right approach? It does not seem that mandatory fields are enforced in other routes at creation. Isn't it verified by the pydantic validation?

pydatalab/src/pydatalab/routes/v0_1/notifications.py
```diff
@@ -0,0 +1,280 @@
...
+                raise BadRequest("recipient_ids must be a list.")
+            for recipient_id in raw_recipient_ids:
+                try:
+                    recipient_object_id = ObjectId(recipient_id)
+                except Exception as exc:
+                    raise BadRequest(f"Invalid recipient_id {recipient_id!r}.") from exc
+
+                if flask_mongo.db.users.find_one(
+                    {"_id": recipient_object_id},
+                    {"_id": 1},
```
gpetretto, line 82: For the option to send to all recipients the selected users are active. Should it be used also here in the search? Or is it fine even if it sent to inactive users?

pydatalab/src/pydatalab/routes/v0_1/notifications.py
```diff
@@ -0,0 +1,280 @@
...
+
+    # deduplicate recipient ids. uses a dict here to preserve order.
+    recipient_ids = list(dict.fromkeys(recipient_ids))
+    if not recipient_ids:
+        raise BadRequest("At least one recipient or all-users option must be provided.")
+
+    grouping = request_json.get("grouping")
+    if grouping is not None and grouping != {}:
+        if not isinstance(grouping, dict):
+            raise BadRequest("grouping must be an object.")
```
gpetretto: why an "object"? Should it say "dictionary"?

pydatalab/src/pydatalab/routes/v0_1/notifications.py
```diff
@@ -0,0 +1,280 @@
...
+    ), status_code
+
+
+@NOTIFICATIONS.route("/notifications", methods=["GET"])
+@with_notification_permissions
+def list_notifications(notification_permissions: dict):
+    include_archived = request.args.get("include_archived") == "1"
+    unread_only = request.args.get("unread_only") == "1"
+    limit = request.args.get("limit", default=50, type=int)
+    limit = max(1, min(limit, 100))
```
gpetretto: Why is the limit capped at 100?

pydatalab/src/pydatalab/routes/v0_1/notifications.py
```diff
@@ -0,0 +1,280 @@
...
+
+
+@NOTIFICATIONS.route("/notifications/mark-all-read", methods=["POST"])
+@with_notification_permissions
+def mark_all_notifications_read(notification_permissions: dict):
+    now = datetime.now(tz=timezone.utc)
+    unread_query = {**notification_permissions, "archived_at": None, "read_at": None}
+    grouped_result = flask_mongo.db.notifications.update_many(
+        {**unread_query, "occurrences.0": {"$exists": True}},
+        {"$set": {"read_at": now, "occurrences.$[].is_new": False}},
```
gpetretto, line 275: why here `read_at` can only be `now` and not passed?
Is the other one used in specific cases? Otherwise, why allowing it there and not here?

gpetretto (commented):
Overall it looks good to me. I left some comments.

pydatalab/src/pydatalab/notifications.py
davidwaroquiers: Good point. I'll add a before_request in the notifications route.


I'd still keep this part here that would be used in case a notification is created by datalab itself (not through a backend route). In that case there would indeed only be a log message in debug but I'm not sure what else could be done in that case. For the API for sure it's better with the added before_request. And it might actually be used from the backend API, e.g. from instruments.

pydatalab/src/pydatalab/notifications.py
davidwaroquiers: The bool is used to count how many notifications have been created vs how many have been grouped within a grouped notification here: https://github.com/datalab-org/datalab/pull/1904/changes/BASE..6b781f04ca30ee93ec73717b05b1ee860cd89635#diff-bfd9c0c43621332e4859cb9958d4f952c05260e58a5537d10ae5115c06a5187bR114

So create_notification is for possible future internal usage. We could drop it and add it later if/when needed. What do you think ?

pydatalab/src/pydatalab/notifications.py
davidwaroquiers: done

pydatalab/src/pydatalab/routes/v0_1/notifications.py
davidwaroquiers: I guess here I would allow even deactivated users as it is an explicit send. In principle deactivated users are allowed to authenticate but not allowed to edit. I'd maybe refer to @ml-evs about what he thinks about it when we undraft the PR

pydatalab/src/pydatalab/routes/v0_1/notifications.py
davidwaroquiers: done

pydatalab/src/pydatalab/routes/v0_1/notifications.py
davidwaroquiers: Added a transaction and a fallback in case transactions are not allowed

pydatalab/src/pydatalab/routes/v0_1/notifications.py
davidwaroquiers: Thought it would be already quite a lot. I'll remove and switch to a different approach in a later PR if that becomes a problem.

pydatalab/src/pydatalab/routes/v0_1/notifications.py
davidwaroquiers: It cannot be set in the other one too. It takes the read_at from the notification if it's there or it takes now. I could make a conditional to add to the update dict. What do you think ?

pydatalab/src/pydatalab/routes/v0_1/notifications.py
davidwaroquiers: True, I removed.

davidwaroquiers:
Thanks for the review @gpetretto I think I addressed the comments or asked some question. Feel free to resolve or propose anything you deem right.

pydatalab/src/pydatalab/routes/v0_1/notifications.py
davidwaroquiers: Note here that in the current mongodb container setup, transactions are not supported so it will fall back.

pydatalab/src/pydatalab/notifications.py
gpetretto: if there is not a clear use case for `create_notification` I think it can be removed. If you already know it will be used for something for the frontend part it can be left.

pydatalab/src/pydatalab/routes/v0_1/notifications.py
gpetretto: ok. In practice I don't think it makes really a big difference. The only difference would be for the sender of the notification that does not get a warning that the user is deactivated.

pydatalab/src/pydatalab/routes/v0_1/notifications.py
gpetretto: ok

pydatalab/src/pydatalab/routes/v0_1/notifications.py
gpetretto: I had misunderstood thinking that in `update_notification` it was a way to select a `read_at` from the input. So for me the problem was more the difference in behaviour. If there is no difference then I don't think that it would be really needed to have a way to set the `read_at` explicitly.

pydatalab/src/pydatalab/notifications.py
davidwaroquiers: ok, I removed

pydatalab/src/pydatalab/routes/v0_1/notifications.py
davidwaroquiers: Ok, I'll keep this conversation open for now

davidwaroquiers:
This is ready for review @ml-evs @be-smith @OMWalmsley

I addressed @gpetretto 's comments, with one being left up for discussion.

As mentioned in the main text of the PR, this is only the backend part for the notifications. I will make the frontend in another PR when this is approved and merged. Note that I have already made a working version of the frontend that uses all this backend functionality. I can show it in case this helps of course.

OMWalmsley (commented):
I think it generally looks good, just a few improvements that could be done in the notifications route.

pydatalab/src/pydatalab/routes/v0_1/notifications.py
```diff
@@ -0,0 +1,305 @@
...
+            raise BadRequest("grouping must be a dict.")
+        grouping = dict(grouping)
+    else:
+        grouping = None
+
+    level = request_json.get("level", "normal")
+    created_by = current_user.person.immutable_id
+
+    def create_for_recipients(session=None):
+        notification_results = []
```
OMWalmsley: `notification_results` shadows name from outer scope

pydatalab/src/pydatalab/routes/v0_1/notifications.py
```diff
@@ -0,0 +1,305 @@
...
+        for recipient_id in recipient_ids:
```
OMWalmsley: `recipient_id` shadows name from outer scope

pydatalab/src/pydatalab/routes/v0_1/notifications.py
```diff
@@ -0,0 +1,305 @@
...
+                    recipient_id=recipient_id,
+                    title=title,
+                    summary=summary,
+                    message=message,
+                    level=level,
+                    created_by=created_by,
+                    grouping=grouping,
+                    session=session,
+                )
+            except (ValidationError, ValueError) as exc:
```
OMWalmsley: exc also shadows name from outer scope

pydatalab/src/pydatalab/routes/v0_1/notifications.py
```diff
@@ -0,0 +1,305 @@
...
+
+            if notification_result is not None:
+                notification_results.append(notification_result)
+
+        return notification_results
+
+    try:
+        hello = flask_mongo.cx.admin.command("hello")
+        supports_transactions = hello.get("msg") == "isdbgrid" or "setName" in hello
+    except Exception:
```
OMWalmsley: Is it possible to narrow down what this Exception should be?
- (e.g. `pymongo.errors.PyMongoError`)

pydatalab/src/pydatalab/routes/v0_1/notifications.py
```diff
@@ -0,0 +1,305 @@
...
+        supports_transactions = False
+
+    if supports_transactions:
```
OMWalmsley: `supports_transactions` could be removed completely and the body of this `if` statement could be moved into `else` statement after `try...except`

pydatalab/src/pydatalab/routes/v0_1/notifications.py
```diff
@@ -0,0 +1,305 @@
...
+        with flask_mongo.cx.start_session() as session:
+            with session.start_transaction():
+                notification_results = create_for_recipients(session=session)
+    else:
+        notification_results = create_for_recipients()
```
OMWalmsley, line 131: Could move this into the `except` statement

pydatalab/src/pydatalab/routes/v0_1/notifications.py
```diff
@@ -0,0 +1,305 @@
...
+            "status": "success",
+            "unread_count": _count_unread_notifications(notification_permissions),
+        }
+    ), 200
+
+
+@NOTIFICATIONS.route("/notifications/<notification_id>", methods=["PATCH"])
+@notification_recipient_only
+def update_notification(
+    notification_id: str,
```
OMWalmsley: Similar to other comment (about `notification_id`).

pydatalab/src/pydatalab/routes/v0_1/notifications.py
```diff
@@ -0,0 +1,305 @@
...
+            "modified_count": grouped_result.modified_count + non_grouped_result.modified_count,
+            "unread_count": 0,
+        }
+    ), 200
+
+
+@NOTIFICATIONS.route("/notifications/<notification_id>", methods=["DELETE"])
+@notification_recipient_only
+def delete_notification(
+    notification_id: str,
```
OMWalmsley: `notification_id` is not used in this function, could possibly have decorator (`@notification_recipient_only`) remove this from `**kwargs`. Since the only two instances using it, do not make use of the `notification_id`.

pydatalab/src/pydatalab/routes/v0_1/notifications.py
davidwaroquiers: Thanks, done.

pydatalab/src/pydatalab/routes/v0_1/notifications.py
davidwaroquiers: Thanks, done.

pydatalab/src/pydatalab/routes/v0_1/notifications.py
davidwaroquiers: Thanks, done.

pydatalab/src/pydatalab/routes/v0_1/notifications.py
davidwaroquiers: Thanks, done.

pydatalab/src/pydatalab/routes/v0_1/notifications.py
davidwaroquiers: Can be indeed. I did it as it is probably cleaner. I would just wondering initially that we could actually make this test once at deployment time somehow and not having to test it everytime (and it would/could be read from current deployment config file or similar). Not sure if that's worth anyway so I propose we keep it as you propose for now.

pydatalab/src/pydatalab/routes/v0_1/notifications.py
davidwaroquiers: Refactored (see above comment)

pydatalab/src/pydatalab/routes/v0_1/notifications.py
davidwaroquiers: Thanks, done.

pydatalab/src/pydatalab/routes/v0_1/notifications.py
davidwaroquiers: Thanks, done.

davidwaroquiers (commented):
@OMWalmsley Thanks for the comments. I think I addressed all of them.

davidwaroquiers:
> I think it generally looks good, just a few improvements that could be done in the notifications route.

Thanks for your review. I believe I've addressed your comments. Feel free to check and resolve if it looks fine to you.

PR #1910, opened 2026-07-10
Add human-readable `last_modified` as a data table column

### LINKED ISSUE #969: Last saved state does not account for blocks & files changing
We need to update `last_modified` on an item when a file or block that is attached to it is added/updated/saved, otherwise the UI is misleading and seems to say that the block has not been saved.

Since the item data does not get redownloaded between the calls, we probably also need to manually set this in the UI at some point, though this could also give false security since the item may not have been saved successfully.

### LINKED ISSUE #1758: Last modified field should reflect changes in uploaded files
This would be helpful for workflows that incrementally pull and create dashboards from the datalab API.

### LINKED ISSUE #1782: Add last_modified as a sortable summary table column
We should have a "last edited x minutes ago" column on all tables.

ml-evs:
Closes #1782
Closes #1758 
Closes #969 

(the latter two by testing it properly).

This PR adds a column to the data tables for last_modified -- it renders it as a human-readable date using `date-fns`, e.g., "2 months ago".

It prevents the last_modified from updating if `save-item` endpoint doens't actually make any change when requested, controlled by whether a new version is minted.

It also saves which columns users could have selected in a table when they choose columns, and if a new column has been added, it shows them by default (rather than hiding it by default).

### DIFF
--- a/pydatalab/src/pydatalab/routes/v0_1/items.py
+++ b/pydatalab/src/pydatalab/routes/v0_1/items.py
@@ -1831,9 +1831,12 @@ def save_item():
             str(e),
         )
 
-    # Return the freshly-minted timestamp when content changed, otherwise fall back to the
-    # item's existing value (None for an item that has never been modified).
-    return jsonify(status="success", last_modified=new_last_modified or existing_last_modified), 200
+    # Report the freshly-minted timestamp when content changed. If no version was minted
+    # the item was unchanged, so flag it as a no-op and echo back the existing timestamp
+    if new_last_modified:
+        return jsonify(status="success", last_modified=new_last_modified), 200
+
+    return jsonify(status="success", unchanged=True, last_modified=existing_last_modified), 200
 
 
 @ITEMS.route("/items/<refcode>/access-token-info", methods=["GET"])
--- a/pydatalab/src/pydatalab/routes/v0_1/items.py
+++ b/pydatalab/src/pydatalab/routes/v0_1/items.py
@@ -1803,32 +1803,25 @@ def save_item():
     # (i.e., a snapshot was minted). If this fails, we log but don't fail the request
     # since the item was already saved.
     new_last_modified = None
-    try:
-        save_version_resp_dict, save_version_status = save_version_snapshot(
-            refcode,
-        )
-        if save_version_status != 200:
-            LOGGER.error(
-                "Failed to save version for item %s after successful update: %s",
-                item_id,
-                save_version_resp_dict,
-            )
-        elif "version" in save_version_resp_dict:
-            new_last_modified = datetime.datetime.now(tz=datetime.timezone.utc).isoformat()
-            flask_mongo.db.items.update_one(
-                {"item_id": item_id},
-                {
-                    "$set": {
-                        "version": save_version_resp_dict["version"],
-                        "last_modified": new_last_modified,
-                    }
-                },
-            )
-    except Exception as e:
+
+    save_version_resp_dict, save_version_status = save_version_snapshot(refcode)
+
+    if save_version_status != 200:
         LOGGER.error(
-            "Exception while saving version for item %s after successful update: %s",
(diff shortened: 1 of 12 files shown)


pydatalab/src/pydatalab/routes/v0_1/items.py
```diff
@@ -1772,6 +1774,12 @@ def save_item():
     item.pop("collections")
     item.pop("creators")
 
+    # `last_modified` is controlled by the versioning branch below: only bump it when a
+    # snapshot is actually saved, so re-submitting identical data leaves it untouched.
+    existing_last_modified = item.pop("last_modified", None)
+    if isinstance(existing_last_modified, datetime.datetime):
+        existing_last_modified = existing_last_modified.isoformat()
```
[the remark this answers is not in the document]
ml-evs: Will double check this, but I believe we fixed them separately in a previous PR

pydatalab/src/pydatalab/routes/v0_1/items.py
```diff
@@ -1804,18 +1814,30 @@ def save_item():
...
+                    }
+                },
             )
     except Exception as e:
         LOGGER.error(
             "Exception while saving version for item %s after successful update: %s",
             item_id,
             str(e),
         )
+        return jsonify(status="error", message="Failed to save version after item update."), 400
```
be-smith:
```suggestion
```

This return will falsely return a 400 that the save failed, whereas on line 1786-1789 has already successfully saved the item. I think this will falsely report to the user that the item wasn't saved even though it was.

webapp/src/components/SampleTable.vue
```diff
@@ -3,15 +3,7 @@
...
-    :global-filter-fields="[
-      'item_id',
-      'name',
-      'refcode',
-      'chemform',
-      'creatorsList',
-      'blocks',
-      'characteristic_chemical_formula',
-    ]"
+    :global-filter-fields="['item_id', 'name', 'refcode', 'chemform', 'creatorsList', 'blocks']"
```
be-smith:
```suggestion
    :global-filter-fields="['item_id', 'name', 'refcode', 'chemform', 'creatorsList', 'blocks', 'characteristic_chemical_formula']"
```

Removing this means you can't search for cells by chemical formula. E.g if you put the active material as NMC and search for NMC the cell won't appear. In fact I don't think cells can get chemform from HasSubstanceInfo which cell doesn't have. Potentially a deeper fix is to give cells a way to populate chemform, as it would also be useful if this NMC appears in the table view

webapp/src/components/CollectionInformation.vue
```diff
@@ -57,7 +57,7 @@
         'refcode',
         'blocks',
         'chemform',
-        'characteristic_chemical_formula',
+        'creatorsList',
```
be-smith, line 61:
```suggestion
        'creatorsList',
        'characteristic_chemical_formula',
```

be-smith (commented):
I think the first comment regarding the `return jsonify(status="error", message="Failed to save version after item update."), 400` is a real error. The others are more querying the design choice. Do we want to be able to search over the Active Cell Material in the sample table.

One thing Claude found is that we do our datetime inconsistently, where some have utc correction and some don't, so depending on the pathway the browser will sometimes convert to the native time or not. Currently making a PR adressing this

webapp/src/components/CollectionInformation.vue
be-smith: As above

pydatalab/src/pydatalab/routes/v0_1/items.py
ml-evs: Fixed this and removed the double error handling

webapp/src/components/SampleTable.vue
ml-evs: also fixed this

Issue #1907, opened 2026-07-09
Simplify login page to avoid exposing app structure


davidwaroquiers:
The current login page already reveals parts of datalab before authentication.

In most web applications, the login page is kept very minimal, typically only showing options such as "Log in with Microsoft", "Log in with Google", email OTP, etc.

Even if no actual data is exposed, showing parts of the application before login can still reveal useful information, such as the custom about page, enabled features (?), or potential future analysis plugins.

It would be preferable to keep the unauthenticated login page as simple and generic as possible.

davidwaroquiers:
Idea is to go from this:

<img width="1175" height="792" alt="Image" src="https://github.com/user-attachments/assets/e9ab07be-4ac3-4982-9000-997ca8339a61" />


To something like this:

<img width="1448" height="1086" alt="Image" src="https://github.com/user-attachments/assets/ecfe23c5-646b-47d9-8339-a92a15e3ac53" />


I feel this could be the default UI for login but if people want to keep the previous UI, we could have a VUE_APP_SIMPLE_LOGIN to switch it on (default off, i.e. current "view"/behaviour)

ml-evs:
Looks great! @BenjaminCharmes already did some work in this with a few splash screen options, unfortunately we were spoiled for choice and never made an individual one of them default...

PR #907, opened 2024-10-04
Add Login Splash Screen

BenjaminCharmes:
Closes #767 

- Login splash screen that displays the login options (GitHub, ORCID, and email)
- Updated route handling to redirect unauthenticated users to the /login page, with the exception of the /about page, which remains accessible without authentication.

PR #1911, opened 2026-07-12
Enable customizable login page

davidwaroquiers:
This PR adds an opt-in VUE_APP_ENABLE_LOGIN_PAGE mode that shows the existing Login.vue splash-style login page before authentication, instead of exposing the normal unauthenticated app shell.

When enabled, the login page only fetches the available authentication mechanisms through a minimal /info/auth-mechanisms endpoint. The login buttons are rendered from those enabled mechanisms, with an option to hide unavailable providers via VUE_APP_LOGIN_HIDE_UNAVAILABLE_AUTH.

The PR also adds optional customization for the left login information panel using the same pattern as CustomAbout.vue: deployments can provide public/custom/components/CustomLoginInfo.vue, while the default login design remains unchanged if no custom component is present.

Documentation for the new options and customization behavior is added in config.md.

Based on #907 

Solves #1907 (at least partially, we're still exposing the full info route right now)

Default screen is the same/similar to the one in #907 (Login.vue):

image

Providing a custom CustomLoginInfo.vue in public/custom/components, e.g.:

```
<template>
  <div class="example-login-info">
    <div class="example-logo" aria-hidden="true">AL</div>
    <h1>Acme Labs datalab</h1>
    <p class="lead">Research data management for materials, chemistry, and process R&amp;D.</p>
    <p>
      A workspace for connecting experimental results, simulations, analytics, and project
      knowledge into reusable research data.
    </p>
    <a class="btn btn-default" href="https://example.com/" target="_blank" rel="noopener">
      Visit Acme Labs
    </a>
  </div>
</template>

<script>
export default {
  name: "CustomLoginInfo",
};
</script>

<style scoped>
.example-login-info {
  max-width: 34rem;
  color: #15212f;
}

.example-logo {
  display: inline-flex;
  width: 5rem;
  height: 5rem;
  margin-bottom: 2rem;
  align-items: center;
  justify-content: center;
  border-radius: 50%;
  background: #15212f;
  color: white;
  font-size: 1.6rem;
  font-weight: 700;
}

h1 {
  margin-bottom: 1rem;
  font-size: 3.5rem;
  font-weight: 700;
}

.lead {
  font-size: 1.35rem;
  font-weight: 600;
}
</style>

```

allows to customize the left part of the login page:

image

With VUE_APP_HIDE_UNAVAILABLE_AUTH=true:

image


webapp/src/router/index.js
```diff
@@ -30,9 +30,9 @@ const routes = [
     component: Samples,
   },
   {
-    path: "/next/login",
+    path: "/login",
     name: "login",
-    alias: "/",
+    alias: "/next/login",
```
davidwaroquiers, line 35: not sure if we need the "next aliases" as the splash screens were inactive before.

webapp/src/components/LoginInfo.vue
```diff
@@ -0,0 +1,21 @@
...
+    datalab is open source (MIT license) and development occurs on GitHub at
+    <a href="https://github.com/datalab-org/datalab"
+      ><font-awesome-icon :icon="['fab', 'github']" />&nbsp;datalab-org/datalab</a
+    >
+    with documentation available on
+    <a href="https://the-datalab.readthedocs.io"
+      ><font-awesome-icon icon="book" />&nbsp;ReadTheDocs</a
+    >.
+  </p>
+  <router-link to="/about" class="btn btn-default">Learn More</router-link>
```
jbouquiaux: "Learn More" button that links to /about bounces back to /login page when logged-out. Button does visibly nothing in this case. Since this login page is primarly visited by logged-out users, I'm wondering if it's worth keeping it. /about page is easily accessible once logged-in anyway

jbouquiaux (commented):
I tested a few login customizations, this seems to work well, thanks! I'd have two comments:

- I tried to open /edit/<id> while logged out --> redirected to /login?next=/edit/<id> --> log in --> I land on the home page instead of back on the item. It seems that the ?next= does not work as expected, or am I missing something?


- Should we keep the"Learn more" button in the default LoginInfo.vue? See in-code comment

webapp/src/components/LoginInfo.vue
davidwaroquiers: I changed to link to the datalab docs

davidwaroquiers:
> I tested a few login customizations, this seems to work well, thanks! I'd have two comments:
> 
> * I tried to open /edit/ while logged out --> redirected to /login?next=/edit/ --> log in --> I land on the home page instead of back on the item. It seems that the ?next= does not work as expected, or am I missing something?
> * Should we keep the"Learn more" button in the default LoginInfo.vue? See in-code comment

the next indeed does not work for now. It would require some additional plumbing (because of the Oauth). I'd keep that for another PR.

davidwaroquiers:
This is ready for review @ml-evs @be-smith

DianaAliabieva (approved):
Hello! For me everything looks great. I agree with  @jbouquiaux about /next pages, but since he already mentioned it, so its fine.

davidwaroquiers:
I've merged main into this branch so it should be ready to be merged.
@ml-evs @be-smith

davidwaroquiers:
I've added the plumbing to get the redirect after Oauth, it wasn't so complex in the end. Rebased on main and self-reviewed. Should be ready to be for full review and merge once approved.

Also sharing a review guide to help:

[review_guide_splash_login_page.pdf](https://github.com/user-attachments/files/32250976/review_guide_splash_login_page.pdf)

And its corresponding md file:

[review_guide_splash_login_page.md](https://github.com/user-attachments/files/32250994/review_guide_splash_login_page.md)

pydatalab/src/pydatalab/routes/v0_1/info.py
```diff
@@ -122,6 +122,31 @@ def get_info():
     )
 
 
+@INFO.route("/info/auth-mechanisms", methods=["GET"])
```
ml-evs: Was there a reason not to use the existing /info/ endpoint for this that already serves this data?

pydatalab/docs/config.md
```diff
@@ -26,6 +26,8 @@ These can be provided as either:
         - `VUE_APP_WEBSITE_TITLE`: the title of the web app, which is displayed in the browser tab and header.
         - `VUE_APP_QR_CODE_RESOLVER_URL`: the URL of a service that can resolve QR codes to *datalab* entries, which is used by the web app to display QR codes for entries (see [datalab-org/datalab-purl](https://github.com/datalab-org/datalab-purl) for more information).
         - `VUE_APP_AUTOMATICALLY_GENERATE_ID_DEFAULT`: whether to automatically generate IDs for new entries in the web app by default, or require a checkbox to be ticked at item creation.
+        - `VUE_APP_ENABLE_LOGIN_PAGE`: whether unauthenticated users should be directed to the dedicated login page. If unset or set to `false`, the web app keeps the original unauthenticated routing behaviour.
+        - `VUE_APP_LOGIN_HIDE_UNAVAILABLE_AUTH`: whether the dedicated login page should hide unavailable authentication mechanisms. If unset or set to `false`, unavailable mechanisms are shown but disabled.
```
ml-evs: I think I'd prefer this to just be the default, rather than an option

ml-evs (commented):
Thanks for this @davidwaroquiers -- couple of comments below that I will implement myself before merging. I'm also going to get rid of the old stub login2 and login3 components as these are no longer accessible when the login page is turned on.

I think we can then iterate on the style of the login page as you prefer, so if you have any real examples we'd be happy to receive them back!

ml-evs (approved):
Hi @davidwaroquiers, thanks again for this. The functionality for getting the user to login when they receive an unauthorized request will be super useful and I think we should turn this on by default pretty soon (or even remove the option to have it turned off!)

As mentioned in my email, I don't really like the wide screen page that we had before so have refactored it into a much simpler small window as below:

Image

I don't know how this looks for your customisations, but I think it should still be possible to entirely customise it via the login component override. If anything it should be easier, as you just need to include a link to the login dropdown slot (or manually recreate it yourself).

As mentioned above I also reverted to use the existing info endpoint (we should lock it down properly if we want to, and reintroduce your minimal one for this case in a separate PR) and also removed the old stubs.

I'm going to merge this once the tests pass, but do let me know if this no longer fulfills your requirements and we can adjust straight away.

PR #1914, opened 2026-07-13
Refactor DynamicDataTable

### LINKED ISSUE #1712: Refactoring the dynamic data table
The dynamic data table component has grown into a bit of behemoth that is now quite tricky to extend with quite a few hardcoded behaviours for certain columns:

  - The template contains a long `v-else-if` chain of filter slots keyed on `column.field` values (`creatorsAndGroups`, `collections`, `type`, `status`, `location`/`supplier`, `date`, `created_at`, `account_status`, `role`, `groups`, `active`, `item_type`, `created_by_info`, …) plus `dataType`-specific branches (`users`, `tokens`).
  - The `filters` data object pre-declares a fixed set of fields with bespoke `matchMode` strings, and `created()` registers a corresponding fixed set of `FilterService` matchers (`exactCreatorMatch`, `exactBlockMatch`, `dateRange`, etc.).
  - `getComponentProps()` is a hardcoded registry mapping every body component (`FormattedItemName`, `BlocksIconCounter`, `UserRoleCell`, …) to the props it needs.
  - Per-table `unique*` computeds (`uniqueCreators`, `uniqueCollections`, `uniqueStatus`, `uniqueTokenCreators`, …) bake assumptions about the shape of `data` into the generic table.
  - Consumers like `StartingMaterialTable.vue` only declare a column list, but the table silently depends on that list lining up with all the special cases above. Adding a new item type means editing `DynamicDataTable.vue`, not just the consumer.

This makes the table hard to reason about, hard to test, and a blocker for supporting arbitrary item types defined by external schemas.

## Proposal: invert control

  Push column behaviour out of `DynamicDataTable` and into the column definition itself. A column descriptor should be able to declare:

  - `field`, `header`, `label`, `icon`, `sortable`, `hidden` (already present)
  - `body`: a component *or* render function, plus the props mapping (replacing the central `getComponentProps` registry)
  - `filter`: an object describing
    - the filter component to render in the slot (e.g. `MultiSelectFilter`, `DateRangeFilter`, `TextFilter`)
    - the `matchMode` (built-in or a matcher function provided inline)
    - an `options` provider — either a static array or a function `(data) => options[]` (replacing the `unique*` computeds)
  - optional `getValue(row)` for derived/virtual fields (e.g. `creatorsAndGroups`)

  Sketch of the consumer side:

  ```js
  {
    field: "collections",
    header: "Collections",
    body: { component: CollectionList, props: (row) => ({ collections: row.collections }) },
    filter: {
      component: CollectionFilter,
      match: (rowValue, filterValue) => /* … */,
      options: (data) => uniqueBy(data.flatMap(d => d.collections ?? []), "collection_id"),
    },
  }
  ```

  `DynamicDataTable` then becomes a thin wrapper over PrimeVue `DataTable` that:

  1. Iterates column descriptors and renders `<Column>` with generic `#body` / `#filter` slots that delegate to whatever the descriptor
  specifies.
  2. Builds the `filters` object dynamically from the descriptors instead of a hardcoded list.
  3. Registers `FilterService` matchers from descriptors on mount, scoped to that table instance.

BenjaminCharmes:
Closes #1712 

## Summary

- DynamicDataTable is now a thin, generic wrapper over PrimeVue DataTable — no more hardcoded field names, v-else-if filter chains, or central getComponentProps registry
- Column descriptors own their body component, filter component, match function, and options provider; adding a new item type only requires editing the consumer
- Extracted reusable filter components (TextFilter, MultiSelectFilter, SingleSelectFilter, DateRangeFilter, CreatorsAndGroupsFilter) and shared match/options helpers into `utils/filterMatchers.js`
- Dynamic FilterService matcher registration from descriptors, scoped per dataType to avoid global collisions
- Cleanup: allUsers passed as an explicit prop, body events forwarded generically via column.body.events, filterOptionsCache receives store.state rather than the full store


BenjaminCharmes:
@ml-evs I forgot to ping you for a review!

ml-evs:
Also pining @DianaAliabieva and @davidwaroquiers for review here as they have been working on this component and will likely need to base their changes on this!

ml-evs (changes requested):
Thanks for this @BenjaminCharmes, looks great! Would you mind incorporating the changes from #1910 before we merge this? I had a look and not entirely sure how I would do it with the current conflicts. Thanks!

BenjaminCharmes:
> Thanks for this @BenjaminCharmes, looks great! Would you mind incorporating the changes from #1910 before we merge this? I had a look and not entirely sure how I would do it with the current conflicts. Thanks!

Yes, I'll have a look !

BenjaminCharmes:
@ml-evs Everything should be good now, let me know if you need anything else!

davidwaroquiers:
> Also pining @DianaAliabieva and @davidwaroquiers for review here as they have been working on this component and will likely need to base their changes on this!

Hi @ml-evs 

Thank you for pinning us. I don't have time to make a thorough review right now (currently on holidays) but here is maybe my comment after a very quick look. This is a very nice work to make the Dynamic data table more generic and thanks @BenjaminCharmes for this. I was wondering if the associated table buttons could (or should) also be somehow described declaratively in the actual tables rather than being hardcoded in DynamicDataTable/DynamicDataTableButtons. Somehow the goal of the PR is to extract all references to actual tables from the DynamicDataTable component to the actual components (e.g. EquipmentTable, SampleTable, ...). What do you think ?


I can make a more thorough review when I'm back in the week of 24-28 August but if you'd prefer to get this merged sooner, please go ahead. Regarding PRs that are somehow related to this one (from me and/or @DianaAliabieva), we will refactor them based on these changes, I think the best is to get this through and rebase/refactor our own PRs.

webapp/src/components/CollectionInformation.vue
```diff
@@ -100,38 +108,138 @@ export default {
...
+              formula: row.chemform,
+              smiles: row.smiles,
+              inchiKey: row.inchi_key,
+              ghsCodes: row.GHS_codes,
+              molarMass: row.molar_mass,
+              cas: row.CAS,
+            }),
+          },
+        },
+        {
```
DianaAliabieva, line 192: I think this file got missed during the filter migration. On main, these three columns all had filter: true:


{ field: "date", header: "Date", label: "Date", filter: true },
{ field: "creatorsAndGroups", ..., filter: true },
{ field: "blocks", ..., filter: true },
In the refactored version they only kept getValue/body for rendering the filter key is just gone, and none of DateRangeFilter, CreatorsAndGroupsFilter, or matchBlocks/blocksOptions are imported at the top of the file (only TextFilter and MultiSelectFilter made it in, for the id/type/status columns).

So they are lost kinda, maybe we should return them? or it was a plan to do like that?

DianaAliabieva (commented):
Hello! I tried to review this PR!
For me everything looks great, only one thing about lost filter values
Have a nice day!

webapp/src/components/DynamicDataTable.vue
davidwaroquiers, line 475: Not directly related to this PR itself but I found out while reviewing. This seems like a bug. Identifying a specific row is based either on item_id or collection_id but for the other tables (users, groups, access tokens), e.g. those of the admin view, the key for identifying a row is different (immutable_id, group_id, _id). This leads to a bug when selecting users/groups/tokens.

To reproduce the bug, go for example in users and select one user:

Image

You get "1 selected" for the dropdown menu of bulk actions.

Then do select all (top left checkbox to select all):

Image

And only the previously selected user stays selected, others stay unselected (and the dropdown menu shows "1 selected").

The easiest fix would be to add a id_key string to specify which key is used to identify a row (user/item/group/...) uniquely.

Although this is not directly related to this refactoring I think it would be good to fix it here already.

webapp/src/components/CollectionInformation.vue
davidwaroquiers: I think the filters should indeed stay the same unless this is intentional ?

webapp/src/components/DynamicDataTable.vue
davidwaroquiers, line 11: Not directly related to this PR but probably easy to fix it here (or in another PR).

Currently if you select all rows (using the top left checkbox) then deselecting just one row keeps the selected all checkbox checked. 

I think allSelected should be recomputed both upon row selection and deselection.

davidwaroquiers (changes requested):
Hi @BenjaminCharmes and @ml-evs 

Thanks for the nice work, this was indeed a good idea to refactor this big chunk of code that was becoming more and more complex. I have a couple of small comments in the code, not directly related to this PR but that I believe are worth fixing here anyway (should be easy to fix, could even be in a separate PR if easier or want to completely separate concerns).


I already mentioned this in a comment, but a nice extension of this work would be to apply the same declarative approach to the toolbar buttons and bulk actions. The columns are now declared by each concrete table, while the toolbar and bulk actions are still centralized and selected through dataType-specific conditions. Each table could instead declare the actions it supports (including their label, visibility, disabled state, and handler) leaving the generic components responsible only for rendering them and managing shared UI. This could also be considered as a second phase or a separate PR if it feels too broad for this refactoring but I think it would make things even more clean and modular.

Happy to discuss if needed.

webapp/src/components/CollectionInformation.vue
ml-evs: I believe this one is now resolved

webapp/src/components/DynamicDataTable.vue
ml-evs: Opened #2090 to track this, think its quite a minor edge case so don't want to hold up this PR any further

PR #1918, opened 2026-07-13
Lower erroneously open permissions of block updates if IDs are known

### LINKED ISSUE #1912: Update blocks is too open
Digging a bit into the blocks routes, I think the update-block route allows any authenticated user to update a block of another user if he knows the id of that block. The delete routes do not have this issue (they have get_default_permissions).

Probably not of super high importance but still, this is a bug I think. Maybe this should be solved once #48 is solved ?

### LINKED ISSUE #1919: Remove aborted support for blocks to operate on collections
We played around with this a while ago, but never enabled the web app routes, and this leaves a bit of a messy permissions issue to clear up so I'd rather remove it until we can reintroduce it properly in a new form.

Probably @jdbocarsly and @OMWalmsley are the only two who realised this code existed still (and Copilot in the review of #1918...)

ml-evs:
Closes #1912
Closes #1919

The proper fix is #48 and related changes but we need this in now. Likely we should add a test that also does this for an async task block, though the guard is added at the start of the route. I've also removed the leftover code for creating blocks that pertain to collections, as this has the same issue, but is harder to fix, and no block was even using this code properly.

### DIFF
--- a/pydatalab/docs/blocks/index.md
+++ b/pydatalab/docs/blocks/index.md
@@ -30,10 +30,8 @@ Data blocks are modular components that:
 The block system exposes several API endpoints:
 
 - `/add-data-block/`: Create and add a new block to an item
-- `/add-collection-data-block/`: Create and add a new block to a collection
 - `/update-block/`: Update an existing block's state
 - `/delete-block/`: Remove a block from an item
-- `/delete-collection-block/`: Remove a block from a collection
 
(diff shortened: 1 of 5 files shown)


ml-evs:
I've reviewed this myself and with two additional tools; I'm happy to bear the responsibility of merging this and releasing what is an easy-to-exploit but low risk bug.

PR #1921, opened 2026-07-14
Fix user login query in case where user has been deleted

### LINKED ISSUE #1920: Potential internal server error at login
Loosly related to #1915. I tried to manually delete a user, but when trying to log in again with the same account, the fact that the browser had the old cookies caused a `StopIteration` error here:
https://github.com/datalab-org/datalab/blob/71b6c9805752cdcdecf4c43e9bee9949d7b35902/pydatalab/src/pydatalab/login.py#L126

The error is not handled and surfaces in the browser as "Internal server error". It is not something that should commonly occur, but is probably worth fixing anyway in light of future potential introduction of user deletion.

gpetretto:
Fixes #1920.
I think it is worth to make this simple check to avoid the backend error.

### DIFF
--- a/pydatalab/src/pydatalab/login.py
+++ b/pydatalab/src/pydatalab/login.py
@@ -118,12 +118,16 @@ def get_by_id(user_id: str) -> LoginUser | None:
 
     """
 
-    user = flask_mongo.db.users.aggregate(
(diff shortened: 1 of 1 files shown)


ml-evs (approved):
Thanks @gpetretto!

PR #1922, opened 2026-07-14
Fix loose permissions when creating a sample by copying from another

davidwaroquiers:
This fixes a security issue in which it was possible for a user (user A) to create a new sample from a copy of a sample of another user (user B) to which user A does not have access (if user A knows the item id). This is not possible in the UI directly but the API still allowed it. Added get_default_permissions to the _copy_sample_from_id helper function and added tests.


ml-evs (approved):
Super helpful, thanks David! I'd put this in the same class as the previous bug, we should do a release soon with these changes but there's no drastic rush to deploy it everywhere given the low risk.

PR #1924, opened 2026-07-15
Fixed editing of roles (UI)

davidwaroquiers:
In the admin menu, when changing the role for a user from one role to another, it says "Are you sure .... from "admin" to "admin". It repeats the "final" role. Also, if we do Cancel, the role seems to be changed in the UI but it's actually not (which is normal, it should not be changed if we do Cancel). When we refresh the page then the correct role is shown (i.e. the one that hasn't changed).

This PR fixes this small UI issue.


ml-evs (approved):
Thanks for the nice fix @davidwaroquiers! I also removed the double dialog that occured when demoting an admin (it previously never triggered due to this bug).

PR #1927, opened 2026-07-16
Fix tz-naive datetime for certain routes

be-smith:
MongoDB stores datetimes as UTC instants, but pymongo doesn't attach that timezone information back onto the datetime objects it returns — they come back naive (tzinfo=None), regardless of whether the value was originally written as timezone-aware.

Most routes have a pydantic model validation stage where naive datetimes have utc reattached if there is no timezone info. Some routes don't have a pydantic model validator and are simply aggregations, so the datetimes are serialised as is. Frontend code that parses this string with new Date(...) then interprets it as local time rather than UTC, showing the wrong time to any user not in UTC.

Routes found affected:
- get_equipment_summary (/equipment/)
- list_versions (/items/<refcode>/versions/, backing the version history modal)

This fix makes CustomJSONEncoder assume UTC for any naive datetime before serializing it, matching what the Pydantic validator already does — so the same fix also protects any other raw/unvalidated route that haven't been specifically identified.

If you want to recreate the bug in the UI, in a timezone that isn't utc save a version, the version history modal then won't say "a few minutes ago" it will be however many hours your timezone is that isn't utc.

Another potential fix: pymongo.MongoClient has a tz_aware=True option, so every datetime read from the db would come back with timezone information attached. Maybe something to discuss as this could affect more code

### DIFF
--- a/pydatalab/src/pydatalab/utils/__init__.py
+++ b/pydatalab/src/pydatalab/utils/__init__.py
@@ -42,7 +42,17 @@ class CustomJSONEncoder(JSONEncoder):
 
     @staticmethod
     def default(o):
-        if isinstance(o, (datetime.date, datetime.datetime)):
+        if isinstance(o, datetime.datetime):
+            # MongoDB stores datetimes as UTC instants. With PyMongo's default
+            # tz_aware=False, those values are returned as naive datetimes even though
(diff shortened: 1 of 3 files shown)


ml-evs (approved):
Thanks @be-smith, looks good! Will merge this before #1910 to make sure the tests are still working, and then will likely need to check this works after the pydantic 2 PR.

PR #1928, opened 2026-07-17
Add docker development environment and CI coverage

### LINKED ISSUE #1902: Docker MongoDB path fails out of the box on macOS
docker-compose.yml bind-mounts the host path directly:

volumes:
  - /data/db:/data/db
INSTALL.md offers this as the alternative to a native MongoDB install (docker compose up database), but on macOS with Docker Desktop this fails immediately:

Error response from daemon: mounts denied:
The path /data/db is not shared from the host and is not known to Docker.

The user has to manually go into Docker Desktop → Settings → Resources → File Sharing and add /data/db — a step INSTALL.md doesn't mention at all. I hit this firsthand this session. A named Docker volume (e.g. mongo-data:/data/db) instead of a host bind-mount would avoid the platform-specific permission dance entirely and is the more idiomatic Compose pattern anyway.

This is install-blocking like #1 and #2, just confined to the Docker path rather than the native path — still worth raising since it's the path INSTALL.md explicitly recommends for anyone who doesn't want to install MongoDB natively.

jbouquiaux:
This PR extracts the Docker development environment and CI coverage from https://github.com/datalab-org/datalab/pull/1845 so they can be reviewed independently from environment-variable handling and seed snapshots.

### DIFF
--- a/.docker/app/Dockerfile
+++ b/.docker/app/Dockerfile
@@ -31,7 +31,6 @@ RUN apt update \
         xauth \
         xvfb \
     && rm -rf /var/lib/apt/lists/*
-COPY webapp ./
 
 # build target: runs the production webpack bundle. Derives from `base` rather
 # than `dev` so production builds don't pull in the Cypress system libraries.
--- a/.docker/app/Dockerfile
+++ b/.docker/app/Dockerfile
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile:1
-FROM node:22.22-bullseye-slim AS build
+FROM node:22.22-bullseye-slim AS base
 SHELL ["/bin/bash", "--login", "-c"]
 
 WORKDIR /app
@@ -13,6 +13,30 @@ COPY webapp/package.json webapp/yarn.lock ./
 # Using a custom node_modules location to avoid mounting it outside of docker
 RUN --mount=type=cache,target=/root/.cache/yarn yarn install --frozen-lockfile --modules-folder /node_modules
 
+# dev target: deps installed, source mounted at runtime — no production build.
+# Also installs the system libraries the Cypress test runner needs (GTK/X11
+# runtime libs plus xvfb for a virtual display), so the webapp test suites can
+# run inside the container.
+FROM base AS dev
+RUN apt update \
+    && apt install -y --no-install-recommends \
+        libgtk2.0-0 \
+        libgtk-3-0 \
+        libgbm-dev \
+        libnotify-dev \
+        libnss3 \
+        libxss1 \
+        libasound2 \
+        libxtst6 \
+        xauth \
+        xvfb \
+    && rm -rf /var/lib/apt/lists/*
+COPY webapp ./
+
+# build target: runs the production webpack bundle. Derives from `base` rather
+# than `dev` so production builds don't pull in the Cypress system libraries.
+FROM base AS build
+COPY webapp ./
 # These get replaced by the entrypoint script for production builds.
 # Set the real values in `.env` files or an external docker-compose.
 ENV NODE_ENV=production
@@ -28,7 +52,6 @@ ARG VUE_APP_AUTOMATICALLY_GENERATE_ID_DEFAULT=magic-generate-id-setting
 ARG VUE_APP_GIT_VERSION=0.0.0+ci
 ENV VUE_APP_GIT_VERSION=${VUE_APP_GIT_VERSION}
 
-COPY webapp ./
 RUN /node_modules/.bin/vue-cli-service build
 
 FROM node:22.22-bullseye-slim AS production
--- a/.docker/app/Dockerfile
+++ b/.docker/app/Dockerfile
@@ -32,6 +32,10 @@ RUN apt update \
         xvfb \
     && rm -rf /var/lib/apt/lists/*
 
+CMD [ "/node_modules/.bin/vue-cli-service", "serve", "--host", "0.0.0.0", "--port", "8081" ]
+
+HEALTHCHECK --interval=30s --timeout=5s --start-period=5s --retries=3 CMD node -e "fetch('http://localhost:8081').then(r => process.exit(r.ok ? 0 : 1)).catch(() => process.exit(1))"
+
 # build target: runs the production webpack bundle. Derives from `base` rather
 # than `dev` so production builds don't pull in the Cypress system libraries.
 FROM base AS build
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -196,6 +196,63 @@ jobs:
...
+
+      - name: Show dev service diagnostics
+        if: failure()
+        run: |
+          docker compose --profile dev logs
+          docker compose --profile dev ps
+
+      - name: Stop dev services
+        if: always()
+        run: docker compose --profile dev down --volumes
```
davidwaroquiers, line 254: Should it have a test that it indeed reloads when there is a change ?

pydatalab/docs/INSTALL.md
```diff
@@ -30,6 +30,35 @@ If you are not familiar with `git` or GitHub, you can do worse than reading thro
...
+```shell
+docker compose exec api-dev /opt/.venv/bin/invoke admin.change-user-role \
+  --display-name "Your Name" --role admin
+```
+
+The development database and uploaded files are stored in separate Docker volumes. To use
+a different database name, set `DATALAB_DB_NAME` when starting the stack, for example:
+
+```shell
+DATALAB_DB_NAME=my-project docker compose --profile dev up
```
davidwaroquiers: Would/could that be useful in prod too ?

docker-compose.yml
```diff
@@ -57,6 +57,82 @@ services:
...
+    command:
+      ["/node_modules/.bin/vue-cli-service", "serve", "--host", "0.0.0.0", "--port", "8081"]
+    volumes:
+      - ./webapp:/app
+    ports:
+      - "8081:8081"
+    environment:
+      - NODE_ENV=development
+      - VUE_APP_API_URL=${VUE_APP_API_URL:-http://localhost:5001}
+      - CHOKIDAR_USEPOLLING=${CHOKIDAR_USEPOLLING:-true}
```
davidwaroquiers: Correct me if I'm wrong, but hot reloading is activated by this variable. I would maybe always set it to true no ?

.docker/app/Dockerfile
```diff
@@ -13,6 +13,30 @@ COPY webapp/package.json webapp/yarn.lock ./
...
+        libgbm-dev \
+        libnotify-dev \
+        libnss3 \
+        libxss1 \
+        libasound2 \
+        libxtst6 \
+        xauth \
+        xvfb \
+    && rm -rf /var/lib/apt/lists/*
+COPY webapp ./
```
davidwaroquiers: I think this is not needed as webapp is mounted as a volume directly in the docker compose

davidwaroquiers (dismissed):
Thanks!
Looks good to me, I just made a few comments. I let you decide what to do on these, then I think you can undraft.

.docker/app/Dockerfile
jbouquiaux: I agree, I'll remove it.

docker-compose.yml
jbouquiaux: I agree

pydatalab/docs/INSTALL.md
jbouquiaux: Maybe... although in prod you can already select the database through the PYDATALAB_MONGO_URI setting, with the database name included in the URI. 
I’d keep DATALAB_DB_NAME as a quick convenience to the development profile for now.

One future possibility would be to make the docker compose more flexible with something like:
`PYDATALAB_MONGO_URI: ${PYDATALAB_MONGO_URI:-mongodb://database:27017/datalabvue}`

but I think this would belong to another PR.

.github/workflows/ci.yml
jbouquiaux: I actually took inspiration from the previous implementation referenced in the [original review on #1845](https://github.com/datalab-org/datalab/pull/1845#pullrequestreview-4555344020), where this was not done.

I believe that testing hot reload would add too much complexity (cypress test for browser reload, dealing with time-out,...)? I would keep it like that for now.

jbouquiaux:
Hi @ml-evs, @OMWalmsley,

This is now ready for review. Happy to have your opinion on the two comments I left open.

pydatalab/docs/INSTALL.md
ml-evs: Not sure I like this setting popping up here, for one it doesn't match our usual format ("DATALAB_" vs "PYDATALAB_" vs "VUE_APP_") -- could we remove for now and add it properly in a future PR? e.g., `DB_NAME = "datalabvue"` by default and can either be set by MONGO_URI or overrided by the new setting?

pydatalab/docs/INSTALL.md
```diff
@@ -30,6 +30,35 @@ If you are not familiar with `git` or GitHub, you can do worse than reading thro
...
+This starts the web app at [http://localhost:8081](http://localhost:8081), the API at
+[http://localhost:5001](http://localhost:5001), and MongoDB at
+`mongodb://localhost:27018`. Changes under `pydatalab/` and `webapp/` are
+bind-mounted into the containers and trigger the respective development servers to reload.
+
+After connection, you can make yourself admin with the command
+```shell
+docker compose exec api-dev /opt/.venv/bin/invoke admin.change-user-role \
+  --display-name "Your Name" --role admin
+```
```
ml-evs: Unless I've missed something, you still need to register an account before this will work?

ml-evs (commented):
Couple of comments below, but this is looking nearly ready to merge!

.github/workflows/ci.yml
davidwaroquiers: true. let's keep it like this.

pydatalab/docs/INSTALL.md
davidwaroquiers: Done (@jbouquiaux is on holidays, I took the liberty of changing it myself)

pydatalab/docs/INSTALL.md
davidwaroquiers: True, I changed "After connection" to "After registration", which is more correct.

davidwaroquiers:
@ml-evs I think this can be merged unless you think we are still missing something here or some comment has not been properly addressed. What do you think ?

ml-evs (dismissed):
All good with me @davidwaroquiers, I made a couple of minor restructurings so that all the `HEALTHCHECK` and `CMD` options now live in the Dockerfiles proper, rather than mixed across the Dockerfiles and docker-compose.yml.

Thanks both @jbouquiaux and @davidwaroquiers for this!

ml-evs (approved):

davidwaroquiers:
> All good with me @davidwaroquiers, I made a couple of minor restructurings so that all the `HEALTHCHECK` and `CMD` options now live in the Dockerfiles proper, rather than mixed across the Dockerfiles and docker-compose.yml.
> 
> Thanks both @jbouquiaux and @davidwaroquiers for this!

Thanks to you!

PR #1930, opened 2026-07-17
Add nominal capacity to cell

be-smith:
Adds theoretical/nominal capacity fields to Cell

Adds a theoretical_capacity field (mAh/g) to the Cell model. Combined with the existing characteristic_mass (active mass), it's used to compute nominal_capacity, shown and editable in its own unit (mAh/Ah) in the UI, and updates live as you type — no save round-trip needed.

The computed value is always normalized and additionally stored in nominal_capacity_mah, so plugins/API consumers can read a unit-independent value without having to handle nominal_capacity_unit conversions themselves.

nominal_capacity/nominal_capacity_mah are derived, not user-settable — they're recomputed server-side from theoretical_capacity * characteristic_mass on every save, so the stored value can't drift from the inputs.

Test plan
- test_cell_nominal_capacity covers: missing inputs, unit conversion (mAh vs Ah), that a client-supplied nominal_capacity is ignored/overwritten, and round-tripping through save/load.Adds a theoretical capacity field that is used combined with active mass to calculate the nominal capacity of the cell.

Stored in `nominal_capacity_mah` for use with plugins or the API


ml-evs:
Superseded by #2091

PR #1936, opened 2026-07-21
Separate blocks from items

gpetretto:
Implementing the separation of blocks from the items as discussed in #48.

They key principles followed for the implementation are those outlined in https://github.com/datalab-org/datalab/issues/48#issuecomment-4811223759 (copy/pasted below for convenience)

<details>
<summary>Initial planning</summary>

1. **No new functionality**. The initial PR focuses exclusively on preserving existing functionalities with the blocks separated from the items. Some choices for the document structure may be driven by functionalities that are planned in the near future, but no additions here.

2. **Blocks as separated entities**. blocks are similar to `File`: the new block Model has parents `Entry + HasOwner + HasRevisionControl`. 
 
3. **Two collections: `data_blocks` + `block_versions`.** Similar to `items`/`item_versions`. An item will always point to the document in the `data_blocks` collection (i.e. the latest block version). Document in `item_versions` points to documents in `block_versions`. 

4. **Items collection**: the `blocks_obj` in the `items` collection can contain either the full block for backward compatibility or a dictionary with `{"immutable_id": ObjectId}`. The order will still be based on `display_order`. This is preparatory for the subsequent steps in which the blocks_obj` and `display_order` will be replaced by `{"blocks": [{"immutable_id": ...}]` to fit the suggested pattern (`{"type": "my-item", "blocks": [{"immutable_id": <ObjectId>}], "files": [{"immutable_id": <ObjectId>}]}`).
 
5. **No reference to the items in the block document**. Can be added in the future if needed.
 
6. **Versioning**. Like in the current data structure, modifying a block will not automatically trigger the creation of a new block version in  `block_versions`, but will just modify the document in `data_blocks`. A new version of the block in `block_versions` will be triggered by an item save, but only if the content of the block has changed from the previous block version stored, to avoid duplicated entries.
 
7. **Restore**. Same mechanism as items. Restoring an old item version writes the old block content back in the current state + a new history entry. At this stage this can happen only when restoring an Item entirely.
 
8. **No changes to File ownership**. Most blocks are associated to a file. Association will remain the same.
 
9. **No changes to the frontend API**. Responses still expose `blocks_obj` + `display_order`, reassembled server-side from `data_blocks`.
 
10. **Block deletion**. As in the current implementation, if a block is deleted it will be removed from the `data_blocks` collection, but preserved in the `block_versions` in case it needs to be restored. Block deletion triggered only when removed from an item. If an Item is deleted, the corresponding blocks in `data_blocks` are deleted (not those in `block_versions`, mimicking what happens for items). This can revised in the future if a block can be shared.

11. **Collections**. `Collection` is `HasBlocks`, but if I am not mistaken there is really no place where a block can be assigned to a collection at the moment in the UI, and it has no versioning. We keep the same structure as Items, but there should be no need to handle backward compatibility.

</details>

A few additional points to be specified concerning the details of the implementation:
* The access to the blocks is currently gated by the access to the item. There is no direct access to the blocks.
* The wat the blocks are stored is preserved. A legacy block will never be changed in a new kind of block here.
* It is not clear to me if for the collections the backward compatibile version should be supported. And at the moment it seems that there is no way of adding blocks to a collection. For this reason The collection blocks are left untouched. Adding the new blocking version with or without full backward compatibility should be trivial.
* Consistently with the current implementation, when a new block is created in the DB it is not versioned until the item to which it belongs to is saved. To handle this properly the `version` attribute of the block starts at `0` and becomes `1` when the item is saved.
* fixed a pre-existing (and apparently known issue): creating and then immediately deleting a block, did not truly remove it. Upon a save the removed block was recreated. Since it was related to the block management it seemed fine to add this fix here.

Note that I used the pydantic2 branch as a base for the implementation, as it seems reasonable as base, given the involvement of the backend. I targeted the PR to that branch, but the idea is to merge this in main after the pydantic2 branch has been merged.


pydatalab/src/pydatalab/models/blocks.py
```diff
@@ -72,3 +76,40 @@ class DataBlockResponse(BaseModel):
...
     experimental acquisition parameters."""
+
+
+# Here to avoid circular import
+class HasBlocks(BaseModel):
+    blocks_obj: dict[str, DataBlockResponse] = Field({})
+    """A mapping from block ID to block data."""
+
+    display_order: list[str] = Field([])
+    """The order in which to display block data in the UI."""
```
davidwaroquiers, line 89: Not sure I like it to have HasBlocks moved here somehow. It could be possible to move DataBlockResponse to another module (e.g. blocks_response.py or something) and keep HasBlocks in traits.py.
Not sure if it's worth. Maybe there is another solution also ? Maybe @ml-evs has an idea/preference or can comment on this.

pydatalab/src/pydatalab/models/blocks.py
```diff
@@ -72,3 +76,40 @@ class DataBlockResponse(BaseModel):
...
+    parent item's `blocks_obj`/`display_order` and in the DOM."""
+
+    blocktype: str
+    """A short string key specifying the type (technique) of the block."""
+
+    data: dict[str, Any] = Field(default_factory=dict)
+    """The block payload, exactly as produced by `DataBlock.to_db()`."""
+
+    version: int = 0
+    """The latest committed version number of this block in `block_versions`.
```
davidwaroquiers, line 113: Wouldn't it be better to somehow follow a logic similar to items ? i.e. having a counter in the version_counter collection based on the immutable_id of the block ? What do you think ? I am actually a bit confused about the version that is in the item and the counter that is in the version_counter in the end. It seems they are always the same (version in item and counter in the corresponding document in version_counter). Probably good to clarify this with @ml-evs 

After our discussion and regarding the logic of having version = 0 here, I propose we also discuss this with @ml-evs 

Somehow we keep exactly the same behavior as it is currently but I have the feeling that we may want to change that somehow (to a better, more consistent behavior).

Behavior:
When we add a block, it already adds the block directly in the item without any save.

What I would expect (to be discussed of course :):
When we add a block in the item, it will add the block only when we save the item.

pydatalab/src/pydatalab/blocks/store.py
```diff
@@ -0,0 +1,388 @@
...
+    return datetime.datetime.now(tz=datetime.timezone.utc).isoformat()
+
+
+def _current_user_id() -> ObjectId | None:
+    if current_user and current_user.is_authenticated:
+        return current_user.person.immutable_id
+    return None
+
+
+def is_block_reference(entry: Any) -> bool:
```
davidwaroquiers: It's maybe a stupid thing but calling the argument entry here is a bit misleading for me because I initially "expected" to have an Entry object (a subclass of it) and then I was wondering how it could only have immutable_id and version as keys. Maybe better to rename it ? Something like "block_or_reference" ?

pydatalab/src/pydatalab/blocks/store.py
```diff
@@ -0,0 +1,388 @@
...
+    """Whether a stored `blocks_obj` entry is a reference to a `blocks` document
+    (``{"immutable_id": ...}``) rather than a legacy embedded block.
+    """
+    return (
+        isinstance(entry, dict) and "immutable_id" in entry and set(entry.keys()) <= _REFERENCE_KEYS
```
davidwaroquiers: Can it happen that "entry" is not a dict ? And if we assume it can, then this "non dict" object is then considered as a legacy embedded block. What would/could happen then ?

pydatalab/src/pydatalab/blocks/store.py
```diff
@@ -0,0 +1,388 @@
...
+    )
+
+
+def authorize_and_get_form(item_id: str | None, block_id: str) -> dict | None:
```
davidwaroquiers:
```suggestion
def authorize_and_get_block_or_reference(item_id: str | None, block_id: str) -> dict | None:
```
Changing the name to make it clearer ?

pydatalab/src/pydatalab/blocks/store.py
```diff
@@ -0,0 +1,388 @@
...
+            continue
+        doc = resolved_docs.get(referenced[block_id])
+        if doc is None:
+            LOGGER.error(
+                "Dangling block reference: blocks_obj[%r] points at missing blocks doc %s",
+                block_id,
+                referenced[block_id],
+            )
+            continue
+        loaded[block_id] = doc.get("data", {})
```
davidwaroquiers: I guess data should always have the DataBlock dict no ? If it doesn't maybe it should be similar to above, i.e. some kind of dangling block reference, and not be returned within the loaded dict ?

pydatalab/src/pydatalab/blocks/store.py
```diff
@@ -0,0 +1,388 @@
...
+        if doc is None:
+            LOGGER.error(
+                "Dangling block reference in version snapshot: blocks_obj[%r] -> %s",
+                block_id,
+                entry,
+            )
+            continue
+        resolved[block_id] = doc.get("data", {})
+
+    return resolved
```
davidwaroquiers: This has exactly the same structure as load_blocks_obj above. Does it make sense to have a common method ?

pydatalab/src/pydatalab/blocks/store.py
```diff
@@ -0,0 +1,388 @@
...
+    if last_version and last_version.get("data") == data:
+        return last_version["version"]
+
+    updated_doc = flask_mongo.db.blocks.find_one_and_update(
+        {"_id": immutable_id},
+        {"$inc": {"version": 1}},
+        return_document=ReturnDocument.AFTER,
+    )
+    if updated_doc is None:
+        return None
```
davidwaroquiers, line 296: I guess this should not happen except if there is something that happens between the find_one above and now. Would it be worth to use a transaction here (I'm thinking here but probably other places too). It also maybe adds too much complexity and also anyway transactions are not active on the current deployment stack (see issue I opened).

pydatalab/src/pydatalab/blocks/store.py
```diff
@@ -0,0 +1,388 @@
...
+    next_version = updated_doc["version"]
+
+    version_entry = BlockVersion(
+        block_immutable_id=immutable_id,
+        block_id=block_doc["block_id"],
+        version=next_version,
+        timestamp=datetime.datetime.now(tz=datetime.timezone.utc),
```
davidwaroquiers, line 303: Why not using _now_isoformat ?

pydatalab/src/pydatalab/blocks/store.py
```diff
@@ -0,0 +1,388 @@
...
+
+    # Restore the raw payload into 'data' so that None-valued fields are not
+    # stripped by model_dump(exclude_none=True), mirroring item snapshots.
+    version_doc = version_entry.model_dump(exclude_none=True)
+    version_doc["data"] = data
+    try:
+        flask_mongo.db.block_versions.insert_one(version_doc)
+    except DuplicateKeyError:
+        # A concurrent snapshot minted this version number first; both captured
+        # the same live payload, so re-pinning the number is safe.
```
davidwaroquiers, line 318: I am struggling to understand why it necessarily captures the same live block here. I would say that either there are concurrent saves and they will get different version numbers (thanks to the find_one_and_update with the increment) and then somehow the "largest versions wins" (and I guess it's fine ?) or there is something weird that happens outside of this function but then there is no way that things are somehow correct.

pydatalab/src/pydatalab/blocks/store.py
```diff
@@ -0,0 +1,388 @@
...
+        restored_from_version=pinned["_id"],
+        user_id=user_id,
+        datalab_version=__version__,
+        data=data,
+    )
+    version_doc = version_entry.model_dump(exclude_none=True)
+    version_doc["data"] = data
+    flask_mongo.db.block_versions.insert_one(version_doc)
+
+    return next_version
```
davidwaroquiers, line 401: Reading this part to understand the logic made me think once more that maybe it would be better to use an approach similar to items for the version counts with a separate collection holding the block versions (could be the version_counter collection or another one). What do you think ?

pydatalab/src/pydatalab/routes/v0_1/blocks.py
```diff
@@ -248,15 +249,34 @@ def add_data_block():
...
+    # the item only stores a reference to them. No version is committed until
+    # the next item version snapshot.
+    block_immutable_id = store.create_block_document(block)
+
     result = flask_mongo.db.items.update_one(
         {"item_id": item_id, **get_default_permissions(user_only=True)},
         {
             "$push": {"display_order": display_order_update},
-            "$set": {f"blocks_obj.{block.block_id}": block.to_db()},
+            "$set": {f"blocks_obj.{block.block_id}": {"immutable_id": block_immutable_id}},
```
davidwaroquiers, line 282: Why not defining the id then do a find_one_and_update with the permissions, followed by the create_block_document ? In that case there is one less query to the items collection. The drawback is that then if the block creation fails we have a dangling reference to a block ? I guess that's why you did it this way ? Maybe a transaction would also be useful here.

pydatalab/src/pydatalab/routes/v0_1/blocks.py
```diff
@@ -487,9 +530,15 @@ def delete_block():
             },
             "$unset": {f"blocks_obj.{block_id}": ""},
         },
+        projection={f"blocks_obj.{block_id}": 1, "display_order": 1},
```
davidwaroquiers, line 482: I always get confused that the returned document is the one before the update by default (I would "expect" somehow to have the new one by default). It's probably stupid but I'd maybe explicitly put return_document=ReturnDocument.BEFORE

pydatalab/src/pydatalab/routes/v0_1/blocks.py
```diff
@@ -487,9 +530,15 @@ def delete_block():
...
     )
 
-    if result.modified_count < 1:
+    stored_entry = (doc_before.get("blocks_obj") or {}).get(block_id) if doc_before else None
+
+    block_was_present = stored_entry is not None or (
+        doc_before is not None and block_id in (doc_before.get("display_order") or [])
```
davidwaroquiers, line 491: Can this happen that the block was not in the list of blocks of the items but it was in the display order ? That seems like a case where the block actually does not exist and cannot be deleted, just that there was some kind of bug or weird thing happening to the display order at some point.

pydatalab/src/pydatalab/versioning.py
```diff
@@ -162,6 +148,54 @@ def save_version_snapshot(
 
     action = VersionAction.MANUAL_SAVE if action is None else action
 
+    # Cut `block_versions` entries for any *referenced* blocks whose payload has
+    # changed since their last committed version, and pin each referenced
+    # `blocks_obj` entry to its now-current version number in this snapshot.
+    # Embedded blocks are captured verbatim in the item snapshot.
+    # This must happen before the de-duplication guard below so that a
+    # block content change is seen as an item change.
+    for block_id, entry in list((item.get("blocks_obj") or {}).items()):
```
davidwaroquiers: I would change the name of "entry" here, it confuses me :) (again :D)
Maybe "block_doc" or "block_obj" (I guess it would be good to have something "similar" in all places where such a thing occur)

pydatalab/src/pydatalab/routes/v0_1/items.py
```diff
@@ -1404,26 +1484,96 @@ def restore_version(refcode):
...
+        # do not validate as block payloads, so validate a *resolved* copy while
+        # keeping each entry's original form for the writes below.
+        validation_data = restored_data
+        if any(
+            block_store.is_block_reference(entry)
+            for entry in (restored_data.get("blocks_obj") or {}).values()
+        ):
+            validation_data = {
+                **restored_data,
+                "blocks_obj": block_store.resolve_snapshot_blocks_obj(restored_data),
```
davidwaroquiers: This check is already done in resolve_snapshot_block_obj. Is there any reason not to directly do this:

```suggestion
        validation_data = {
                **restored_data,
                "blocks_obj": block_store.resolve_snapshot_blocks_obj(restored_data),
```

pydatalab/src/pydatalab/routes/v0_1/items.py
```diff
@@ -1404,26 +1484,96 @@ def restore_version(refcode):
...
     user_id = None
     if current_user.is_authenticated:
         user_id = current_user.person.immutable_id
 
+    # Apply each snapshotted blocks_obj entry in its own form: embedded payloads
+    # are written back inline as before, while the references get
+    # the payload from the `block_versions` and becomes the new current `blocks`
+    # state, plus a new RESTORED `block_versions` entry.
+    restored_block_pins: dict[str, dict] = {}
+    if "blocks_obj" in restored_data:
```
davidwaroquiers, line 1551: Not sure the if is needed here as the loop anyway uses "or {}"

pydatalab/src/pydatalab/routes/v0_1/items.py
```diff
@@ -1404,26 +1484,96 @@ def restore_version(refcode):
...
+                    "Dropping unrestorable block reference %s (%s) while restoring %s",
+                    block_id,
+                    entry,
+                    refcode,
+                )
+                if isinstance(restored_data.get("display_order"), list):
+                    restored_data["display_order"] = [
+                        b for b in restored_data["display_order"] if b != block_id
+                    ]
+                continue
```
davidwaroquiers, line 1580: As this should not happen. I'm wondering if it should not be a BadRequest or something. It seems that if it happens it is a bug (from somewhere else) no ? Then if we "just" log it and drop, we may not discover it while using. What do you think ?

davidwaroquiers (commented):
Very good!
I have some small comments on naming of variables + a few comments on some of the logic:
- versions logic
- HasBlocks outside of traits.py
- maybe a few other things

pydatalab/src/pydatalab/models/blocks.py
gpetretto: Indeed moving `DataBlockResponse` should be an option to avoid the circular import. I think this is a matter of preference.

pydatalab/src/pydatalab/models/blocks.py
gpetretto: indeed this is partly to keep exactly the symmetry with what the blocks are doing now. 
Other ways of achieving the same would be possible, but maybe a bit more involved on the side of the code. Or maybe it could also be possible to alter the general behaviour of the backend. I did not verify if this would require any change in the routes.

pydatalab/src/pydatalab/blocks/store.py
gpetretto: I have renamed all the occurrences of `entry` where they are referring to the block content, replaced with `blocks_obj_value`. It would be easy to rename it again if the naming does not sound right. At least it should avoid the ambiguity.

pydatalab/src/pydatalab/blocks/store.py
gpetretto: in general it should not. Since this is used in different places it is a protection against what comes in and does risk giving internal errors. I can remove the check or put a warning. Any preference?

pydatalab/src/pydatalab/blocks/store.py
gpetretto: I renamed it to `authorize_and_get_blocks_data`, so that it does not commit to the explicit content of the blocks_obj or its for in case of future further refactoring.

pydatalab/src/pydatalab/blocks/store.py
gpetretto: Indeed this seems better.

pydatalab/src/pydatalab/blocks/store.py
gpetretto: done

pydatalab/src/pydatalab/blocks/store.py
gpetretto: As described in the docstring, this should be more a case of a concurrent action. I am not sure if the transaction will help here

pydatalab/src/pydatalab/blocks/store.py
gpetretto: Because `timestamp` in `BlockVersion` is a `datetime`, like in `ItemVersion`, while `_now_isoformat` is used to fill the `last_modified` attribute of `Block` which comes from `Entry` and is of type `IsoformatDateTime`.

pydatalab/src/pydatalab/blocks/store.py
gpetretto: As discussed, I will try to implement this is another branch, so we have a comparison.

pydatalab/src/pydatalab/routes/v0_1/blocks.py
gpetretto: ok

pydatalab/src/pydatalab/routes/v0_1/blocks.py
gpetretto: This would keep the equivalence with the original procedure. That one did `find_one` and the query was addressing the `display_order`. So in the case where the block_id was missing, but still in the display_order the document would have resulted as modified, returning a 200. Without this condition, instead the new version would return a 400. Here I switched to `find_one_and_update` to save one action on the DB, but then still tried to reproduce exactly the same logic. It can be questioned if the logic is correct (and properly preserved also...)

pydatalab/src/pydatalab/versioning.py
gpetretto: done. see above.

pydatalab/src/pydatalab/routes/v0_1/items.py
gpetretto: right, done.

pydatalab/src/pydatalab/routes/v0_1/items.py
gpetretto: Removing it would have the impact on the second loop, that instead works on `current_item`. Again, this is a case that in general should not happen, but it is a cheap check. Can be considered for removal.

pydatalab/src/pydatalab/routes/v0_1/items.py
gpetretto: Simply raising here in the middle of the loop would risk leaving the db in an inconsistent state. One would probably need to make a full check before and raise if one block is missing, and if not proceed with the update. Do you think it would be better?

gpetretto:
Sorry @ml-evs, not sure if it is possible with the `bump-pydantic-final-final` gone, but would it be possible to reopen this PR and point it towards main? 

If it is impossible, I have already rebased locally and I can open a new one. It's just that in that case there are at least a few comments that I would need to port to the new PR as well.

ml-evs:
Hi @gpetretto, sorry about this, been trying to find all the PRs that got autoclosed and reopen them. I guess this is a fault with the new PR stack approach... I'll go through and sort these out as best I can ready for next week

ml-evs:
Commenting to reopen...

ml-evs:
Do you now have the power to change the base branch to main @gpetretto? I undeleted the pydantic branch and tried to rebase manually, it would be best if we could keep this PR with all the comments IMO

ml-evs:
commenting again to reopen

gpetretto:
Hi @ml-evs, thanks for the attempts. 
I cannot change the target branch here on github. From a few tests it seems I cannot do here (or in other repositories) if the PR is closed (I can change the target branch for an open PR). But from what you say github does not allow to reopen it because of its target branch. So it seems a deadlock. I can try to directly force push my locally rebased version to my branch, but don't think this would actually change anything for GH.

If nothing works I can open a new PR and copy/paste in the new one the comments that were not addressed here.

Issue #1764, opened 2026-05-22
Tagging system


davidwaroquiers:
### Problem

Organizing and discovering items in large datalab instances can become difficult when relying only on:

- flat table views,
- metadata fields,
- and collections.

A tagging system would provide a lightweight and flexible way to classify, group, and filter items across projects, techniques, users, and workflows.

This is somewhat related to:

- the tree-based organization discussion,
- and the existing Collections feature,

but addresses a different organizational paradigm based on labels and cross-cutting categories.

### Suggested improvement

Add support for attaching tags to items/datasets/samples.

Examples:

- battery
- operando
- XRD
- high-priority
- publication-ready

Tags could then be used for:

- filtering,
- searching,
- grouping,
- dashboards,
- and automated workflows/views.

### Design considerations

- Free-form vs managed tags: Should users be able to create arbitrary free-form tags or should administrators be able to define and manage an approved tag vocabulary? Hybrid approach with free-form personal tags + centrally managed official tags.
- Flat tags vs structured tags: Should tags be simple labels (e.g. "operando", "XRD") or key-value pairs (e.g. "technique=XRD", "status=validated")
- UX/UI: autocomplete, suggestions, tag aliases/synonyms ? prevent duplicate variants (e.g. XRD and xrd), color-coded ?

### Integration with existing features

- The goal of tags should be clearly distinguished/explained from Collections, directories, ...
- Can tags be used to generate "virtual folders" (see tree-based issue) automatically ?

ml-evs:
Going to reopen this to track the sub-issues of post-merge features we want

PR #1937, opened 2026-07-21
Tagging system for items

gpetretto:
As anticipated in #1764, this PR introduces tags for annotating items with a dedicated management page and two permission scopes.
The feature is off by default, enabled through the `PYDATALAB_ENABLE_TAGS` variable.

Note that the branch is based on the pydatinc2 one. The idea is that it will be merged in main after the pydantic2 branch is merged in main.

Here are the main features of the tags:

## Scopes

Two scopes:
* **Global**: only managed by admins, all users can use them
* **Personal**: users can manage their personal tags. Only available to the user that created it

Notes:
* A user may apply a tag to an item only if it's global or their own
  personal tag; they may remove any tag from an item they can edit
  (including another user's personal tag).
* Names are unique **within a scope**. Identity is the immutable id.

## Data model
- `Tag` entry in a new `tags` collection.
- The scope is handled through `scope` and `owner` attributes. Explicitly decided not to use `HasOwner` as, after internal discussions, we considered that belonging to multiple creators or groups may open the road to all sort of accessibility issues. This seemed a different was of scoping the ownership. If preferred the same logic could probably be achieved using `HasOwner` and enforcing in the code that only one creator can be set, with no creator meaning global tag. 
- Items has a `tags` field  through the `HasTags` mixin

## Frontend
* Tags field on item pages: searchable multi-select.
* `/tags` management page: create/edit/delete the tags.
* Samples table: Tags column with a tag filter.


ml-evs (commented):
- fix e2e tests by making the test user an admin
- can we easily click on a tag and have it apply the filter?
- can we set up the global tags pre-defined at the server level?

Maybe future PR? 

- group admin (and future project) admin level tags (future PR)

gpetretto:
> * fix e2e tests by making the test user an admin
> 
>     * can we easily click on a tag and have it apply the filter?
> 
>     * can we set up the global tags pre-defined at the server level?
> 
> 
> Maybe future PR?
> 
>     * group admin (and future project) admin level tags (future PR)

* The e2e testing has been fixed by adding a task to create an initial admin user for the tests. Seems easy enough and to run the e2e tests locally the developer just needs to run the same task manually once.
* I have added the filtering on the click of the tags in the table. I am not sure if the graphical visualization when the tag is highlighted with the mouse hover is the best, but I have tried a few different options and it is probably a matter of taste. Changing should be a couple of lines of CSS anyway.
* After some further discussions, it seemed that the option to pre load tags from a file is not clear how to best deal with it. If this needs to be useful for a real system, it is not clear how this would behave under a redeployment: if a set of tags is pre-defined, and an admin renames one of the tags before the system is redeployed, how to ensure that the original tag is not recreated? This would probably require adding to the tags collection a custom "identifier" other than the name and the immutable_id, that can be used in the pre-fill. Not sure if it is worth. On the other hand, for the tests it means that the set of tags is predefined in the beginning and if the tests act to modify those tags the outcome of the tests may depend on the order in which they are executed.
* renamed the "personal" tags to "user-defined"
* Not implemented the group scope for the tags as it may go in subsequent PR.

pydatalab/src/pydatalab/models/tags.py
```diff
@@ -13,7 +13,7 @@ class Tag(Entry):
 
     - `AccessScope.GLOBAL`: available to (and usable by) everyone; created and
       managed by administrators only. Global tags have no `owner`.
-    - `AccessScope.USER`: a personal tag owned by exactly one user; only that
+    - `AccessScope.USER`: a user-defined tag owned by exactly one user; only that
       user can list, use, edit and delete it.
```
davidwaroquiers, line 17: Maybe good to clarify that any tag that is applied to an item or any other object that would have HasTags will be visible to any user who has access to that item. Not sure if this should be here, in HasTags, somewhere in the docs, or in multiple of these.

davidwaroquiers (commented):
I quickly reviewed the changes. Looks all good to me. Maybe the only thing could be to explicitly mention somewhere (maybe both in the code docstrings, e.g. in HasTags, and in the documentation, there is no doc about the tags for now, probably good to add when this has been reviewed by @datalab-org/datalab-dev )

pydatalab/tasks.py
```diff
@@ -1,3 +1,4 @@
+# This file was edited with the assistance of an AI model and requires human review from the contributor.
```
davidwaroquiers, line 1: Forgot to remove after review ;-)

davidwaroquiers (approved):
All good to me

pydatalab/src/pydatalab/models/tags.py
gpetretto: I agree that it would be better to specify this in the docstrings. The natural place may be `AccessScope` description, but it is not obvious that this would be a general behaviour. Maybe `HasTags`?

gpetretto:
@ml-evs we think this might be ready for your review.

pydatalab/src/pydatalab/routes/v0_1/tags.py
```diff
@@ -0,0 +1,336 @@
...
+    data = request_json.get("data", {})
+
+    name = data.get("name")
+    if not name:
+        return jsonify(status="error", message="A tag name is required."), 400
+
+    try:
+        scope = AccessScope(data.get("scope") or AccessScope.USER.value)
+    except ValueError:
+        return (
```
OMWalmsley: Lots of the routes have custom `jsonify` errors, these could be converted into werkzeug exceptions (https://werkzeug.palletsprojects.com/en/stable/exceptions/) as has been done in various other places in the code base.

pydatalab/src/pydatalab/routes/v0_1/tags.py
```diff
@@ -0,0 +1,336 @@
...
+            jsonify(status="error", message=f"Invalid tag scope {data.get('scope')!r}."),
+            400,
+        )
+
+    if scope == AccessScope.GLOBAL:
+        if not _is_admin():
+            return (
```
OMWalmsley: Similar to previous comment (Werkzeug errors)

pydatalab/src/pydatalab/routes/v0_1/tags.py
```diff
@@ -0,0 +1,336 @@
...
+                jsonify(status="error", message="Only administrators can create global tags."),
+                403,
+            )
+        owner = None
+    else:
+        owner = _current_user_id()
+        if owner is None:
+            return (
+                jsonify(
+                    status="error",
```
OMWalmsley: And here and various other places I haven't listed

pydatalab/src/pydatalab/routes/v0_1/tags.py
```diff
@@ -0,0 +1,336 @@
...
+                    message="You must be logged in to create a user-defined tag.",
+                ),
+                401,
+            )
+
+    if _name_conflict_exists(name, scope, owner):
+        return (
+            jsonify(status="error", message=f"A tag named {name!r} already exists."),
+            409,  # 409: Conflict
```
OMWalmsley: Here (Werkzeug errors). For example: `werkzeug.exceptions.Conflict`

pydatalab/src/pydatalab/mongo.py
```diff
@@ -75,6 +78,56 @@ def files_lookup() -> dict:
...
+                    {
+                        "type": "tags",
+                        "immutable_id": str(match["_id"]),
+                        "name": match.get("name"),
+                        "description": match.get("description"),
+                        "color": match.get("color"),
+                        # `scope` lets the UI mark user-defined tags distinctly.
+                        "scope": match.get("scope"),
+                    }
+                )
```
OMWalmsley: Possibly remove `continue`
```suggestion
                if match is not None:
                    resolved_tags.append(
                        {
                            "type": "tags",
                            "immutable_id": str(match["_id"]),
                            "name": match.get("name"),
                            "description": match.get("description"),
                            "color": match.get("color"),
                            # `scope` lets the UI mark user-defined tags distinctly.
                            "scope": match.get("scope"),
                        }
                    )
```

pydatalab/src/pydatalab/models/tags.py
davidwaroquiers: Let's indeed put it in HasTags, and then it can be added later in the user doc somewhere at some point.

OMWalmsley (commented):
This feature seems to be working on my machine. My comments are mostly about replacing the custom `jsonify` errors with `werkzeug.exceptions`, similar to what has been done previously. This should hopefully standardise things a bit more and reduce the line count slightly.

pydatalab/src/pydatalab/routes/v0_1/tags.py
davidwaroquiers: Thanks @OMWalmsley for this comment. We were wondering what should be used within datalab. We actually looked at how it's done and basically custom jsonify errors (~110) are more common in datalab than werkzeug errors (~50) so we went for the jsonify errors initially but I agree that it would be good to standardize all of this. We can of course convert all these custom jsonify errors to their corresponding werkzeug exceptions. More generally, maybe it would be good to open an issue to standardize all of these in the other routes as well ? @ml-evs may have some comment on this too ?

davidwaroquiers: Following today's meeting with @ml-evs and @OMWalmsley, the plan is to use werkzeug exceptions when it corresponds to a clear api route exception, but raise a proper and adequate error/exception when it is a runtime/datalab related error.

gpetretto: I have replaced most of the jsonify exceptions in `tags.py` with werkzeug.

PR #1942, opened 2026-07-22
Added ability to have multiple named API keys per user

### LINKED ISSUE #1803: Support for multiple API keys per user for e.g., service accounts
Initially these will work as separate keys that can be labelled and individually revoked, but could eventually evolve to also have different permissions per key. Especially useful for beholder.

Initially, user will have option to create multiple API keys, with text label, which can be individually revoked. To enable this, we should:

- store digest of API key (i.e., first 4 and last 4 characters and show them to the user)
- potentially use a prefix for all hashed keys to show this somehow a new format e.g., "datalab-v1-"
- UI can be extended to control this via account settings modal -- will to extend API route to refer to an individual key and handle revoking them.

OMWalmsley:
image

This PR resolves #1803
## Features
- Allows users to create multiple API keys with names.
- Allows users to delete the created API keys.
- Allows users to see the name of the API key and the first four and last four letters of the API key (referred to as the digest).
- Allow the user to copy the API key of the last created key. This happens only after creation and once the page is refreshed this option disappears (replicating the behaviour from before).
### Legacy key support

image

Legacy keys are still shown in the menu under the name legacy key and with the name "Legacy key" and the digest "Unknown". These can then be deleted as normal. Legacy API keys can also be used as normal.

## DB schema changes
db.api_keys:
- `name`
- `user`
- `digest`
- `hash`
In short the user_id now links the user with the api_key instead of the previous _id. 
-  conftest.py: `insert_user` had to change to create a valid api key.  
- login.py: `get_by_api_key` had to be changed to use the `"user_id"`.

Combined into new pydantic models alongside access tokens.

## Creates new routes:
- POST `/api-keys`
- GET `/api-keys`
- DELETE `/api-keys/<id>`
## Frontend additions
- Changed parts of EditAccountSettingsModel.vue to add the functionality


pydatalab/src/pydatalab/routes/v0_1/auth.py
```diff
@@ -1063,17 +1063,69 @@ def get_authenticated_user_info():
...
+@AUTH.route("/api-keys/<id>", methods=["DELETE"])
+def delete_api_key(id):
+    """Deletes the api key associated with the given id. (After checking the user own the key)"""
+    if current_user.is_authenticated:
+        db_request = flask_mongo.db.api_keys.find_one({"_id": ObjectId(id)}, {"user_id": 1})
+        if db_request["user_id"] != current_user.id:
+            return (
+                jsonify({"status": "failure", "message": "User does not own this api key."}),
+                401,
+            )
```
ml-evs: This will fail if the key doesn't exist in the db (as `db_request` will be None). Better to add the permissions into the filter directly and check if the document exists -- you can then make the error message vague to avoid hinting whether the key exists, e.g., something like

```suggestion
        doc = flask_mongo.db.api_keys.find_one({"_id": ObjectId(id), "user_id": current_user.id}, {"user_id": 1})
        if not doc:
            return (
                jsonify({"status": "failure", "message": "API key not found."}),
                404,
            )
```

pydatalab/src/pydatalab/routes/v0_1/auth.py
```diff
@@ -1063,17 +1063,69 @@ def get_authenticated_user_info():
...
+        result = flask_mongo.db.api_keys.delete_one({"_id": ObjectId(id)})
+        if result.deleted_count == 1:
+            return jsonify(), 204
+        else:
+            return jsonify({"status": "failure", "message": "Problem deleting the key"}), 400
```
ml-evs: Here and in other places you can import the Werkzeug classes for Success, NotFound (just grep the code) etc. that avoid needing to make a custom JSON response and error status code, can be useful

ml-evs (commented):
Hi @OMWalmsley, thanks for this! Some quick comments below, and a more general one:

We need to make sure a key generated in the "legacy" style, i.e., user_id matches the key's `_id` in the `api_keys` collection also works correctly. We can either do this via the prefix mentioned in the issue, i.e., assume unprefixed keys are "legacy" and treat others in this new way, or we can simply check if the key ID matches a user and rely on the randomness of the keys to save us.

OMWalmsley:
> Hi @OMWalmsley, thanks for this! Some quick comments below, and a more general one:
> 
> We need to make sure a key generated in the "legacy" style, i.e., user_id matches the key's `_id` in the `api_keys` collection also works correctly. We can either do this via the prefix mentioned in the issue, i.e., assume unprefixed keys are "legacy" and treat others in this new way, or we can simply check if the key ID matches a user and rely on the randomness of the keys to save us.

Thanks @ml-evs for the comments,

I will start working on the fixing the problems you have identified.

After a bit of thought we probably need a way to delete legacy keys from the UI. Otherwise after migrating a datalab instance user will have old API keys they can't delete (due to it not being viewable).

I'll probably aim to implement this by searching for API keys with the `user_id` and then listing them on the frontend with the name "legacy key" and the digest "unknown".

For backwards compatibility in terms of allowing legacy API key access to the system. I would propose comparing the key ID to to the users but also checking that the mongodb document that is return does not contain the digest or user_id field. Using the `$exists` option (e.g. `{digest: {$exists: False}, user_id:{$exists:False}}`).

pydatalab/src/pydatalab/routes/v0_1/auth.py
OMWalmsley: Thanks for pointing this out. 
I have been unable to find a Success class, I have instead went with `Response("", status=204)`. The others return statements now use the Werkzeug classes.

OMWalmsley:
Probably the last task will be testing the routes, so the patch coverage is not as poor. Possibly also some frontend tests.

davidwaroquiers:
Interesting! Maybe one thing that could be interesting (not sure if this could or should be in this PR) would be to have an optional (or default ?) expiration for api keys, e.g. they expire after X months.

pydatalab/tests/server/conftest.py
```diff
@@ -251,7 +251,7 @@ def insert_user(
     real_mongo_client.get_database(TEST_DATABASE_NAME).users.insert_one(demo_user)
     hash = sha512(api_key.encode("utf-8")).hexdigest()
     real_mongo_client.get_database(TEST_DATABASE_NAME).api_keys.insert_one(
-        {"_id": id, "hash": hash}
+        {"_id": ObjectId(), "user_id": str(id), "name": "testing API key", "hash": hash}
```
ml-evs:
```suggestion
        {"_id": ObjectId(), "user_id": id, "name": "testing API key", "hash": hash}
```

I think we can use the bare ID here

pydatalab/tests/server/test_auth.py
```diff
@@ -90,6 +92,200 @@ def test_magic_link_auth_can_be_disabled(unauthenticated_client, app, database,
...
+    json = {"name": "Test_API_KEY"}
+    request_result = client.post("/api-keys", json=json)
+
+    assert request_result.status_code == 200
+
+    result_from_database = api_keys_db.find_one({"name": "Test_API_KEY"})
+
+    assert result_from_database is not None
+    assert result_from_database["digest"][4:-4] == "..."
+    # delete_test_gen_api_keys(database, ["Test_API_KEY"])
```
ml-evs:
```suggestion
```

ml-evs (commented):
Looking good and well tested @OMWalmsley, thanks! I might still play around with the style slightly but all the functionality is there.

Couple of thoughts before we merge this:

- Do you think we should prevent people from making/revoking API keys if they are logged in via an API key? Feels like a nice extra layer of security by depth and would prevent someone from revoking all keys when if one were otherwise leak
- We could also add email notifications when a new key is created, if the user has a registered/verified contact email

pydatalab/tests/server/conftest.py
OMWalmsley: Currently making this change does make 3 of the tests fail. I will have a look into why this happens

OMWalmsley: `current_user.id` is down as a string hence it won't be able to compare the conftest `id` which in this change was an numerical value to the string of `current_user.id`. So some type conversion is required somewhere either:
- In the main functions converting them into integers
- In the mongoDB request
- In conftest (as is currently done)

ml-evs: Ok, lets leave it as-is for now then.

ml-evs: Testing this a bit locally, I think this is actually an indication of another issue, so let's revisit. When creating a new-style API key the `user_id` is a string rather than the `ObjectId` we get for item-specific access tokens. Likely we should also unify the model between these objects -- can discuss properly in person!

pydatalab/src/pydatalab/routes/v0_1/auth.py
```diff
@@ -1063,27 +1063,79 @@ def get_authenticated_user_info():
         return jsonify({"status": "failure", "message": "User must be authenticated."}), 401
 
 
-@AUTH.route("/get-api-key/", methods=["GET"])
+@AUTH.route("/api-keys", methods=["POST"])
 def generate_user_api_key():
     """Returns metadata associated with the currently authenticated user."""
     if current_user.is_authenticated:
+        request_json = request.get_json()
```
ml-evs:
```suggestion
        request_json = request.get_json()
        
        if not request_json.get("name"):
            raise RuntimeError("API key must have a name")
```

pydatalab/src/pydatalab/routes/v0_1/auth.py
```diff
@@ -1063,27 +1063,79 @@ def get_authenticated_user_info():
...
+        all_api_keys = flask_mongo.db.api_keys.find(
+            {"user_id": ObjectId(current_user.id)}, {"hash": 0, "user_id": 0}
         )
+        # find potential legacy keys
+        legacy_key = flask_mongo.db.api_keys.find_one(
+            {
+                "_id": ObjectId(current_user.id),
+                "name": {"$exists": False},
+                "digest": {"$exists": False},
+                "user_id": {"$exists": False},
```
ml-evs:
```suggestion
                "user_id": {"$exists": False},
                "type": {"$exists": False},
```

to avoid clashing with the access token type

ml-evs (commented):
Thanks @OMWalmsley! I think the final thing (as well as minor comments below) is to make sure that all of this machinery works when `access_token`s are also present in the db. We don't necessarily need to use the same schema for each, but might be nice!

pydatalab/src/pydatalab/routes/v0_1/auth.py
```diff
@@ -1063,27 +1063,79 @@ def get_authenticated_user_info():
...
-        flask_mongo.db.api_keys.update_one(
-            {"_id": ObjectId(current_user.id)},
-            {"$set": {"hash": sha512(new_key.encode("utf-8")).hexdigest()}},
-            upsert=True,
+        flask_mongo.db.api_keys.insert_one(
+            {
+                "name": request_json["name"],
+                "user_id": ObjectId(current_user.id),
+                "digest": new_key[:4] + "..." + new_key[-4:],
+                "hash": sha512(new_key.encode("utf-8")).hexdigest(),
```
ml-evs: Consider adding a version here too, e.g., "1" for this new type of API key

OMWalmsley:
New feature. Ability to see creation date. I have also started schemas for the API keys and token access keys.
image
Some styling still needs clearing up.

pydatalab/tests/server/conftest.py
OMWalmsley: This has now been changed so  that the `user_id` or `user` as it's called is just an `ObjectId`

pydatalab/src/pydatalab/routes/v0_1/auth.py
OMWalmsley: This feature has been added alongside a new schema

OMWalmsley:
> Looking good and well tested @OMWalmsley, thanks! I might still play around with the style slightly but all the functionality is there.
> 
> Couple of thoughts before we merge this:
> 
>     * Do you think we should prevent people from making/revoking API keys if they are logged in via an API key? Feels like a nice extra layer of security by depth and would prevent someone from revoking all keys when if one were otherwise leak
> 
>     * We could also add email notifications when a new key is created, if the user has a registered/verified contact email

The ability for API key users to view, create or delete API keys has been removed. The tests have been updated and three more tests have been added to verify the behaviour when using an API key.

pydatalab/src/pydatalab/models/api_keys.py
```diff
@@ -0,0 +1,47 @@
+import datetime
```
ml-evs: This is minor, but I think this should actually live in a different module, e.g., `pydatalab.permissions` or somewhere that doesn't break imports -- the pydantic models in `pydatalab.models` are loosely intended to be the scientific rather than logistical ones, and putting it here will mean our docs generate pages for them. As this isn't part of the "public API" we probably want to avoid this.

ml-evs (commented):
Quick comment, can discuss the rest shortly!

pydatalab/src/pydatalab/permissions.py
```diff
@@ -113,6 +114,17 @@ def wrapped_route(*args, **kwargs):
     return wrapped_route
 
 
+def exclude_api_key(func):
```
ml-evs, line 178: The decorator approach is nice (though needs a bit of docstring explanation) but I'm not sure I like this being a user attribute -- instead it should be handled per request with a flask global context I think -- not much needed to change but means `get_by_id` changes can be reverted

ml-evs:
```
from flask import g

if g["api_key_login"]:
    raise Forbidden
```

pydatalab/src/pydatalab/login.py
ml-evs, line 175: I think this is where you would set the [flask global context](https://flask.palletsprojects.com/en/stable/appcontext/) or in the [request context](https://flask.palletsprojects.com/en/stable/reqcontext/)


```suggestion
    api_key = request.headers.get("DATALAB-API-KEY", None)
    
    
	from flask import g
	if api_key:
	    g["api_key_session"] = True
```

pydatalab/src/pydatalab/routes/v0_1/auth.py
```diff
@@ -1064,27 +1066,91 @@ def get_authenticated_user_info():
...
+@AUTH.route("/api-keys", methods=["POST"])
+@exclude_api_key
 def generate_user_api_key():
     """Returns metadata associated with the currently authenticated user."""
     if current_user.is_authenticated:
+        request_json = request.get_json()
+
+        if not request_json.get("name"):
+            raise RuntimeError("API key must have a name")
         new_key = "".join(random.choices(ascii_letters, k=KEY_LENGTH))  # noqa: S311
```
ml-evs: Probably we should replace this with `new_key = secrets.token_urlsafe(16)` or similar

pydatalab/src/pydatalab/models/api_keys.py
OMWalmsley: They have been moved to `pydatalab.permissions`

pydatalab/src/pydatalab/routes/v0_1/auth.py
OMWalmsley: Replaced with `new_key = secrets.token_urlsafe(KEY_LENGTH)`

pydatalab/src/pydatalab/login.py
OMWalmsley: This change has been performed, using the flask global context

pydatalab/src/pydatalab/permissions.py
OMWalmsley: Docstring has been added, and it has been removed from the user attribute.

pydatalab/src/pydatalab/routes/v0_1/auth.py
```diff
@@ -1063,27 +1064,86 @@ def get_authenticated_user_info():
...
-        new_key = "".join(random.choices(ascii_letters, k=KEY_LENGTH))  # noqa: S311
-        flask_mongo.db.api_keys.update_one(
-            {"_id": ObjectId(current_user.id)},
-            {"$set": {"hash": sha512(new_key.encode("utf-8")).hexdigest()}},
-            upsert=True,
+
+    request_json = request.get_json()
+
+    if not request_json.get("name"):
+        raise RuntimeError("API key must have a name")
```
ml-evs:
```suggestion
        raise BadRequest("API key must have a name")
```

pydatalab/src/pydatalab/routes/v0_1/auth.py
```diff
@@ -1063,27 +1064,86 @@ def get_authenticated_user_info():
...
+        # Deal with potential legacy key
+        doc = flask_mongo.db.api_keys.find_one(
+            {
+                "_id": ObjectId(api_id),
+                "user": {"$exists": False},
+                "digest": {"$exists": False},
+                "name": {"$exists": False},
+                "type": {"$exists": False},
+            },
         )
```
ml-evs, line 1139: Worth a check here; otherwise any user can remove another users key if they know their public user ID

```suggestion
        # Deal with potential legacy key
        if ObjectId(api_id) == ObjectId(current_user.id):
	        doc = flask_mongo.db.api_keys.find_one(
	            {
	                "_id": ObjectId(api_id),
	                "user": {"$exists": False},
	                "digest": {"$exists": False},
	                "name": {"$exists": False},
	                "type": {"$exists": False},
	            },
			)
```

pydatalab/src/pydatalab/routes/v0_1/auth.py
```diff
@@ -1063,27 +1064,86 @@ def get_authenticated_user_info():
         return jsonify({"status": "failure", "message": "User must be authenticated."}), 401
 
 
-@AUTH.route("/get-api-key/", methods=["GET"])
+@AUTH.route("/api-keys", methods=["POST"])
+@authenticate("User must be an authenticated admin to request an API key.")
```
ml-evs:
```suggestion
@authenticate("User must be an authenticated user to request an API key.")
```

pydatalab/src/pydatalab/routes/v0_1/auth.py
```diff
@@ -1063,27 +1064,86 @@ def get_authenticated_user_info():
...
+    if legacy_key:
+        legacy_key["digest"] = "Unknown"
+        legacy_key["name"] = "Legacy key"
+        all_keys.insert(0, legacy_key)
+
+    return jsonify({"api_keys": all_keys}), 200
+
+
+@AUTH.route("/api-keys/<api_id>", methods=["DELETE"])
+@authenticate("User must be an authenticated admin to request an API key.")
```
ml-evs:
```suggestion
@authenticate("User must be an authenticated to revoke an API key.")
```

pydatalab/src/pydatalab/permissions.py
```diff
@@ -113,6 +158,37 @@ def wrapped_route(*args, **kwargs):
...
+
+    return function_wrap
+
+
+def exclude_api_key(func):
+    """Decorator to ensure that ensures client using api keys cannot access a route"""
+
+    @wraps(func)
+    def wrapped_route(*args, **kwargs):
+        # current_user must be accessed to regenerate the api_key_session
```
ml-evs: Dont think this comment is relevant anymore

```suggestion
```

ml-evs (commented):
Couple more suggestions before merging, only one is non-cosmetic

ml-evs (approved):
Looks great now, thanks @OMWalmsley!

PR #1958, opened 2026-08-03
Fixed position of the "actions" menu when items are selected

### LINKED ISSUE #2012: Datatable layout shift when an item is selected
On the main Item datatables, the "1 selected..." button appears when an item is selected, causing the table to shift on the page. We should move the "1 selected..." button to avoid this.

davidwaroquiers:
When selecting an item in a table, the lines would be moved downwards for the actions menu to take its place (which was a bit annoying and not super user friendly).
This PR is a small UX improvement that moves the "actions" menu up in the "table menu" bar.
What it looks like when items are selected:

image

I also added the number of items when there is no selection:

image

Following up @ml-evs comments, I also modified responsiveness for smaller screens:

**Below breaking point xs:**
image

image

**Between xs and sm:**
image

image

**Between sm and lg**
image

image

**Above breaking point lg**
image

image

This should be ready for review.


ml-evs (commented):
Thanks @davidwaroquiers, this looks nice, could you show what it looks like in a narrow window too please?

davidwaroquiers:
> Thanks @davidwaroquiers, this looks nice, could you show what it looks like in a narrow window too please?

That's a good point, what is the intended narrowest use case ? Should it go down to smartphone ?

One thing to mention is that currently, the three buttons for adding an item, addning a batch of items and scanning a qr code just go out of the screen when it's too narrow. Should be easy to fix and I can do that in this PR but I'd like to know up to which point :)

davidwaroquiers:
> Thanks @davidwaroquiers, this looks nice, could you show what it looks like in a narrow window too please?

I tried to make something reasonable there down to the smallest screen possible (bootstrap's xs size). What do you think ?

webapp/src/components/DynamicDataTableButtons.vue
```diff
@@ -18,257 +18,305 @@
...
+          v-if="dataType === 'equipment'"
+          data-testid="add-equipment-button"
           class="btn btn-default btn-action"
-          aria-label="Scan QR code"
-          title="Scan QR code"
-          @click="$emit('open-qr-scanner-modal')"
+          @click="$emit('open-create-equipment-modal')"
         >
-          <font-awesome-icon icon="qrcode" /> Scan QR code
+          Add an equipment
```
DianaAliabieva:
```suggestion
          Add equipment
```

webapp/cypress/e2e/equipment.cy.js
```diff
@@ -41,7 +41,7 @@ describe("Equipment table page", () => {
   it("Loads the equipment page without any errors", () => {
     cy.findByText("About").should("exist");
     cy.findByText("Equipment").should("exist");
-    cy.findByText("Add an item").should("exist");
+    cy.get('[data-testid="add-equipment-button"]').should("contain.text", "Add an equipment");
```
DianaAliabieva:
```suggestion
    cy.get('[data-testid="add-equipment-button"]').should("contain.text", "Add equipment");
```

webapp/src/components/DynamicDataTableButtons.vue
```diff
@@ -18,257 +18,305 @@
...
+              >
+                Batch share
+              </a>
+              <a
+                v-if="!['collectionItems', 'users', 'tokens', 'groups'].includes(dataType)"
+                data-testid="delete-selected-button"
+                class="dropdown-item"
+                @click="confirmDeletion"
+              >
+                Delete selected
```
DianaAliabieva:
```suggestion
                Delete items
```

DianaAliabieva (changes requested):

ml-evs (commented):
Functionality-wise this look very nice, but this has become quite a big PR now, in particular in webapp/src/components/DynamicDataTableButtons.vue! @davidwaroquiers and @DianaAliabieva could you assess how cleanly this would work with #1914 and decide which should be merged first?

davidwaroquiers:
I just rebased this on main and reviewed it again. Should be ready to be reviewed and merged once approved.

PR #1962, opened 2026-08-04
Fix ephemeral race condition in e2e test of batch sample table

davidwaroquiers:
In the cypress test for the sample table page, there might be some race condition between the save of the testB and the copy of testB to testB_COPY.
Not sure if this is the only reason (and if it is the right one...) why we sometimes get random failures.

I've added a wait for the save to actually happen before making a copy.


ml-evs (commented):
Thanks @davidwaroquiers, looks useful! I'm going to manually rerun the tests a few times to see if the old behaviour triggers in the CI, and if not then will merge this!

ml-evs (approved):
Seems to work well, thanks!

PR #1975, opened 2026-08-06
adding a button to clean all filters

DianaAliabieva:
Adding a button to clean all filters when multiple ones are selected. 
image
image
image


davidwaroquiers:
Note to @ml-evs, we'd see the contextual actions for each row in the same "column" as the column in which this clear filter is right now.

ml-evs (commented):
Just mentioning again, could this be based on #1914 please? Otherwise looks good!

DianaAliabieva:
> Just mentioning again, could this be based on #1914 please? Otherwise looks good!

Hello, the PR is ready for the check and now it is based on #1914

PR #1995, opened 2026-08-10
Selected deps and GH actions updates

ml-evs:
This PR cherry-picks several dependabot PRs with security concerns and GH actions updates.


ml-evs:
Merging this even with a minor test failure, I think it's likely intermittent but if not will patch in a separate PR.

PR #2001, opened 2026-08-11
Improving location support

### LINKED ISSUE #1970: Improving location support
We want to support hierarchical locations for samples and starting materials. #1838 (with issue #1711) added support for hierarchical editing/selection of locations based on all locations found dynamically in the inventory. We should extend this to a) work for samples too (#1950) and b) use a proper endpoint to query all unique locations to prefill this field.

Initially this should be tied to user auth, i.e., give me all locations at which samples this user has access to are based. We may think of then adding location as a proper entry type going forward that would allow annotation and better validation of prebaked locations in the lab, but for now the hierarchy can still be set using the special string format "A > B > C" to indicate sublocations. Likely the `HasLocation` trait prototyped in #1950 can be extended to validate this, and the API route can ignore any values that don't match this.

OMWalmsley:
Resolves #1970

Locations are where things are stored. Currently we are using a single string in the database to store the location and then frontend application logic on top to create a nested hierarchy. Moving forward it would be ideal to move some of this logic to the backend to prevent locations being generated from stale/deprecated data that has been stored in the frontend.
## Features
- Introduces `GET \locations` endpoint.
  - Which returns a flat list of all locations and a nested list of all locations.
- Introduces `HasLocation` attribute for all items
## Example
An example return from the endpoint is here
`{"flat_locations":["Hub1 > Hub2 > Place 1","LHud One","Lab 1 > Shelf 1 > Place A","Lab 4 > Shelf 1","lab1"],"nested_locations":{"Hub1":{"Hub2":{"Place 1":{}}},"LHud One":{},"Lab 1":{"Shelf 1":{"Place A":{}}},"Lab 4":{"Shelf 1":{}},"lab1":{}}}
`
## Testing
Currently three tests have been written to test this endpoint functionality (`GET \locations`).

## Further planned functionality
Implement this into the frontend so that it queries this endpoint instead of using frontend storage.

### DIFF
--- a/pydatalab/schemas/cell.json
+++ b/pydatalab/schemas/cell.json
@@ -1221,6 +1221,19 @@
       "title": "Display Order",
       "type": "array"
     },
+    "location": {
+      "anyOf": [
+        {
+          "type": "string"
+        },
+        {
+          "type": "null"
+        }
+      ],
+      "default": null,
+      "description": "The place where the item is located.",
+      "title": "Location"
+    },
     "collections": {
       "description": "Inlined info for the collections associated with this item.",
       "items": {
--- a/pydatalab/schemas/equipment.json
+++ b/pydatalab/schemas/equipment.json
@@ -1057,6 +1057,19 @@
       "title": "Display Order",
       "type": "array"
     },
+    "location": {
+      "anyOf": [
+        {
+          "type": "string"
+        },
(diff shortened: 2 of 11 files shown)


OMWalmsley:
This PR has been superseeded by #2002

PR #2011, opened 2026-08-13
Improvements to item versioning UI

### LINKED ISSUE #1784: Show creator of "diff" in top-level versioning UI

### LINKED ISSUE #2010: Spurious item versions generated
We're still in a situation where item versions are generated in some cases where they should not be. The root cause is that before declaring a new version, the old version is pulled up from the database and compared with the new suggested one (which is already saved by the endpoint). The new version has gone through various pydantic models in order to normalise it; of course, the original saved version did *at one point* go through these models, so the format is nearly the same, but when saving to the versions collection no pydantic model is used, ~and therefore raw dates get passed in as strings rather than bson date objects.~ this is not actually true, but fundamentally difference in schema between pydantic and the database are messing things up.

So; smallest fix would be to *at least* bson encode the versions at save time, so that `date = "1990-01-01"` becomes `date: {"$date": "..."})` in the versions db too.

ml-evs:
Headline changes: 

- permissions updates now mint a new version with a special action
- timestamps are properly written as timestamps when restoring
- more computed fields are ignored to exclude extraneous version updates
- new tests for this
- updates version history UI to show who made the change, the user agent, and the diff

Closes #2010, closes #1784

### DIFF
--- a/pydatalab/src/pydatalab/routes/v0_1/items.py
+++ b/pydatalab/src/pydatalab/routes/v0_1/items.py
@@ -995,6 +995,21 @@ def _process_item_permissions(
             400,
         )
 
+    # Record the new ownership in the item's history as a version
+    version_response, version_status = save_version_snapshot(
+        refcode, action=VersionAction.PERMISSIONS_UPDATE
+    )
+    if version_status != 200:
+        LOGGER.error(
+            "Failed to save version for item %s after updating permissions: %s",
+            refcode,
+            version_response,
+        )
+    elif "version" in version_response:
+        flask_mongo.db.items.update_one(
+            {"refcode": refcode}, {"$set": {"version": version_response["version"]}}
+        )
+
     return {"status": "success"}, 200
 
 
--- a/pydatalab/src/pydatalab/routes/v0_1/items.py
+++ b/pydatalab/src/pydatalab/routes/v0_1/items.py
@@ -1526,7 +1526,7 @@ def restore_version(refcode):
 
     # Update metadata to reflect the restore action
     restored_data["version"] = next_version_number
-    restored_data["last_modified"] = datetime.datetime.now(tz=datetime.timezone.utc).isoformat()
+    restored_data["last_modified"] = datetime.datetime.now(tz=datetime.timezone.utc)
 
     # Validate restored data against the item model
     item_type = current_item["type"]
--- a/pydatalab/src/pydatalab/versioning.py
+++ b/pydatalab/src/pydatalab/versioning.py
@@ -2,7 +2,6 @@
 
 import datetime
 
-from bson import json_util
 from flask import request
 from flask_login import current_user
 from pydantic import ValidationError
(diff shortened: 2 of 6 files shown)


pydatalab/src/pydatalab/routes/v0_1/items.py
```diff
@@ -995,6 +995,21 @@ def _process_item_permissions(
...
+    if version_status != 200:
+        LOGGER.error(
+            "Failed to save version for item %s after updating permissions: %s",
+            refcode,
+            version_response,
+        )
+    elif "version" in version_response:
+        flask_mongo.db.items.update_one(
+            {"refcode": refcode}, {"$set": {"version": version_response["version"]}}
+        )
```
be-smith, line 1011:
```suggestion
    elif "version" in version_response:
        flask_mongo.db.items.update_one(
           {"refcode": refcode},
            {
                "$set": {
                    "version": version_response["version"],
                    "last_modified": datetime.datetime.now(tz=datetime.timezone.utc),
                }
            },
        )
```

be-smith: Should we bump last modified when updating permissions?

webapp/src/components/VersionHistoryModal.vue
```diff
@@ -201,15 +191,29 @@
...
+const IGNORED_FIELDS = [
+  "last_modified",
+  "version",
+  "_id",
+  "immutable_id",
+  "refcode",
+  "creators",
+  "groups",
+  "creator_ids",
+  "group_ids",
```
be-smith, line 210:
```suggestion

```

be-smith: This PR is adding version control for permission changes - should we not then display this in the diff?

be-smith (changes requested):
I'm still going through the version history modal logic but it works well on my end. Couple of changes suggested based on what behaviour you want?

webapp/src/components/VersionHistoryModal.vue
be-smith: I think this is handled by your early exit from the modal if the version is labelled as a permissions update, so these still won't be shown in the diff as the diff isn't shown

pydatalab/src/pydatalab/routes/v0_1/items.py
ml-evs: no (thanks)

pydatalab/src/pydatalab/routes/v0_1/items.py
```diff
@@ -995,6 +995,21 @@ def _process_item_permissions(
...
+    # Record the new ownership in the item's history as a version
+    version_response, version_status = save_version_snapshot(
+        refcode, action=VersionAction.PERMISSIONS_UPDATE
+    )
+    if version_status != 200:
+        LOGGER.error(
+            "Failed to save version for item %s after updating permissions: %s",
+            refcode,
+            version_response,
+        )
```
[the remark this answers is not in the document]
ml-evs: Acceptable risk -- the version history is not (yet) a proper audit, but would rather fail on versioning than prevent permissions from being updated

pydatalab/src/pydatalab/versioning.py
```diff
@@ -130,18 +155,24 @@ def save_version_snapshot(
...
             return {"status": "success", "message": "No changes detected, version not saved."}, 200
 
-    # Atomically get the next version number
-    next_version_number = get_next_version_number(refcode)
+        LOGGER.debug(
+            "Changes detected for %s, saving new version. Old: %s, New: %s",
+            refcode,
+            json_util.dumps(last_data),
+            json_util.dumps(current_data),
+        )
```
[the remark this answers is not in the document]
ml-evs:
```suggestion
```

PR #2045, opened 2026-08-28
Tagging system for items

### LINKED ISSUE #1764: Tagging system

gpetretto:
Reopening #1937 after the pydantic 2 branch has been deleted. All the comments raised there should have been addressed.
Reporting the text of the original PR.

---

As anticipated in #1764, this PR introduces tags for annotating items with a dedicated management page and two permission scopes.
The feature is off by default, enabled through the `PYDATALAB_ENABLE_TAGS` variable.

Note that the branch is based on the pydatinc2 one. The idea is that it will be merged in main after the pydantic2 branch is merged in main.

Here are the main features of the tags:

## Scopes

Two scopes:
* **Global**: only managed by admins, all users can use them
* **Personal**: users can manage their personal tags. Only available to the user that created it

Notes:
* A user may apply a tag to an item only if it's global or their own
  personal tag; they may remove any tag from an item they can edit
  (including another user's personal tag).
* Names are unique **within a scope**. Identity is the immutable id.

## Data model
- `Tag` entry in a new `tags` collection.
- The scope is handled through `scope` and `owner` attributes. Explicitly decided not to use `HasOwner` as, after internal discussions, we considered that belonging to multiple creators or groups may open the road to all sort of accessibility issues. This seemed a different was of scoping the ownership. If preferred the same logic could probably be achieved using `HasOwner` and enforcing in the code that only one creator can be set, with no creator meaning global tag. 
- Items has a `tags` field  through the `HasTags` mixin

## Frontend
* Tags field on item pages: searchable multi-select.
* `/tags` management page: create/edit/delete the tags.
* Samples table: Tags column with a tag filter.

Closes #1764


.github/workflows/ci.yml
```diff
@@ -323,6 +326,10 @@ jobs:
             exit 1
           fi
 
+      - name: Create the test admin user
+        # The first admin cannot be created over the API, so create one directly.
+        run: docker compose exec -T api uv run invoke admin.seed-e2e-admin
```
ml-evs:
```suggestion
        run: docker compose exec -T api uv run invoke dev.seed-e2e-admin
```

pydatalab/src/pydatalab/models/utils.py
```diff
@@ -162,6 +162,13 @@ class UserRole(str, Enum):
     MANAGER = "manager"
 
 
+class AccessScope(str, Enum):
+    """The scope that controls who can list, use and manage an entity (e.g. a tag)."""
+
+    GLOBAL = "global"
+    USER = "user"
```
ml-evs: These need better documentation if they are to be managed here; not clear if global == public or only logged-in accounts. I'm going to move this into tags.py and rename `TagsAccessScope`.

pydatalab/tasks.py
```diff
@@ -364,6 +364,58 @@ def manually_register_user(
...
+        db.users.update_one(
+            {"_id": user_id}, {"$set": {"account_status": AccountStatus.ACTIVE.value}}
+        )
+        print(f"User <{contact_email}> already exists ({user_id}); ensured active.")
+
+    db.roles.update_one({"_id": user_id}, {"$set": {"role": UserRole.ADMIN.value}}, upsert=True)
+    print(f"Granted the admin role to <{contact_email}>.")
+
+
+admin.add_task(seed_e2e_admin)
```
ml-evs:
```suggestion
dev.add_task(seed_e2e_admin)
```

Probably this is a dev task rather than admin, though it could be extended to be more useful (e.g., add an ORCID by default, rather than email which is not widely used).

pydatalab/src/pydatalab/mongo.py
```diff
@@ -75,6 +78,55 @@ def files_lookup() -> dict:
     }
 
 
+def resolve_tags_for_docs(docs: list[dict]) -> None:
+    """Inline tag details into each doc's `tags` field, in place.
+
+    Tag references (mappings carrying an `immutable_id`) are resolved against
```
ml-evs, line 84: I think this is the bit I'm most concerned about -- open permissions queries and then the fact that a user can see another user tags after they have been shared. I'll open a follow-up issue I think (see #2124)

ml-evs (approved):
Hi @gpetretto, thanks also for this! Even harder to review than the other PR, but empirically this works well. I've left some comments below and resolved the ones I have taken (minor) executive action on. I've also opened a couple of sub-issues off of #1764 to track future work. But otherwise I'm happy to merge this in the "turned off by default for now" state.

webapp/src/components/TagSelect.vue
ml-evs, line 1: It would be nice if tag select is prepopulated with all the tags as user can see, rather than requiring a key press -- given that there are unlikely to be that many tags (see #2125)

PR #2052, opened 2026-08-30
Install and serve Bokeh properly through yarn

jdbocarsly:
From the early days of datalab, we ran into issues properly serving Bokeh through our bundler, so we served it via NPM. This PR attempts to fix that, and also add some tests (thanks, Claude)


ml-evs:
https://github.com/datalab-org/datalab/blob/e1a5e090eaf0738329cd65e91d8ff685566fbcc8/CONTRIBUTING.md?plain=1#L43


;)

jdbocarsly:
~~Rebased this onto branch `jdb/fix-local-component-test-runner` in order to run and fix the component tests. However, it could also be rebased directly on main if needed, since the CI component tests were always running fine.~~

Note: that has been merged now so this is on main again.

ml-evs (approved):
Works nicely, thanks @jdbocarsly!

PR #2056, opened 2026-08-30
Move component test runner port to fix for macOS

jdbocarsly:
I noticed that `yarn test:component` wasn't working on my machine. This may be a MacOS issue, or may be a more general issue. 

The issue turns out to be that Cypress runs a server for these tests that defaults to port 8080, and there is some issue with communicating with it on that port. By forcing it to port 8082, it works on my machine. It would be good to verify that this doesn't break anything for others!

### DIFF
--- a/webapp/cypress.config.ts
+++ b/webapp/cypress.config.ts
@@ -8,6 +8,8 @@ export default defineConfig({
     defaultCommandTimeout: 10000,
   },
   component: {
+    // Avoid port 8080's macOS WebSocket issue and the usual local dev-server port, 8081.
+    port: 8082,
     devServer: {
       framework: "vue-cli",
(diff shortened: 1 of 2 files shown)


webapp/cypress.config.ts
```diff
@@ -8,6 +8,8 @@ export default defineConfig({
     defaultCommandTimeout: 10000,
   },
   component: {
+    // Avoid port 8080's macOS WebSocket issue and the usual local dev-server port, 8081.
```
ml-evs, line 11: @be-smith 🙃

ml-evs (approved):
I will give @be-smith the satisfaction of telling you what the 8080 port issue is -- otherwise LGTM

Issue #1898, opened 2026-07-01
Add simple username/password login for dev/testing


jbouquiaux:
Testing multi-user scenarios locally (access rights, data sharing, groups, ...) requires juggling OAuth accounts (or we have missed something obvious to do it?), which is cumbersome.

Proposal: add a simple username/password login behind a dedicated flag (e.g., PYDATALAB_ENABLE_TEST_USERS). Test users would be created via invoke dev.create-test-user --username alice --password alice. Credentials stored in a local JSON file — no security requirements, dev-only. The UI would show a clear warning when this mode is active.

@davidwaroquiers has a working implementation on our fork ([Matgenix/datalab#5](https://github.com/Matgenix/datalab/pull/5)) that we'd be happy to contribute, for instance a part of the dev profile PR https://github.com/datalab-org/datalab/pull/1845.

davidwaroquiers:
I also considered having just username, no password as this is really intended for dev/testing only. That would simplify even more the added code.

PR #2062, opened 2026-08-31
Passwordless test login

davidwaroquiers:
Added an additional login procedure for dev/tests. This is purposedly simplistic, just a username, and to login, you just click on the user in the list. The goal is to be able to test various access features / groups etc ... without the need to have multiple OAuth logins.

Adapted from the previous PR here: https://github.com/Matgenix/datalab/pull/5
In this previous PR, a password credentials existed. Given the actual purpose of this, i.e. just be able to have different authenticated users easily in a purely test/dev environments, the password credentials added unnecessary code.

Logic in datalab for access to items/anything is:
1/ Authentication (who am I)
2/ Tenant access (do i have access to to this datalab instance)
3/ resource-level access control (can I access this specific "object")

This PR allow to kind of "bypass" 1/ and 2/ (not really bypass, just make it super easy for testing/development).

### DIFF
--- a/webapp/src/components/UnsafeTestingPasswordlessLogin.vue
+++ b/webapp/src/components/UnsafeTestingPasswordlessLogin.vue
@@ -1,0 +1,117 @@
+<template>
+  <Modal v-model="isOpen" :is-large="false">
+    <template #header>
+      <font-awesome-icon icon="exclamation-triangle" class="text-danger" />
+      Unsafe passwordless test login
+    </template>
+    <template #body>
+      <div class="alert alert-danger" data-testid="testing-passwordless-warning">
+        No authentication is performed. Anyone with access to this instance can impersonate any user
+        listed below. Never enable this login mechanism in production.
+      </div>
+      <div v-if="error" class="alert alert-warning" data-testid="testing-passwordless-error">
+        {{ error }}
+      </div>
+      <div v-else-if="users === null" class="text-center text-muted py-3">
+        <font-awesome-icon icon="spinner" spin /> Loading test users…
+      </div>
+      <div
+        v-else-if="users.length === 0"
+        class="text-muted"
+        data-testid="testing-passwordless-empty"
+      >
+        No passwordless test users are configured.
+      </div>
+      <div v-else class="list-group" data-testid="testing-passwordless-users">
+        <button
+          v-for="user in users"
+          :key="user.username"
+          type="button"
+          class="list-group-item list-group-item-action"
+          @click="login(user.username)"
+        >
+          <span class="d-flex justify-content-between align-items-center">
+            <span class="text-left">
+              <strong>{{ user.display_name || user.username }}</strong>
+              <small class="d-block text-muted">{{ user.username }}</small>
+            </span>
+            <span class="d-flex align-items-center">
+              <RoleBadge :role="user.role" class="mr-2" />
+              <span
+                class="d-flex align-items-center"
+                :title="`Account status: ${user.account_status}`"
+                :aria-label="`Account status: ${user.account_status}`"
+              >
+                <UserStatusCell :status="user.account_status" />
+              </span>
+            </span>
+          </span>
+        </button>
+      </div>
+    </template>
+    <template #footer>
+      <button
+        type="button"
+        class="btn btn-secondary"
+        data-testid="testing-passwordless-close"
+        @click="isOpen = false"
+      >
+        Close
+      </button>
+    </template>
+  </Modal>
+  <button
+    type="button"
+    class="dropdown-item btn login btn-link unsafe-testing-login"
+    aria-label="Open unsafe passwordless test login"
+    data-testid="testing-passwordless-open"
+    @click="open"
+  >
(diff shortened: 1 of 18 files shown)


davidwaroquiers:
@jbouquiaux could you test and review (probably more the backend part) this ? Likely we can add something in the dev-docker documentation with the command to add a user with the docker compose (something like "docker compose exec api-dev uv run invoke dev.create-test-user   --username gloubs    --role user")
@DianaAliabieva could you test and review (probably more the frontend part) this ?

webapp/src/components/UnsafeTestingPasswordlessLogin.vue
```diff
@@ -0,0 +1,117 @@
...
+        this.error = "Unable to load passwordless test users.";
+      }
+    },
+    async login(username) {
+      try {
+        await loginTestingPasswordless(username);
+        window.location.reload();
+      } catch {
+        this.error = "Unable to log in as that test user.";
+      }
```
DianaAliabieva, line 106: Bug: user list disappears after a failed login attempt.

error here is used both for "failed to load the list" and "failed to log in as this user". Since the template shows either the error OR the list (v-if/v-else), a failed login hides the already-loaded user list - even though it's still valid data. The person then has to close and reopen the modal just to try a different user.

Suggest splitting this into two separate states (e.g. a load error and a login error) so a failed login attempt shows the error alongside the list instead of replacing it.

Also, the user buttons aren't disabled while login() is in flight, so a fast double-click can fire two concurrent login requests - worth guarding against while you're in here.

davidwaroquiers: Fixed

ml-evs (commented):
Hi @davidwaroquiers, thanks for this. We definitely need something that solves this problem but I think I've come down against the idea of adding password-based login in this way.

I think my suggestion would be instead to hook into the email login approach and enable that by default on dev servers. We can then make `invoke dev.serve` provide an authenticated link with a token for testing the developers "normal" user by default, and an extra invoke task very similar to the one you added here that creates a test-user and prints a token link for the developer.

What do you think? Happy to implement this if you are okay with it, or go back to this idea if you still prefer a password.

davidwaroquiers:
> Hi @davidwaroquiers, thanks for this. We definitely need something that solves this problem but I think I've come down against the idea of adding password-based login in this way.
> 
> I think my suggestion would be instead to hook into the email login approach and enable that by default on dev servers. We can then make `invoke dev.serve` provide an authenticated link with a token for testing the developers "normal" user by default, and an extra invoke task very similar to the one you added here that creates a test-user and prints a token link for the developer.
> 
> What do you think? Happy to implement this if you are okay with it, or go back to this idea if you still prefer a password.

Hi @ml-evs 

Thank you for the message. Just to mention that it is even not with a password it's "easier". Here is with a few screenshots:

1/ Login page:

image

2/ List of clickable passwordless users with some info (their role, the group(s) they belong to):

image

And you just click on the user you want to be logged in. That eases tests with different users with different roles, belonging to different groups, in the future it could also help test group admins, project admins etc ...).

It is quite perpendicular to almost everything in datalab. It only touches four small points:
- The configuration flag exposed in the /info response (ENABLE_UNSAFE_TESTING_PASSWORDLESS_LOGIN -- could be renamed as it is very long but at least it's clear :D)
- The invoke dev.create-test-user task (only active when the above flag is True)
- Two endpoints on the existing AUTH blueprint group: "GET /login/testing-passwordless/users" to get the list of test users, and "POST /login/testing-passwordless" to start the normal Flask login session for the selected (clicked) user
- The user picker with the buttons shown above

After a user is selected, everything follows the normal datalab "flow path" (using the Flask login session for access etc ...).

What do you think ?

ml-evs:
Ah sorry, I think I just searched for "password" and thought I was responding to #1898 :upside_down_face: I guess my question then becomes whether adding this to the UI itself is that useful, and if it really needs an API route? I'd prefer a task that lists these login users in the terminal and offers login links for each. Do you find the extra UI component that much more useful than having e.g., a few private browser windows logged in as different accounts?

davidwaroquiers:
> Ah sorry, I think I just searched for "password" and thought I was responding to #1898 🙃 I guess my question then becomes whether adding this to the UI itself is that useful, and if it really needs an API route? I'd prefer a task that lists these login users in the terminal and offers login links for each. Do you find the extra UI component that much more useful than having e.g., a few private browser windows logged in as different accounts?

Not that much more useful indeed. I think the idea of having one invoke task that gives the list of passwordless users and their "properties" (roles, groups, ...) and the link is all good. I'll remove the UI part. As for the API route, we at least need the one to login no ?

davidwaroquiers:
Done, should be ready for review

ml-evs:
Could you take a look at https://github.com/datalab-org/datalab/commit/95aa6b09868a3e4be1332908caa495dbbbe45e63 @davidwaroquiers? I think this is the same functionality but without needing any extra config or routes, just lets you trigger token creation using your test accounts and prints the links you would use to login

PR #2063, opened 2026-08-31
Fix flaky SampleTable component test

jdbocarsly:
Apply a few changes to hopefully fix the flakiness of the SampleTable component test in CI on electron. These generally passed locally, but often failed on the CI. The issue appears to be related to 2 issues: 
- The datalab primevue theme we use wasn't applied in the component test, hiding a button at some screen sizes
- the slower CI runner meant that some interactions hadn't loaded by the time the test runner was looking for them, specifically in the column filters.


ml-evs (approved):
Thanks, let's give it a go! I'm still seeing flakiness on the e2e sample test but this one does seem robust now

PR #2069, opened 2026-09-01
Separate blocks from items

### LINKED ISSUE #48: Move blocks to their own collection instead of being embedded documents within items

gpetretto:
Reopening #1936 after the pydantic 2 branch has been deleted. Note that, aside from the addressed comments, this also introduces a separate counter for the blocks versions that was not present in #1936.
Reporting below in text of the initial PR and I will copy/paste the comments that were not resolved there.

--- 

Implementing the separation of blocks from the items as discussed in #48 (closes #48)

They key principles followed for the implementation are those outlined in https://github.com/datalab-org/datalab/issues/48#issuecomment-4811223759 (copy/pasted below for convenience)

<details>
<summary>Initial planning</summary>

1. **No new functionality**. The initial PR focuses exclusively on preserving existing functionalities with the blocks separated from the items. Some choices for the document structure may be driven by functionalities that are planned in the near future, but no additions here.

2. **Blocks as separated entities**. blocks are similar to `File`: the new block Model has parents `Entry + HasOwner + HasRevisionControl`. 
 
3. **Two collections: `data_blocks` + `block_versions`.** Similar to `items`/`item_versions`. An item will always point to the document in the `data_blocks` collection (i.e. the latest block version). Document in `item_versions` points to documents in `block_versions`. 

4. **Items collection**: the `blocks_obj` in the `items` collection can contain either the full block for backward compatibility or a dictionary with `{"immutable_id": ObjectId}`. The order will still be based on `display_order`. This is preparatory for the subsequent steps in which the blocks_obj` and `display_order` will be replaced by `{"blocks": [{"immutable_id": ...}]` to fit the suggested pattern (`{"type": "my-item", "blocks": [{"immutable_id": <ObjectId>}], "files": [{"immutable_id": <ObjectId>}]}`).
 
5. **No reference to the items in the block document**. Can be added in the future if needed.
 
6. **Versioning**. Like in the current data structure, modifying a block will not automatically trigger the creation of a new block version in  `block_versions`, but will just modify the document in `data_blocks`. A new version of the block in `block_versions` will be triggered by an item save, but only if the content of the block has changed from the previous block version stored, to avoid duplicated entries.
 
7. **Restore**. Same mechanism as items. Restoring an old item version writes the old block content back in the current state + a new history entry. At this stage this can happen only when restoring an Item entirely.
 
8. **No changes to File ownership**. Most blocks are associated to a file. Association will remain the same.
 
9. **No changes to the frontend API**. Responses still expose `blocks_obj` + `display_order`, reassembled server-side from `data_blocks`.
 
10. **Block deletion**. As in the current implementation, if a block is deleted it will be removed from the `data_blocks` collection, but preserved in the `block_versions` in case it needs to be restored. Block deletion triggered only when removed from an item. If an Item is deleted, the corresponding blocks in `data_blocks` are deleted (not those in `block_versions`, mimicking what happens for items). This can revised in the future if a block can be shared.

11. **Collections**. `Collection` is `HasBlocks`, but if I am not mistaken there is really no place where a block can be assigned to a collection at the moment in the UI, and it has no versioning. We keep the same structure as Items, but there should be no need to handle backward compatibility.

</details>

A few additional points to be specified concerning the details of the implementation:
* The access to the blocks is currently gated by the access to the item. There is no direct access to the blocks.
* The wat the blocks are stored is preserved. A legacy block will never be changed in a new kind of block here.
* It is not clear to me if for the collections the backward compatibile version should be supported. And at the moment it seems that there is no way of adding blocks to a collection. For this reason The collection blocks are left untouched. Adding the new blocking version with or without full backward compatibility should be trivial.
* Consistently with the current implementation, when a new block is created in the DB it is not versioned until the item to which it belongs to is saved. To handle this properly the `version` attribute of the block starts at `0` and becomes `1` when the item is saved.
* fixed a pre-existing (and apparently known issue): creating and then immediately deleting a block, did not truly remove it. Upon a save the removed block was recreated. Since it was related to the block management it seemed fine to add this fix here.

Note that I used the pydantic2 branch as a base for the implementation, as it seems reasonable as base, given the involvement of the backend. I targeted the PR to that branch, but the idea is to merge this in main after the pydantic2 branch has been merged.


pydatalab/src/pydatalab/models/blocks.py
```diff
@@ -72,3 +76,42 @@ class DataBlockResponse(BaseModel):
...
+
+# Here to avoid circular import
+class HasBlocks(BaseModel):
+    """Trait mixin for models that can have data blocks attached to them."""
+
+    blocks_obj: dict[str, DataBlockResponse] = Field({})
+    """A mapping from block ID to block data."""
+
+    display_order: list[str] = Field([])
+    """The order in which to display block data in the UI."""
```
gpetretto, line 89: From https://github.com/datalab-org/datalab/pull/1936#discussion_r3636229174:
@davidwaroquiers:
"Not sure I like it to have HasBlocks moved here somehow. It could be possible to move DataBlockResponse to another module (e.g. blocks_response.py or something) and keep HasBlocks in traits.py.
Not sure if it's worth. Maybe there is another solution also ? Maybe @ml-evs has an idea/preference or can comment on this."

@gpetretto:
"Indeed moving DataBlockResponse should be an option to avoid the circular import. I think this is a matter of preference."

pydatalab/src/pydatalab/models/blocks.py
```diff
@@ -72,3 +76,42 @@ class DataBlockResponse(BaseModel):
...
+    parent item's `blocks_obj`/`display_order` and in the DOM."""
+
+    blocktype: str
+    """A short string key specifying the type (technique) of the block."""
+
+    data: dict[str, Any] = Field(default_factory=dict)
+    """The block payload, exactly as produced by `DataBlock.to_db()`."""
+
+    version: int = 0
+    """The latest committed version number of this block in `block_versions`.
```
gpetretto, line 113: From https://github.com/datalab-org/datalab/pull/1936#discussion_r3636264754

@davidwaroquiers
"Wouldn't it be better to somehow follow a logic similar to items ? i.e. having a counter in the version_counter collection based on the immutable_id of the block ? What do you think ? I am actually a bit confused about the version that is in the item and the counter that is in the version_counter in the end. It seems they are always the same (version in item and counter in the corresponding document in version_counter). Probably good to clarify this with @ml-evs

After our discussion and regarding the logic of having version = 0 here, I propose we also discuss this with @ml-evs

Somehow we keep exactly the same behavior as it is currently but I have the feeling that we may want to change that somehow (to a better, more consistent behavior).

Behavior:
When we add a block, it already adds the block directly in the item without any save.

What I would expect (to be discussed of course :):
When we add a block in the item, it will add the block only when we save the item."

@gpetretto 
"indeed this is partly to keep exactly the symmetry with what the blocks are doing now.
Other ways of achieving the same would be possible, but maybe a bit more involved on the side of the code. Or maybe it could also be possible to alter the general behaviour of the backend. I did not verify if this would require any change in the routes."

pydatalab/src/pydatalab/blocks/store.py
```diff
@@ -0,0 +1,414 @@
...
+        # This should only happen if the counter document is corrupted.
+        return result["counter"]
+
+
+def is_block_reference(blocks_obj_value: Any) -> bool:
+    """Whether a stored `blocks_obj` value is a reference to a `blocks` document
+    (``{"immutable_id": ...}``) rather than a legacy embedded block.
+    """
+    return (
+        isinstance(blocks_obj_value, dict)
```
gpetretto, line 98: From https://github.com/datalab-org/datalab/pull/1936#discussion_r3636393867

@davidwaroquiers 
"Can it happen that "entry" is not a dict ? And if we assume it can, then this "non dict" object is then considered as a legacy embedded block. What would/could happen then ?"

@gpetretto 
"in general it should not. Since this is used in different places it is a protection against what comes in and does risk giving internal errors. I can remove the check or put a warning. Any preference?"

pydatalab/src/pydatalab/blocks/store.py
```diff
@@ -0,0 +1,414 @@
...
+        return last_version["version"]
+
+    # Allocate the next number from the shared atomic counter.
+    next_version = get_next_block_version_number(immutable_id)
+
+    version_entry = BlockVersion(
+        block_immutable_id=immutable_id,
+        block_id=block_doc["block_id"],
+        version=next_version,
+        timestamp=datetime.datetime.now(tz=datetime.timezone.utc),
```
gpetretto, line 324: From https://github.com/datalab-org/datalab/pull/1936#discussion_r3637811117

@davidwaroquiers 
"Why not using _now_isoformat ?"

@gpetretto 
"Because timestamp in BlockVersion is a datetime, like in ItemVersion, while _now_isoformat is used to fill the last_modified attribute of Block which comes from Entry and is of type IsoformatDateTime."

pydatalab/src/pydatalab/routes/v0_1/blocks.py
```diff
@@ -256,15 +257,34 @@ def add_data_block():
...
+    # the item only stores a reference to them. No version is committed until
+    # the next item version snapshot.
+    block_immutable_id = store.create_block_document(block)
+
     result = flask_mongo.db.items.update_one(
         {"item_id": item_id, **get_default_permissions(user_only=True)},
         {
             "$push": {"display_order": display_order_update},
-            "$set": {f"blocks_obj.{block.block_id}": block.to_db()},
+            "$set": {f"blocks_obj.{block.block_id}": {"immutable_id": block_immutable_id}},
```
gpetretto, line 282: From https://github.com/datalab-org/datalab/pull/1936/changes#r3638903276

@davidwaroquiers 
"Why not defining the id then do a find_one_and_update with the permissions, followed by the create_block_document ? In that case there is one less query to the items collection. The drawback is that then if the block creation fails we have a dangling reference to a block ? I guess that's why you did it this way ? Maybe a transaction would also be useful here."

@gpetretto (replying here)
"Indeed this would avoid the dangling reference. A transaction can be considered, but not sure if it is worth"

pydatalab/src/pydatalab/routes/v0_1/blocks.py
```diff
@@ -439,9 +479,18 @@ def delete_block():
...
+        return_document=ReturnDocument.BEFORE,
     )
 
-    if result.modified_count < 1:
+    stored_blocks_obj_value = (
+        (doc_before.get("blocks_obj") or {}).get(block_id) if doc_before else None
+    )
+
+    block_was_present = stored_blocks_obj_value is not None or (
+        doc_before is not None and block_id in (doc_before.get("display_order") or [])
```
gpetretto, line 491: From https://github.com/datalab-org/datalab/pull/1936#discussion_r3639063306

@davidwaroquiers 
"Can this happen that the block was not in the list of blocks of the items but it was in the display order ? That seems like a case where the block actually does not exist and cannot be deleted, just that there was some kind of bug or weird thing happening to the display order at some point."

@gpetretto 
"This would keep the equivalence with the original procedure. That one did find_one and the query was addressing the display_order. So in the case where the block_id was missing, but still in the display_order the document would have resulted as modified, returning a 200. Without this condition, instead the new version would return a 400. Here I switched to find_one_and_update to save one action on the DB, but then still tried to reproduce exactly the same logic. It can be questioned if the logic is correct (and properly preserved also...)"

pydatalab/src/pydatalab/routes/v0_1/items.py
```diff
@@ -1493,26 +1580,91 @@ def restore_version(refcode):
...
     user_id = None
     if current_user.is_authenticated:
         user_id = current_user.person.immutable_id
 
+    # Apply each snapshotted blocks_obj value in its own form: embedded payloads
+    # are written back inline as before, while the references get
+    # the payload from the `block_versions` and becomes the new current `blocks`
+    # state, plus a new RESTORED `block_versions` entry.
+    restored_block_pins: dict[str, dict] = {}
+    if "blocks_obj" in restored_data:
```
gpetretto, line 1606: From https://github.com/datalab-org/datalab/pull/1936#discussion_r3643408749

@davidwaroquiers 
"Not sure the if is needed here as the loop anyway uses "or {}""

@gpetretto 
"Removing it would have the impact on the second loop, that instead works on current_item. Again, this is a case that in general should not happen, but it is a cheap check. Can be considered for removal."

pydatalab/src/pydatalab/routes/v0_1/items.py
```diff
@@ -1493,26 +1580,91 @@ def restore_version(refcode):
...
+                    "Dropping unrestorable block reference %s (%s) while restoring %s",
+                    block_id,
+                    blocks_obj_value,
+                    refcode,
+                )
+                if isinstance(restored_data.get("display_order"), list):
+                    restored_data["display_order"] = [
+                        b for b in restored_data["display_order"] if b != block_id
+                    ]
+                continue
```
gpetretto, line 1635: From https://github.com/datalab-org/datalab/pull/1936#discussion_r3643435781

@davidwaroquiers 
"As this should not happen. I'm wondering if it should not be a BadRequest or something. It seems that if it happens it is a bug (from somewhere else) no ? Then if we "just" log it and drop, we may not discover it while using. What do you think ?"

@gpetretto 
"Simply raising here in the middle of the loop would risk leaving the db in an inconsistent state. One would probably need to make a full check before and raise if one block is missing, and if not proceed with the update. Do you think it would be better?"

gpetretto (commented):
Import pending comments from #1936.

ml-evs (approved):
Hi @gpetretto, thanks for the hard work on this. The only way for me to really get this is in is to review the functionality and skim the code, but I'm convinced that things work and have pushed a couple of minor changes in extra commits (adding one more test, moving some things out of the item routes). I've also tried to follow your conversation with @davidwaroquiers across the two PRs but it seems like most of it has been resolved and does not need my attention, so I will resolve the comments.

I think block versioning is the part that I'm most concerned about, but happy to merge and treat this as experimental for a little while until we're convinced. I don't really know what we expect the UI for block versions to look like yet, but perhaps that would be a nice thing to work on next.

davidwaroquiers:
> Hi @gpetretto, thanks for the hard work on this. The only way for me to really get this is in is to review the functionality and skim the code, but I'm convinced that things work and have pushed a couple of minor changes in extra commits (adding one more test, moving some things out of the item routes). I've also tried to follow your conversation with @davidwaroquiers across the two PRs but it seems like most of it has been resolved and does not need my attention, so I will resolve the comments.
> 
> I think block versioning is the part that I'm most concerned about, but happy to merge and treat this as experimental for a little while until we're convinced. I don't really know what we expect the UI for block versions to look like yet, but perhaps that would be a nice thing to work on next.

Hi @ml-evs thanks for the hard work also on merging this (and the other PRs). Block versioning was indeed the part that was most complex to think about. In principle the current PR keeps the versioning exactly as it was before (i.e. you can't really have a block version without the item, the version is in the item itself, @gpetretto correct me if I'm wrong) and we should indeed think carefully how we want to deal with block versions themselves.

PR #2073, opened 2026-09-02
custom item UI

jbouquiaux:
## Summary

This PR supersedes the closed PR https://github.com/datalab-org/datalab/pull/1846 and brings the custom-item UI work onto the main branch.

This adds:

- dynamic registration, creation, display and editing of plugin item types;
- automatic rendering of simple custom fields from validated `datalab_*` schema hints;
- optional custom Vue panels for more complex interfaces;
- automatic panel collection during native and Docker builds (new with respect to the previous PR);
- canonical storage for quantities while preserving the user's display-unit preference separately (new handling discussed with @davidwaroquiers);

There is also now a copier template to help with plugin creation, and I extracted the two examples out of datalab. 

- [Copier plugin template](https://github.com/Matgenix/datalab-item-plugin-template)
- [Schema-rendered example plugin](https://github.com/Matgenix/datalab-item-plugin-example)
- [Custom Vue panel example](https://github.com/Matgenix/datalab-item-plugin-example-custom-vue)

This currently live in the Matgenix organization, but this would make sens to move at least the template to datalab-org?


pydatalab/docs/plugins.md
```diff
@@ -139,14 +150,179 @@ As a rule of thumb, if you would want to *filter* the item list by it, it is a f
...
+resulting mixture live — none of which the core panel can do on its own.
 
-Eventually, such item types will allow for rich descriptions of unitful quantites, semantic annotations and URIs for fields, and cross-linking between items via custom relationships.
+!!! warning "Custom panels are trusted, compiled code"
+    Panel `.vue` files are compiled into the webapp bundle and run in every user's browser, so
+    installing or updating a UI plugin means rebuilding the frontend. The Docker commands above
+    perform collection before the development or production frontend is compiled. Only install
+    panels from sources you trust.
 
 ## Plugin installation
```
gpetretto: Not strictly in this PR, buy the title of the previous section is `Installing plugins`, this is `Plugin installation`. Maybe there could be a more suitable title?

webapp/src/components/CreateItemModal.vue
```diff
@@ -170,6 +170,18 @@ export default {
...
     },
+    effectiveAllowedTypes() {
+      // The passed allowedTypes plus any dynamically-registered (custom/plugin)
+      // creatable types whose base type fits this modal's context, so e.g. a
+      // samples modal offers sample-derived custom types but not equipment-derived ones.
+      const dynamic = Object.keys(this.$store.state.schemas || {}).filter(
+        (type) =>
+          itemTypes[type]?.isDynamic &&
+          itemTypes[type]?.isCreateable &&
+          this.allowedTypes.includes(itemTypes[type]?.baseType),
```
davidwaroquiers, line 181: I tried to make a custom type deriving from Item. The backend registers it and it can be created through /new-sample/, but /info/types returns base_type: null, so this filter never offers it in any creation modal. Also related to https://github.com/Matgenix/datalab-item-plugin-template/issues/1

webapp/src/components/custom/CustomFieldsPanel.vue
```diff
@@ -0,0 +1,400 @@
...
+      const transform = field.quantity.transforms[unit];
+      const value =
+        rawValue === "" ? null : Number(rawValue) * transform.scale + (transform.offset || 0);
+      const itemData = { [field.name]: value };
+      if (field.quantity.displayUnitField) {
+        itemData[field.quantity.displayUnitField] = unit;
+      }
+      this.$store.commit("updateItemData", { item_id: this.item_id, item_data: itemData });
+    },
+    updateDisplayUnit(field, unit) {
```
davidwaroquiers, line 335: After our discussion, thinking and trying, I would maybe ask the user what he wants to do when changing the unit of an existing value? Currently, this keeps the displayed number unchanged and reinterprets it in the new unit, which silently changes the canonical value, for example, changing 20 degree C to Fahrenheit results in 20 Fahrenheit rather than 68 Fahrenheit. A small confirmation dialog could offer “Convert the value” (20 degree C to 68 Fahrenheit) or “Keep the number unchanged” (20 degree C => 20 Fahrenheit), showing the resulting value so the difference is clear. If the field is still empty in the database, we can simply change the selected unit without prompting or converting anything. This would make potentially significant scientific-data changes explicit and avoid accidental conversions.

webapp/src/components/CreateItemModal.vue
davidwaroquiers: Note that I managed to add a sub-Item custom type ("my sub item" type, custom-item-002) through the API and it appears in the Samples navigation bar:

Image

Clicking on it shows nothing:

Image

pydatalab/src/pydatalab/routes/v0_1/info.py
```diff
@@ -161,6 +162,57 @@ def list_block_types():
...
+    best_model: type | None = None
+    for builtin_type, builtin_model in ITEM_MODELS.items():
+        if builtin_type not in BUILTIN_ITEM_TYPES:
+            continue
+        if issubclass(model, builtin_model):
+            # Prefer the most specific (most-derived) built-in base.
+            if best_model is None or issubclass(builtin_model, best_model):
+                best_type = builtin_type
+                best_model = builtin_model
+    return best_type
```
davidwaroquiers, line 186: This is probably one place where the issue with creation of Item-derived (not Sample-derived) custom types originates. (see https://github.com/datalab-org/datalab/pull/2073/changes#r3932752960)
As Item is not a builtin item type, it gets best_type None for non-Sample-derived (or other builtin item types) custom types

webapp/src/server_fetch_utils.js
```diff
@@ -1347,6 +1353,16 @@ export async function loadItemSchemas() {
   // in the Vuex store.
   try {
     const supportedTypes = await getSupportedSchemasList();
+    // Register any server-side types not hardcoded in the frontend registry
+    // (custom/plugin types) so they are recognised across the app.
+    supportedTypes.forEach((typeInfo) =>
+      registerDynamicItemType(typeInfo.id, {
+        title: typeInfo.attributes?.title,
+        base_type: typeInfo.attributes?.base_type,
```
davidwaroquiers, line 1361: Here, base_type is null for Item-derived custom types I think  so it probably links to problem here: https://github.com/datalab-org/datalab/pull/2073/changes#r3932752960

webapp/src/resources.js
```diff
@@ -75,7 +80,10 @@ export const customBlockTypes = {
   "insitu-xrd": { description: "XRD insitu", component: XRDInsituBlock, name: "XRD insitu" },
 };
 
-export const itemTypes = {
+// Reactive so that components re-render when dynamic (plugin) types are
```
davidwaroquiers, line 83: Does this mean that we can install a plugin on a deployed server without redeploying ? Or it is there to make sure everything is loaded if there are some concurrent "things" started ?

webapp/src/resources.js
```diff
@@ -137,8 +145,63 @@ export const itemTypes = {
     labelColor: "#9C2007",
     display: "block",
   },
+});
+
+// Maps built-in backend type strings to their base information components.
+const BASE_TYPE_COMPONENTS = {
```
davidwaroquiers, line 151: For Item-derived custom types, we probably need a base ItemInformation component.

webapp/src/resources.js
```diff
@@ -137,8 +145,63 @@ export const itemTypes = {
...
+    navbarColor: color,
+    navbarName: display,
+    lightColor: lightTint(color),
+    labelColor: color,
+    isCreateable: true,
+    display: display.toLowerCase(),
+    isDynamic: true,
+    baseType: base_type || null,
+    hiddenFields: hidden_fields || [],
+    customPanel: PLUGIN_PANELS[type] || null,
```
davidwaroquiers, line 201: If it's a customPanel, what happens with the color (in particular the lightTint) ?

Maybe in the future (not now, not in this PR) we could also imagine having the possibility to customize more of these.

webapp/src/components/custom/CustomFieldsPanel.vue
```diff
@@ -0,0 +1,400 @@
...
+}
+/* Wraps a FormattedItemName so a selected reference reads as a filled field; the
+   badge inside is the click target, so this box is intentionally not itself a link. */
+.ref-display {
+  display: flex;
+  align-items: center;
+  background-color: #f8f9fa;
+  flex: 1;
+}
+</style>
```
davidwaroquiers, line 400: Here, these are hardcoded. Shouldn't there be some link/reuse of the schema hints colors if they are defined ?
More broadly, how can we define a theme/set of colors/css variables for a given custom type.
Not in this PR, but just mentioning.

pydatalab/tasks.py
```diff
@@ -112,13 +119,169 @@ def generate_schemas(_):
...
+    return f"{class_name}Panel"
+
+
+def _panel_item_type(entry_point) -> str:
+    """Return the server-facing type registered for an item plugin panel."""
+    model = entry_point.load()
+    item_type = model.model_json_schema()["properties"]["type"]["default"]
+    if not item_type.startswith("_"):
+        item_type = f"_{item_type}"
+    return item_type
```
davidwaroquiers, line 150: This is a broader question also on the backend. Pinging @ml-evs for feedback. I am wondering if it would not be better to have builtin datalab types be prefixed with e.g. "core-", hence being core-samples, core-equipment, core-starting-materials, ... and have plugins have their own slug prefix (enforcing that all installed plugins 1/ have different slugs and 2/ of course cannot be "core-"). Somehow that would make things more readable in a way. What do you think ?

pydatalab/tasks.py
```diff
@@ -112,13 +119,169 @@ def generate_schemas(_):
...
+    (class name + "Panel").
+
+    ``dev.install`` runs this automatically for a complete source checkout. The
+    explicit task remains useful for troubleshooting or refreshing generated
+    output without reinstalling plugins.
+    """
+    try:
+        _collect_plugin_panels()
+    except FileNotFoundError as exc:
+        raise SystemExit(str(exc)) from exc
```
davidwaroquiers, line 279: Mentioned this also in another comment. Can we install a new plugin and register it without redeploying completely ? If that is the case (which would be nice but not needed, at least this stage), I would maybe document this a bit more. If it's not possible, no need to change anything to this PR

davidwaroquiers (changes requested):
Thanks! Great work, I have left a couple of comments.
The main one is probably the one about custom types derived from Item, which is currently not working.
Another important comment but also related to the backend part, is about the naming/prefix for builtins/core vs plugin provided types. Pinging @ml-evs for this, to see what to do there.

webapp/src/components/CreateItemModal.vue
davidwaroquiers: Discussion with @jbouquiaux: all custom types will land in the Samples page for now.

webapp/src/components/custom/CustomFieldsPanel.vue
davidwaroquiers: Discussion: popup with choice between the two

davidwaroquiers: Would likely be better to move DatalabQuantityExtra (and rename it DatalabQuantity) and DatalabQuantityAffine (not sure if it's the right name) to a models/utils.py or models/quantity.py or models/units.py.

That would open the possibility to use this elsewhere without having to rely on something in schema_hints which is not a very clear place for that.

webapp/src/resources.js
davidwaroquiers: Need to redeploy. Closing this

pydatalab/tasks.py
davidwaroquiers: Need to redeploy. Closing this

pydatalab/docs/plugins.md
```diff
@@ -139,14 +150,179 @@ As a rule of thumb, if you would want to *filter* the item list by it, it is a f
...
+annotations on the field:
+
+| `Field(json_schema_extra=…)` key | Effect in the UI |
+|---|---|
+| `datalab_include_field_in_summary` | also show the field as a column in list / summary views |
+| `datalab_hidden` | store the field but don't render it directly |
+| `datalab_quantity` | declare canonical storage plus plugin-defined display-unit conversions for a floating-point field |
+| `datalab_ref_types` | render as an item-search selector restricted to these item types — i.e. a link to another item (built-in *or* custom) |
+| `datalab_section` | group this field into its own titled card |
+| `datalab_multiline` | render a string as a multi-line text area |
```
gpetretto, line 176: Some of the properties of `DatalabFieldExtra` are not mentioned in this table (e.g. `datalab_exclude_from_db`). Is it on purpose or maybe those options were added later?

pydatalab/docs/plugins.md
```diff
@@ -139,14 +150,179 @@ As a rule of thumb, if you would want to *filter* the item list by it, it is a f
...
+
+A few keys on the model's `model_config` control the type as a whole:
+
+| `ConfigDict(json_schema_extra=…)` / config key | Effect in the UI |
+|---|---|
+| `title` | display name of the type (navbar, create dialog) |
+| `datalab_ui_color` | accent colour for the navbar, field labels and the item's reference badge |
+| `datalab_ui_hidden_fields` | base-component sections to hide (`status`, `collections`, `description`, `substance_information`, `synthesis_information`) |
+| `datalab_section_title` | title of the default custom-fields card |
```
gpetretto, line 185: Here `datalab_base_type` from `DatalabModelExtra` is missing. Not sure if it is needed.

pydatalab/src/pydatalab/models/schema_hints.py
```diff
@@ -0,0 +1,242 @@
...
+            raise ValueError(f"Invalid datalab model schema hints on {name!r}: {exc}") from exc
+
+    schema_properties = model.model_json_schema(by_alias=False).get("properties", {})
+
+    claimed_display_unit_fields: dict[str, str] = {}
+
+    for field_name, field in model.model_fields.items():
+        field_extra = field.json_schema_extra
+        if not isinstance(field_extra, dict):
+            continue
```
gpetretto, line 124: Should this raise (or at lease warn) if it is not a dict and not `None`?

pydatalab/docs/plugins.md
```diff
@@ -139,14 +153,173 @@ As a rule of thumb, if you would want to *filter* the item list by it, it is a f
...
+resulting mixture live — none of which the core panel can do on its own.
 
-Eventually, such item types will allow for rich descriptions of unitful quantites, semantic annotations and URIs for fields, and cross-linking between items via custom relationships.
+!!! warning "Custom panels are trusted, compiled code"
+    Panel `.vue` files are compiled into the webapp bundle and run in every user's browser, so
+    installing or updating a UI plugin means rebuilding the frontend. The Docker commands above
+    perform collection before the development or production frontend is compiled. Only install
+    panels from sources you trust.
 
 ## Plugin installation
```
gpetretto: Not strictly in this PR, buy the title of the previous section is `Installing plugins`, this is `Plugin installation`. Maybe there could be a more suitable title?

pydatalab/docs/plugins.md
```diff
@@ -139,14 +153,173 @@ As a rule of thumb, if you would want to *filter* the item list by it, it is a f
...
+        json_schema_extra={
+            "datalab_include_field_in_summary": True,
+            "datalab_quantity": {
+                "canonical_unit": "mol/L",
+                "display_units": {
+                    "mol/L": {"scale": 1.0},
+                    "mmol/L": {"scale": 0.001},
+                },
+                "default_display_unit": "mol/L",
+                "display_unit_field": "concentration_display_unit",
```
gpetretto, line 229: `display_unit_field` is not very well documented. Not mentioned in the table above and from the code I could see that the validation is strict in the fact that the possible values in the `concentration_display_unit` should match those in this `display_units`. Probably would helpful to clarify.

webapp/src/components/CreateItemModal.vue
jbouquiaux: This should be fixed with https://github.com/datalab-org/datalab/pull/2073/changes/9bbeaeefe28cbf5e7ad04c22e4ce20337df950d7.

Direct `Item` subclasses are now with `base_type:"items"` and inherited `base_fields`. They use the new Item information panel. 

And as discussed, for now every registered custom type is created and listed through the Samples page.

webapp/src/components/custom/CustomFieldsPanel.vue
jbouquiaux: Screenshot 2026-09-07 at 15 54 41


I've added this small pop up so that changing the unit is unambiguous about what is stored in the DB (https://github.com/datalab-org/datalab/pull/2073/changes/7edb29346763e40657261e7e76483439def14cd2
)

jbouquiaux: > Would likely be better to move DatalabQuantityExtra (and rename it DatalabQuantity) and DatalabQuantityAffine (not sure if it's the right name) to a models/utils.py or models/quantity.py or models/units.py.
> 
> That would open the possibility to use this elsewhere without having to rely on something in schema_hints which is not a very clear place for that.

I agree. Done in https://github.com/datalab-org/datalab/pull/2073/changes/1d9b5020d9ba8c3f5a1baaaddfc130be9ea3c11b

pydatalab/tasks.py
jbouquiaux: I'd say that built-in type identifiers should remain unchanged to avoid migration and compatibility issues. Now for plugin types, I agree that this would make things more readable. I did not touch this at the moment since this might requires more discussion.

webapp/src/resources.js
jbouquiaux: A custom panel does not change the type-color. Custom types use datalab_ui_color when provided, with lightTint derived from it. Otherwise they receive the existing fallback color. Sure more elaborate theme customization could be handled later.

jbouquiaux:
> Thanks! Great work, I have left a couple of comments. The main one is probably the one about custom types derived from Item, which is currently not working. Another important comment but also related to the backend part, is about the naming/prefix for builtins/core vs plugin provided types. Pinging @ml-evs for this, to see what to do there.

Thanks for the tests and comments @davidwaroquiers, I've tried to address them, see above.

pydatalab/src/pydatalab/routes/v0_1/info.py
```diff
@@ -161,6 +162,60 @@ def list_block_types():
...
+        return None, None
+
+    best_type: str | None = "items"
+    best_model: type[Item] | None = Item
+    for builtin_type, builtin_model in ITEM_MODELS.items():
+        if builtin_type not in BUILTIN_ITEM_TYPES:
+            continue
+        if issubclass(model, builtin_model):
+            # Prefer the most specific (most-derived) built-in base.
+            if best_model is None or issubclass(builtin_model, best_model):
```
gpetretto: Can `best_model` be `None` here?

pydatalab/src/pydatalab/routes/v0_1/info.py
```diff
@@ -161,6 +162,60 @@ def list_block_types():
...
+    return best_type, best_model
+
+
+def _get_model_schema_extra(item_type: str) -> dict:
+    """Return the `json_schema_extra` dict declared on the model's config, if any."""
+    model = ITEM_MODELS.get(item_type)
+    if model is None:
+        return {}
+    extra = model.model_config.get("json_schema_extra") or {}
+    if callable(extra):
```
gpetretto, line 196: Which case does this handle?

pydatalab/src/pydatalab/models/units.py
```diff
@@ -0,0 +1,74 @@
...
+
+from pydatalab.models.utils import BaseModel
+
+
+class DatalabUnitTransform(BaseModel):
+    """Affine conversion from a display unit to a field's canonical unit."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    scale: float = 1.0
```
gpetretto: This could be defined 
```python
scale:  float = Field(default=1.0, gt=0, allow_inf_nan=False)
```
and avoid the custom validation below. Same for `offset`.

pydatalab/src/pydatalab/routes/v0_1/items.py
```diff
@@ -1748,10 +1756,19 @@ def save_item():
     if isinstance(existing_last_modified, datetime.datetime):
         existing_last_modified = existing_last_modified.isoformat()
 
+    # Fields explicitly set to null in the incoming update must be cleared in the DB.
+    # model_dump(exclude_none=True) silently drops them from `item`, so we $unset separately.
+    # Skip any field still present in `item` (e.g. re-filled by a model validator):
+    # the same path in both $set and $unset is a MongoDB conflict error.
+    fields_to_unset = {k for k, v in updated_data.items() if v is None and k not in item}
```
gpetretto: Doesn't this change slightly the logic? It seems that a field that was explicitly set to `None` would still be set in the db. Now it is unset. 
It seems that this implicitly assumes that any nullable field is defined with `None` as a default. Or am I missing something?

gpetretto (commented):
Thank @jbouquiaux.
Here are my comments.

pydatalab/docs/plugins.md
jbouquiaux: This is intentional.  `datalab_exclude_from_db/load` are hints from prior work and used in blocks, not for custom type rendering. They are still included in `DatalabFieldExtra` so that the new schema also describes them.

jbouquiaux: This is intentional. datalab_exclude_from_db/load are hints from prior work and used in blocks, not for custom type rendering. They are still included in DatalabFieldExtra so that the new schema also describes them.

pydatalab/docs/plugins.md
jbouquiaux: Actually, `datalab_base_type`is no longer used now that the base type resolution is done only through the model hierarchy. I'll remove this hint and regenerate the schemas

pydatalab/docs/plugins.md
jbouquiaux: Sure, I would suggest something like "Deployment integration" here since this describes Docker/Ansible integration.

pydatalab/src/pydatalab/models/schema_hints.py
jbouquiaux: Done

pydatalab/docs/plugins.md
jbouquiaux: I agree, I'll clarify that

pydatalab/src/pydatalab/routes/v0_1/info.py
jbouquiaux: No, it's initialized to `Item`. I'll change that

pydatalab/src/pydatalab/models/units.py
jbouquiaux: I agree, this is much more direct.

pydatalab/src/pydatalab/routes/v0_1/info.py
jbouquiaux: This handles the callable form supported by Pydantic. Since we cannot inspect it here like a dictionary, we treat it as having no Datalab-specific hints. I’ll add a comment to clarify that.

pydatalab/src/pydatalab/routes/v0_1/items.py
jbouquiaux: I think you're right. This was based on an incorrect assumption. The save path use `exclude_unset=True`, so a `None` is retained and stored as `null`. Using $unset could change the distinction between a missing field and one set to `None`. I'll revert that change and its test

jbouquiaux (commented):
Hi @gpetretto, thanks for your comments. I've tried to address them.

PR #2079, opened 2026-09-03
Add routes and event handling for hot-loading data into bokeh

jdbocarsly:
I am currently playing with adding a data schema to the private magnetometry block, with the goal of contributing a clean version of the block updates back to datalab once we are happy it. The goal is to have parsed data validated with a  pydantic schema, and stored in the db. The API will then be able to serve data directly. 

As part of this, I am working on the ability to request data in different units (e.g., temperature in K vs °C or diffraction data in 2θ or Q). This requires a new route and changes to the bokeh event handling in the frontend, so it can't live in the magnetometry plugin alone. These core changes are being made in this branch, which will co-develop with the plugin.

Note: besides the new changes, this also migrates the existing bokeh event handling from DataBlockBase.vue to BokehPlot.vue

### DIFF
--- a/pydatalab/src/pydatalab/routes/v0_1/blocks.py
+++ b/pydatalab/src/pydatalab/routes/v0_1/blocks.py
@@ -457,6 +457,62 @@ def delete_block():
     )  # could try to switch to http 204 is "No Content" success with no json
 
 
+@BLOCKS.route("/blocks/<string:item_id>/<string:block_id>/data", methods=["GET"])
+def get_block_data(item_id: str, block_id: str):
+    """Serve individual data columns from a block, in the units asked for.
+
+    Each `column` query parameter names one column and, after a colon, the units
+    to give it in; omitting the units gives the column as it is stored:
+
+        ?column=moment:emu%2Fg&column=temperature
+
+    This exists so that a plot can change what it shows without the block being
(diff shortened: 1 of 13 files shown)


pydatalab/src/pydatalab/routes/v0_1/blocks.py
ml-evs, line 1: Just remember that there is also https://github.com/datalab-org/datalab/pull/1737 that started adding things like this; simply removing the item_id from the block route should be enough to ensure its forwards-compatible with #2069

ml-evs (commented):
Looks interesting, will have to play around with it!

pydatalab/src/pydatalab/routes/v0_1/blocks.py
jdbocarsly: yes, noted. This, combined with https://github.com/datalab-org/sepPya/pull/3 (private repo atm, but the new block base and schema could eventually be contributed back to enhance the core block) would make that sort of data-providing behavior generic on any block with a schema. The schema also get more involved, and can include unit conversions.

One thing to note: in #1737, I think the XRD pattern is still parsed each time and not saved in the db, since it is in processed? I was thinking the data could be stored in the database directly , since a typical data set is <<1 MB. XRD is also typically < 1 MB. We would need a fallback to handle the case when the data is too large, though. Do you think this is worthwhile?

PR #2095, opened 2026-09-11
Remove defunct `trigger_async` causing `KeyError` when using in situ plugin

jrneilson:
In-situ XRD/NMR/UVVis blocks trigger their async processing step by sending a bare
`{"trigger_async": true}` event (no `event_name`) from the webapp. `DataBlock.process_events`
did `event.pop("event_name")` with no default, so this raised `KeyError: 'event_name'`,
surfaced by the async worker as `Error during processing: 'event_name'` — unconditionally,
for every in-situ block use, regardless of the uploaded data.

## Root cause chain
1. `webapp/src/components/datablocks/XRDInsituBlock.vue` (`onFolderSelected`/`onGranularitySubmit`)
   calls `updateBlockFromServer(..., { trigger_async: true })` when folder names are set.
   The NMR/UVVis in-situ blocks do the same.
2. `pydatalab/src/pydatalab/routes/v0_1/blocks.py` schedules this as an async task and passes
   `event_data` straight through to the background worker.
3. The worker calls `block.process_events(event_data)` since `event_data` is truthy.
4. `pydatalab/src/pydatalab/blocks/base.py`'s `process_events` does `event.pop("event_name")`
   with no default → `KeyError: 'event_name'`, since the dict only has `trigger_async`.
5. Caught by the async worker's outer handler and recorded as `Error during processing: 'event_name'`.

## Fix
`event.pop("event_name", None)` — a control-only event is now treated as unregistered
and skipped rather than raising.

## Test plan
- Added `test_process_events_ignores_events_without_event_name`, covering the exact
  `{"trigger_async": true}` shape sent by the in-situ blocks.
- Existing `test_base_block` still passes, confirming real events still dispatch correctly.
- Verified end-to-end: uploading an in-situ XRD zip and setting folder names through the
  web UI no longer errors, and the block loads.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_018B6NG9uZJ66A6yMxpvaw4t

### DIFF
--- a/pydatalab/src/pydatalab/blocks/base.py
+++ b/pydatalab/src/pydatalab/blocks/base.py
@@ -262,8 +262,10 @@ class DataBlock:
             events = [events]
 
         for event in events:
-            # Match the event to any registered by the block
-            if (event_name := event.pop("event_name")) in self.event_names:
+            # Match the event to any registered by the block. Events without an
+            # `event_name` (e.g. a bare `{"trigger_async": True}` control flag sent
+            # by the in-situ blocks) are not registered events, so skip them.
+            if (event_name := event.pop("event_name", None)) in self.event_names:
                 # Bind the method to the instance before calling
                 bound_method = self.__class__.events_by_name[event_name].__get__(
                     self, self.__class__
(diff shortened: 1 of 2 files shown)


ml-evs (approved):
Thanks for raising this @jrneilson and sorry for the delay -- I've added an extra commit that fixes the root cause of this bug in the UI side ready for the next release.

PR #2099, opened 2026-09-16
Add location field to all items, sourced from the /locations API

### LINKED ISSUE #2049: Add locations UI block to all items and use `/locations` endpoint
As we move to merge #2001, we need to:
- [x] add the locations UI component to all items
- [x] populate the allowed values with the `/locations` endpoint

BenjaminCharmes:
Closes #2049

Replaces the client-side scan of `starting_material_list`/`equipment_list` with the new `GET /locations` endpoint (#2001), cached in Vuex (`locations_list`). Suggestions now cover every location the current user has access to (plus `PREDEFINED_LOCATIONS`), instead of only items already loaded into those two lists.

Also adds the `LocationInput` field to `SampleInformation.vue` and `CellInformation.vue` (samples/cells didn't have it at all before).


ml-evs (approved):
Thanks for this @BenjaminCharmes! I added a couple of commits to do the following:

- use the nested hierarchy list to pre-populate the suggestions without needing the user to start typing each segment
- added a component test
- added a shared getter across all pages to grab the locations once and cache them

PR #2104, opened 2026-09-16
Redesign account settings to improve usability

### LINKED ISSUE #1481: Add help modal to the API key section of account settings
This modal could show some useful info about what the API key is and how to use it with e.g., tabbed examples for `cURL`, the datalab-python API etc.

### LINKED ISSUE #1643: Updating user display name unecessarily triggers email verification pipeline

ml-evs:
This PR redesigns the basic account settings modal into a tabbed layout to prevent the submit button being lost.

It also adds a series of extra documentation to be clear around how to update avatars, what email verification means, how to connect accounts etc and improves the view of the user activity with some summary stats.

It also:

- Closes #1481 with some example key usage and links to docs
- Closes #1643 by gating email verification in both server and app

It now looks like this:

image

image

image


davidwaroquiers:
Hi @ml-evs 

Very nice! Just a couple of usability comments (just deployed the branch for testing), I did not look at the code/implementation (yet, I can do that maybe after some UX fixes if you think they are relevant).

- The Submit button is a bit "weird". I'd call it Save instead
- I'd have the Submit button inactive when nothing has been modified. Note that it is inactive when it is in "email adding mode":

image

- Related to submit/save, I would maybe slightly change the flow to add an email:
- When clicking "Add an email address", you would have an additional "Add" button to "confirm" the addition together with the "Cancel" button (as now). You actually can't add two addresses at the same time and it's a bit weird that when you click Submit, it directly closes the thing.
- Another option could be to have another Modal for adding the mail address.
- In any case, the added mail(s) would only be saved when the submit/save button is clicked
- I did not find a way to define a mail as primary except by re-adding it (it does not add an additional one, the primary mail is just the last one added).
- correction: I found out how the primary can be modified, just that it was a bit confusing for me. Maybe it would be easier to have only one "Email addresses" section, where you can add an email, set any of the mails there as primary (primary may then be moved up with the blue color for example)), and also allow to delete (did not find how to delete a mail either)
- Not sure if this is envisioned (or useful or whatever), but I did not find a way to add another "Connected account". I feel like it's not really necessary but just wanted to mention.
- In the api keys, when the modal has been closed, you can't get back the api key (which is good), but the copy button stays there and active, I would either disable it or remove it and have some kind of message that says that the key is not shown on purpose

ml-evs:
Thanks for the feedback @davidwaroquiers!

>     * The Submit button is a bit "weird". I'd call it Save instead


>     * I'd have the Submit button inactive when nothing has been modified. Note that it is inactive when it is in "email adding mode":
> 

Agreed, done both
 
>     * Related to submit/save, I would maybe slightly change the flow to add an email:
> 
>     * When clicking "Add an email address", you would have an additional "Add" button to "confirm" the addition together with the "Cancel" button (as now). You actually can't add two addresses at the same time and it's a bit weird that when you click Submit, it directly closes the thing.
> 
>     * Another option could be to have another Modal for adding the mail address.
> 
>     * In any case, the added mail(s) would only be saved when the submit/save button is clicked
> 
>     * I did not find a way to define a mail as primary except by re-adding it (it does not add an additional one, the primary mail is just the last one added).
> 
>     * correction: I found out how the primary can be modified, just that it was a bit confusing for me. Maybe it would be easier to have only one "Email addresses" section, where you can add an email, set any of the mails there as primary (primary may then be moved up with the blue color for example)), and also allow to delete (did not find how to delete a mail either)

These are a bit more tricky as there's no dedicated endpoint, I think this PR is an improvement on the previous but would be happy to rework in a later PR if its unclear.
 
>     * Not sure if this is envisioned (or useful or whatever), but I did not find a way to add another "Connected account". I feel like it's not really necessary but just wanted to mention.

I'm not sure I follow, but you can add any connected account that the server supports. We expect only one account per "type", e.g., 1 ORCID, 1 GitHub and I think I'd prefer to keep that constraint for now. If a server implements a new type (in e.g., a fork) then as long as they return the correct info in the info endpoint the UI here should work (or require minimal changes). Again this could be revamped properly in a future PR.

>     * In the api keys, when the modal has been closed, you can't get back the api key (which is good), but the copy button stays there and active, I would either disable it or remove it and have some kind of message that says that the key is not shown on purpose

Fixed, thanks!

Issue #2103, opened 2026-09-16
Clarify Windows support in documentation and make memray optional


ml-evs:
Reported by @alioramuss -- memray does not support Windows, so our full dev environment cannot be used on Windows. This may be the only issue, though I think many parts of the server require Linux (definitely remote filesystems and various parts may use hardcoded POSIX-style paths that would need to be adapted). This needs to at least be clarified, and the hard blocker of memray removed.

ml-evs:
Closed by #2106

PR #2106, opened 2026-09-19
Skip pytest-memray on Windows so the dev group installs

alioramuss:
Following up on the #development thread about Windows dev environments.

memray has no Windows support. The build refuses outright with RuntimeError: memray does not support this platform (win32). It comes in through pytest-memray, so uv sync --all-extras --dev --locked fails and takes the whole dev group down with it rather than just the profiler, which is why I ended up on --no-dev with no pre-commit or mkdocs either.

This gates pytest-memray behind sys_platform != 'win32' and registers the limit_memory mark in the pytest ini. The mark registration is needed because test_blocks.py::test_large_fake_xrd_data_block_serialization uses it, and with filterwarnings = ["error"] an unregistered mark becomes a collection error wherever the plugin is absent.

Linux and macOS are unchanged: uv export --only-group dev still yields memray 1.19.3 and pytest-memray 1.9.0 there, and the lock diff is two marker annotations with no version churn. Lock regenerated with uv 0.11.8 to match the CI pin.

With this applied the full suite runs on Windows 10 x64, Python 3.11.16, MongoDB 8 locally. Most of the remaining failures are one separate bug, in the companion PR.

Addresses #2103


ml-evs (changes requested):
Looks fine, can you also remove the AI canary comments here please?

I would also suggest finding the tests that use memray and adding a pytest.skip if the current platform is win32.

alioramuss:
both done. canary comments are gone from pyproject and the test file

added the skipif to test_large_fake_xrd_data_block_serialization, which is the only test carrying the limit_memory mark. i kept the marker registration in the ini alongside it though, because the skip on its own doesn't get you there. PytestUnknownMarkWarning is raised when the decorator is evaluated at import time, before skipif is looked at, so with filterwarnings = ["error"] the module still fails to collect when pytest-memray isn't installed. checked both ways on pytest 9.1.1: skipif alone gives a collection error, skipif plus the registration gives a clean skip

verified on windows with no -W override, test_blocks.py collects and reports SKIPPED tests/server/test_blocks.py:826: pytest-memray is unavailable on Windows

also dropped the canary comments on #2107 while i was there. happy to take the registration out if you'd rather handle it a different way

ml-evs (approved):
Thanks for this @alioramuss, super helpful to test this on Windows. I've added an extra commit to the docs describing our support Windows (i.e., its not official, but if changes like this and the other PR get things working 99% then we're happy to make them).

I think the test skip is more appropriate here as the functionality covered by that test is already tested elsewhere, and this case specifically is testing for memory blowups. I'd rather this be clearly marked as untested on Windows (though I find it very unlikely that it will break).

PR #2107, opened 2026-09-19
Use `shutil.disk_usage` instead of `os.statvfs` for free space

alioramuss:
os.statvfs exists only on Unix, so get_space_available_bytes() raises AttributeError: module 'os' has no attribute 'statvfs' on Windows. It is called by save_uploaded_file(), so every POST /upload-file/ returns a 500. In a full Windows run this accounted for 17 of the 21 failures, across test_blocks, test_files, test_backup and test_starting_materials.

shutil.disk_usage().free reports the same quantity and works everywhere. On POSIX it is f_bavail * f_frsize rather than f_bavail * f_bsize, which is the more correct of the two where fragment and block sizes differ.

Adds tests/test_file_utils.py covering the happy path and the uninitialised-directory RuntimeError.

Verified on Windows 10 x64, Python 3.11.16, MongoDB 8 running locally. Before: 21 failures. After: 5. All 17 upload failures clear, and the four tests pytest-dependency was skipping behind test_upload now run, three of them passing.

The fourth exposes a separate, test-side issue worth a follow-up: test_get_file_and_delete builds its expected location as f"{CONFIG.FILE_DIRECTORY}/{file_id}/{name}" with forward slashes, while save_uploaded_file builds the real one with os.path.join, so they differ by separator on Windows. Not touched here.

The five remaining failures are unrelated to this change: two from the GNU tree dependency in remote_filesystems, two from test_export removing a file the download response still holds open, and the separator one above.


pydatalab/src/pydatalab/file_utils.py
```diff
@@ -1,3 +1,4 @@
+# This file was edited with the assistance of an AI model and requires human review from the contributor.
```
ml-evs: Could you remove the AI assistance comments when you are happy with the code? These are intended to check that AI-generated PRs get at least one round of human review before being submitted/marked ready.

pydatalab/tests/test_file_utils.py
```diff
@@ -0,0 +1,21 @@
+# This file was edited with the assistance of an AI model and requires human review from the contributor.
```
ml-evs: Also here

ml-evs (commented):
Thanks for this @alioramuss, this is a useful change!

Regarding Windows support, as mentioned in #2103, I think if we can isolate the issues to the remote filesystems, then a simple runtime check that disables them when working on Windows would be useful, with a big warning that we don't want people using Windows for production deployments.

pydatalab/src/pydatalab/file_utils.py
```diff
@@ -32,18 +32,19 @@ class NotModified(RuntimeError):
...
 def get_space_available_bytes() -> int:
-    """For the configured file location, return the number of available bytes, as
-    ascertained from the filesystem blocksize and available block count (via Unix-specific
-    statvfs system call).
+    """For the configured file location, return the number of bytes available to
+    this user.
+
+    Uses `shutil.disk_usage`, which is available on every platform; the previous
+    `os.statvfs` call exists only on Unix, so every upload raised
+    `AttributeError: module 'os' has no attribute 'statvfs'` on Windows.
```
ml-evs:
```suggestion
```

These kind of docstrings (describing history and rationale for a speciifc diff) are the hallmark of current AI models and I find them more of a hindrance than a help -- simply changing the top line as you have done is fine!

ml-evs (commented):
LGTM, one final comment on the docstring and then I'll merge.

ml-evs (approved):

PR #2108, opened 2026-09-19
Make magic links single use.

davidwaroquiers:
Magic links were documented as single-use and 1 hour limit before expired. The single-use was actually not enforced, you could use the token as many times as you would like in the 1 hour limit. This PR fixes it.


ml-evs (commented):
Thanks for this @davidwaroquiers, unfortunately there's a few complications here. These links did used to be single use (from what I remember) but the issue is that many email providers (especially outlook) try to do link defusal and visit them. I don't think having a 30 or 60 minute window is that bad by default, but if we did want to move to proper single-use links then we need an extra way to provide a sign-in page that the link goes to that provides a POST button that actually sends the token. Likely with LLMs this wouldn't be too difficult to add to the API now but previously it was too much hassle than it was worth. What do you think?

davidwaroquiers:
Ah, thanks, did not know that. I'll close this for now as it's not super urgent. I think it would still be fine to do something like that at some point. Will see when I have time

